PR #74, opened 2020-04-17
[SOL-62] Copy-to-clipboard feature for Transactions and Addresses

### LINKED ISSUE #62: UX: Add an easy copy to clipboard feature for all addresses and signatures in the UI
Wherever there's a transaction signature or account address it'd be nice to have a "copy to clipboard" feature so I don't need to manually highlight the text.

nick-brown:
- Fixes #62 

copyToClipboard


src/components/Signature.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+const popover = (
+  <div className="popover fade bs-popover-right show">
+    <div className="arrow" />
+    <div className="popover-body">Copied!</div>
+  </div>
+);
+
+function Signature({ text }: SignatureProps) {
```
jstarry: What do you think about making this a bit more generic and calling it `Copyable`?

1. Account addresses aren't signatures
2. We could use this for non `<code>` wrapped text too!

Also, I think it would be nice if this was more of a HOC. It'd be cool if it accepted `children` as well `text` for its props. So that way it could be used like so:

```jsx
<Copyable text={transaction.signature}>
  <code>{transaction.signature}</code>
</Copyable>
```

nick-brown: Sounds good.  It occurred to me after the PR that addresses weren't signatures.  I like the added flexibility of accepting children.

Will change in the next 48 hours and re-submit.  Thanks for the feedback!

PR #150, opened 2020-06-02
Rename blockhash label to reduce ambiguity

jstarry:
Fixes: https://github.com/solana-labs/explorer/issues/143

Changes:
- Rename "Blockhash" label to "Nonce" when viewing a nonced transaction
- Use "Recent blockhash" for all other transactions
- Add info tooltip to add some context

> Transactions use a previously confirmed blockhash as a nonce prevent double spends

Screenshot 2020-06-03 16 47 10
Screenshot 2020-06-03 16 47 23

### DIFF
--- a/src/components/TransactionDetails.tsx
+++ b/src/components/TransactionDetails.tsx
@@ -197,7 +197,7 @@ function StatusCard({ signature }: Props) {
               {isNonce ? (
                 "Nonce"
               ) : (
-                <InfoTooltip text="Transactions use a previously confirmed blockhash as a nonce prevent double spends">
+                <InfoTooltip text="Transactions use a previously confirmed blockhash as a nonce to prevent double spends">
                   Recent Blockhash
                 </InfoTooltip>
               )}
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
Hmm, interesting idea, but my feeling is that this won't be consistent with docs.solana.com or our code. To this point, we've only used the word "Nonce" to mean durable nonce, for hashes provided by the nonce program. Everywhere else, we call the hash a "recent blockhash", which I realize isn't actually correct for Nonced transactions. Would it be possible to change this label depending on whether the transaction used a durable nonce instruction? Nonce vs Recent Blockhash

jstarry:
Good points!

> Would it be possible to change this label depending on whether the transaction used a durable nonce instruction? Nonce vs Recent Blockhash

Yes, not a problem. I updated the PR with that change

src/components/TransactionDetails.tsx
```diff
@@ -186,7 +193,15 @@ function StatusCard({ signature }: Props) {
 
         {blockhash && (
           <tr>
-            <td>Blockhash</td>
+            <td>
+              {isNonce ? (
+                "Nonce"
+              ) : (
+                <InfoTooltip text="Transactions use a previously confirmed blockhash as a nonce prevent double spends">
```
CriesofCarrots:
```suggestion
                <InfoTooltip text="Transactions use a previously confirmed blockhash as a nonce to prevent double spends">
```

PR #161, opened 2020-08-02
Integrate some components from solanabeach frontend

nishadsingh1:


nishadsingh1:
nvm moved back to https://github.com/solana-labs/solana/pull/11331

PR #168, opened 2020-10-08
Solarweave Explorer Integration

TheLoneRonin:
## Key Features

This pull request implements Solarweave in a completely unobtrusive way. It has automatic fallback compatibility to the original implementation. **Therefore, no actual cache from Solarweave needs to be present for the Explorer to work.** However, it will try to query Solarweave first.

*Please note that when retrieving blocks by slot and blockhash will not work unless an existing Solarweave Cache is implemented.*

### Search by blockhash and slot

You can now search for blocks using slots:

image

You can also search for blocks by using blockhashes:

image

### The new block page

The new block page showcases an overview of the block as well as all the transactions in the block.

image

If an invalid block is inputted. It will simply just display this page.

### Custom Solarweave Cluster Info

If you wanted to retrieve data from a custom Solarweave Database you can use the updated Cluster Info page as seen here.

### Full fallback compatibility

There is automatic fallback compatibility for the original block explorer. This means that even if there is no Solarweave cache. The Explorer will still work as intended. Just remember that block pages will not work if there is no Solarweave cache.

## Other notes

- Block History does not use Solarweave. Please see updates on: https://github.com/ArweaveTeam/arweave/issues/268 if considering using Arweave for Block History in the future.


jstarry:
Awesome, I'd love to dig through this but could you please open a PR on the monorepo https://github.com/solana-labs/solana instead? This repo is just a mirror

TheLoneRonin:
@jstarry, just created this pull request! Cheers!

https://github.com/solana-labs/solana/pull/12750

PR #173, opened 2020-12-16
adding KIN to the token registry

Blahartinger:


CriesofCarrots:
@oJshua , I think this needs to be in the monorepo, doesn't it?

mvines:
Ack, yes.  We'll need to rewrite the history in this repo to remove that commit or future exports from the monorepo into this mirror will fail

oJshua:
Ah, yes. I can rewrite the history.

oJshua:
Thanks guys. @mvines @CriesofCarrots, Rewrote the history.

PR #176, opened 2021-03-30
Adding doce finance token icon

ararog:


jstarry:
@ararog can you please host your logo elsewhere? A number of logos are hosted on ipfs now in the registry here: https://github.com/solana-labs/token-list/blob/e799eb4485bbfc77b99dd069d43d264ffa617c4b/src/tokens/solana.tokenlist.json We're trying to move away from needing to add anything directly to the explorer codebase, thanks!

PR #177, opened 2021-04-28
Fix volunerability issues

AmagiDDmxh:
Running `npm audit fix` and iterate serveral times to fix all volunerabilities issue


jstarry:
Fixed here, thanks: https://github.com/solana-labs/solana/pull/17067

PR #202, opened 2023-02-13
Add unit labels to TPS chart

ngundotra:
Custom y-tick axes labels to prevent chart.js OOM when large number of transactions is returned.

Example behavior shown for testnet:
image


steveluscher (approved):
giphy-1

steveluscher:
Cool fix! I bet there's still some code in here that's causing these components to mount and unmount like crazy on startup and/or render when they don't have to.

¯\_(°ペ)_/¯

ngundotra:
Yeah one hundred percent

PR #227, opened 2023-02-28
feat: support more ATA program's instructions

yihau:
#### Problem

We can only show the ATA program's `Create` instruction

#### Summary of Changes

- add create idempotent
- add recover nested

There are some example txs on devnet.

Create: `2tviVeEsy2ncWFVC45HMuwA9Wtmumzof8YA3DUasdA6fWHkRApc8EfQN77J7QAyumjjytDLEyw27uJAVB77uohju`
CreateIdempotent: `4cxPepPaacejJxaWW85jNkFiQdTEPUM7UQqpy1dbH8v61bUAUiqUdeP6ubEjc6SGyRFL4yEeoKh31hJdAzPW671W`
RecoverNested: `2Hymc8AgGayksndQAJnV1FuUiy6n4XNAGxWZBFGfZJRHAUJtnnn1cnDAHE5hzFe6TnRKFmybpESpPXc8GgX9i2Xn`

### DIFF
--- a/src/components/instruction/associated-token/types.ts
+++ b/src/components/instruction/associated-token/types.ts
@@ -3,6 +3,16 @@
 import { enums, type, Infer } from "superstruct";
 import { PublicKeyFromString } from "validators/pubkey";
 
+export type CreateIdempotentInfo = Infer<typeof CreateIdempotentInfo>;
+export const CreateIdempotentInfo = type({
+  source: PublicKeyFromString,
+  account: PublicKeyFromString,
+  wallet: PublicKeyFromString,
+  mint: PublicKeyFromString,
+  systemProgram: PublicKeyFromString,
+  tokenProgram: PublicKeyFromString,
+});
+
 export type RecoverNestedInfo = Infer<typeof RecoverNestedInfo>;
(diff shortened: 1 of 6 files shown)


src/components/instruction/associated-token/types.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable @typescript-eslint/no-redeclare */
+
+import { enums, type, Infer } from "superstruct";
+import { PublicKeyFromString } from "validators/pubkey";
+
+export type CreateIdempotentInfo = Infer<typeof CreateIdempotentInfo>;
+export const CreateIdempotentInfo = type({
```
steveluscher, line 7: Is it possible, instead, to _just_ export the const, and then use `Infer<typeof CreateIdempotentInfo>` wherever the _type_ is needed? This redeclaration is confusing.

steveluscher (approved):
giphy

src/components/instruction/associated-token/types.ts
yihau: I copied it from another type.ts which is in this repo haha. I think we can have another refactor PR for all of those.

Issue #229, opened 2023-03-12
[Bug] Warning message shown for a Token which misleads the audience


solankirohan:
**Describe the bug**
After creating a spl-token by following the recommended steps like creating a metadatav2 instruction with valid token metadata format and a uri which points to a json file, explorer site shows a message - "Warning! Token names and logos are not unique. This token may have spoofed its name and logo to look like another token. Verify the token's mint address to ensure it is correct."

**To Reproduce**
1. Create an spl-token by following the steps mentioned [here](https://spl.solana.com/token#example-creating-your-own-fungible-token)
2. After successfully minting the token open the mint address in the [explorer](https://explorer.solana.com) site. 
3. Observe the Warning message in pink - "Warning! Token names and logos are not unique. This token may have spoofed its name and logo to look like another token. Verify the token's mint address to ensure it is correct."

**Example Links**
- https://explorer.solana.com/address/EZAxbt341NUi7MYCX6rUdiUS87A2Tycv2otfBBwRcGsa
- https://explorer.solana.com/address/o1Mw5Y3n68o8TakZFuGKLZMGjm72qv4JeoZvGiCLEvK

**Expected behavior**
If the metadata is provided with valid uri then explorer should **NOT** show the warning message.
Earlier it was required to register the token by raising a PR on the [registry](https://github.com/solana-labs/token-list) repo, but now this repo is EOL and it is now required to create a metadatav2 instruction with valid token metadata format and a uri which points to a json file. Hence, the newly created token should be treated same as any other token which is part of the initial registry list and should not show the warning message. This message misleads the token audience who may feel it could be an incorrect token. NOTE: No such message is shown on solscan site.

**Screenshots**
<img width="1188" alt="Screenshot 2023-03-12 at 8 50 22 PM" src="https://user-images.githubusercontent.com/6348477/224570436-4170da20-60a7-47a2-8361-22b0bd7ec85f.png">

**Additional context**

codefuturance:
Hi, we are having the same issue here.  Our token is minted according to the standards provided in the EOL announcement of the SPL Token list repo. Our token has a valid token metadata format and a uri which points to a json file. But, the spoofed token warning message is displayed: 

https://explorer.solana.com/address/FEXoYTghBykVXy8Rvm7QSVEvC2NvucGJWTGipoPbyDi1/metadata

[...]

PR #230, opened 2023-03-12
only show warning if token is not registered nor does it have a valid token metadata

solankirohan:
This is a proposed solution for Issue #229 .

If we do not find the token data in the registry list nor does it have valid token metadata then only show the warning message. 

NOTE: Another proposal would be to completely remove the "unverified" flag and treat all tokens same.


steveluscher:
Shipped as #288.

PR #242, opened 2023-05-09
[feat] Add support for State Compression data to explorer

ngundotra:
To visit please explore GNsnin9c2nDGp78E69tGXyMScWfysnu2PuxQxXy1jh3R on `mainnet-beta`.

Screen Shot 2023-05-09 at 3 00 47 PM

### DIFF
--- a/app/address/[address]/concurrent-merkle-tree/page-client.tsx
+++ b/app/address/[address]/concurrent-merkle-tree/page-client.tsx
@@ -1,0 +1,27 @@
+'use client';
+
+import { ConcurrentMerkleTreeCard } from '@components/account/ConcurrentMerkleTreeCard';
+import { ParsedAccountRenderer } from '@components/account/ParsedAccountRenderer';
+import { PROGRAM_ID } from '@solana/spl-account-compression';
+import React from 'react';
+
+type Props = Readonly<{
+    params: {
+        address: string;
+    };
+}>;
+
+function ConcurrentMerkleTreeCardRenderer({
+    account,
+    onNotFound,
+}: React.ComponentProps<React.ComponentProps<typeof ParsedAccountRenderer>['renderComponent']>) {
+    const rawData = account?.data?.raw;
+    if (!rawData || account.owner.toBase58() !== PROGRAM_ID.toBase58()) {
+        return onNotFound();
+    }
+    return <ConcurrentMerkleTreeCard data={rawData} />;
+}
+
+export default function MetaplexNFTMetadataPageClient({ params: { address } }: Props) {
+    return <ParsedAccountRenderer address={address} renderComponent={ConcurrentMerkleTreeCardRenderer} />;
+}
--- a/app/address/[address]/concurrent-merkle-tree/page.tsx
+++ b/app/address/[address]/concurrent-merkle-tree/page.tsx
@@ -1,0 +1,21 @@
+import { Metadata } from 'next/types';
+
+import ConcurrentMerkleTreePageClient from './page-client';
+
+type Props = Readonly<{
+    params: {
+        address: string;
+    };
+}>;
+
+export async function generateMetadata({ params: { address } }: Props): Promise<Metadata> {
+    return {
+        description: `Contents of the SPL Concurrent Merkle Tree at address ${address} on Solana`,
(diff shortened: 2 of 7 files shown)


app/address/[address]/concurrent-merkle-tree/page.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export async function generateMetadata({ params: { address } }: Props): Promise<Metadata> {
+    return {
+        description: `Contents of the SPL Concurrent Merkle Tree at address ${address} on Solana`,
+        title: `Concurrent Merkle Tree | ${address} | Solana`,
+    };
+}
+
+export default function ConcurrentMerkleTreePage(props: Props) {
+    console.log(props);
```
Comment (jnwng) [inline app/address/[address]/concurrent-merkle-tree/page.tsx]: think you forgot this

app/components/account/ConcurrentMerkleTreeCard.tsx
```diff
@@ -0,0 +1,90 @@
+import { ConcurrentMerkleTreeAccount } from '@solana/spl-account-compression';
+import { PublicKey } from '@solana/web3.js';
+
+import { Address } from '../common/Address';
+import { TableCardBody } from '../common/TableCardBody';
+
+export function ConcurrentMerkleTreeCard({ data }: { data: Buffer }) {
+    const cmt = ConcurrentMerkleTreeAccount.fromBuffer(Buffer.from(data));
+    console.log('Account:', cmt);
```
jnwng: overlooked?

app/components/account/ConcurrentMerkleTreeCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+                        <tr>
+                            <td>Next Leaf Index</td>
+                            <td className="text-lg-end">
+                                <span className="text-monospace">{rightMostIndex}</span>
+                            </td>
+                        </tr>
+                        <tr>
+                            <td>Remaining Leaves</td>
+                            <td className="text-lg-end">
+                                <span className="text-monospace">{Math.pow(2, treeHeight) - rightMostIndex + 1}</span>
```
jnwng, line 82: maybe one more number which is the max capacity of the tree so i dont have to do more math

jnwng (approved):

app/components/account/ConcurrentMerkleTreeCard.tsx
ngundotra: How about `Total Possible Leaves` ?

ngundotra:
Current view:

Screen Shot 2023-05-09 at 3 16 42 PM

app/address/[address]/concurrent-merkle-tree/page-client.tsx
```diff
@@ -0,0 +1,27 @@
+'use client';
```
Comment (steveluscher) [inline app/address/[address]/concurrent-merkle-tree/page-client.tsx:1]: You don't actually need to split these into two files and `use client` unless there are _actual_ client functions in here (like `useCluster()`). Can you un-do this, and collapse it back into one file called `page.tsx` without the `use client`? The goal is to push `use client` as far out to the leaves as possible.

app/address/[address]/layout.tsx
```diff
@@ -31,6 +31,7 @@ import { useAnchorProgram } from '@providers/anchor';
 import { CacheEntry, FetchStatus } from '@providers/cache';
 import { useCluster } from '@providers/cluster';
 import { useTokenRegistry } from '@providers/mints/token-registry';
+import { PROGRAM_ID as ACCOUNT_COMPRESSION_ID } from '@solana/spl-account-compression';
```
Comment (steveluscher) [inline app/address/[address]/layout.tsx:34]: Did this end up pulling a crapton of code into this route, for just this one string? If so, maybe inline the string.

Comment (steveluscher) [reply app/address/[address]/layout.tsx:34]: I guess it doesn't matter, since this module is used elsewhere.

app/address/[address]/concurrent-merkle-tree/page-client.tsx
Comment (ngundotra) [reply app/address/[address]/concurrent-merkle-tree/page-client.tsx:1]: Wait, doesn't the `ParsedAccountRender` use `useCluster` when fetching accountData? So I think this split should be okay, right?

Comment (steveluscher) [reply app/address/[address]/concurrent-merkle-tree/page-client.tsx:1]: Oof. Hunh. What a mess.

PR #246, opened 2023-05-11
Eliminate `coingecko-api` NPM package

steveluscher:
Eliminate `coingecko-api` NPM package

# Summary

There's no reason not to just go to the JSON API directly. This NPM package hasn't been updated since 2020 anyway.

# Test Plan

1. Load homepage
2. Load Orca token page

Observe that both load same data as the control, and that they update on their own.


steveluscher:
@mergify rebase

PR #249, opened 2023-05-31
Feat/page title token name

### LINKED ISSUE #243: Render token name in page title
Consider this mint: https://explorer.solana.com/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v

The page title for that page is: `Transaction History | EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v | Solana`

The code that renders that page title is here:

https://github.com/solana-labs/explorer/blob/9014bf54330a2589471c75e01dbb98c914e7fb48/app/address/%5Baddress%5D/page.tsx#L10-L15

The improvement is to render the _token name_ instead of the mint address, so that in Google the page title appears as `Transaction History | Token | USDC | Solana`

Considerations:

* It's very important not to create two URLs with the same page title. Consider the case where two tokens have identical names; how would you prevent their URLs from rendering the same title?
* Apply the same technique to all token-accessible routes underneath `/address/**`

jdubpark:
For the title metadata generation, Next.js doesn't allow using client-side hooks on server-side components (see [this issue](https://github.com/vercel/next.js/issues/46372)). Thus, we can't use `useTokenRegistry` as it's marked with 'use client' for frontend. 

When users visit each address page with the token name title, both the client-side and server-side rendering fetches the token list. These identical fetches are optimized by Next.js via [automatic fetch request deduping](https://nextjs.org/docs/app/building-your-application/data-fetching#automatic-fetch-request-deduping).

> It's very important not to create two URLs with the same page title. Consider the case where two tokens have identical names; how would you prevent their URLs from rendering the same title?
1. We could append a small slice of address to the token name, e.g. USD Coin (EPjFW), but this wouldn't prevent same token names with the same "small slice" (using vanity address). 
2. Another method could be just highlighting the official tokens, e.g. USD Coin (official), and leaving the other non-official tokens as-is, but this requires an active management of "official" tokens.

Fixes #243.

### DIFF
--- a/app/utils/get-readable-title-from-address.ts
+++ b/app/utils/get-readable-title-from-address.ts
@@ -1,4 +1,4 @@
-import { Strategy, TokenInfo } from '@solana/spl-token-registry';
+import { Strategy } from '@solana/spl-token-registry';
 import { Cluster } from '@utils/cluster';
 import getTokenList from '@utils/get-token-list';
 
@@ -34,14 +34,12 @@ export default async function getReadableTitleFromAddress(props: AddressPageMeta
 
     try {
         const tokenList = await getTokenList(cluster, Strategy.Solana);
-        if (typeof tokenList.get(address)?.name === 'undefined') {
+        const tokenName = tokenList.get(address)?.name;
+        if (tokenName == null) {
             return address;
         }
-
-        const tokenName = (tokenList.get(address) as TokenInfo).name;
-        const tokenAddressBlob = address.slice(0, 2) + '\u2026' + address.slice(-2);
-
-        return `Token | ${tokenName} (${tokenAddressBlob})`;
+        const tokenDisplayAddress = address.slice(0, 2) + '\u2026' + address.slice(-2);
+        return `Token | ${tokenName} (${tokenDisplayAddress})`;
     } catch {
         return address;
     }
--- a/app/utils/get-readable-title-from-address.ts
+++ b/app/utils/get-readable-title-from-address.ts
@@ -26,9 +26,14 @@ export default async function getReadableTitleFromAddress(props: AddressPageMeta
     try {
         const tokenList = await getTokenList(parsedCluster, Strategy.Solana);
         if (typeof tokenList.get(address)?.name === 'undefined') {
+            console.log(tokenList.get(address))
             return address;
         }
-        return `Token | ${(tokenList.get(address) as TokenInfo).name}`;
+
+        const tokenName = (tokenList.get(address) as TokenInfo).name;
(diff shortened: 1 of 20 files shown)


jdubpark:
ping @steveluscher @ngundotra

steveluscher:
> We could append a small slice of address to the token name, e.g. USD Coin (EPjFW), but this wouldn't prevent same token names with the same "small slice" (using vanity address).

Good catch! You could always do a range slice to eliminate this. `USD Coin (EP…1v)`

app/utils/get-readable-title-from-address.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (cluster === 'devnet') parsedCluster = Cluster.Devnet
+    else if (cluster === 'testnet') parsedCluster = Cluster.Testnet
+    else if (cluster === 'custom') parsedCluster = Cluster.Custom
+
+    return await getTokenList(parsedCluster, Strategy.Solana)
+        .then((tokens: TokenInfoMap) => {
+            if (typeof tokens.get(address)?.name === 'undefined') throw new Error('invalid token address')
+            return `Token | ${(tokens.get(address) as TokenInfo).name}`
+        })
+        .catch(() => address) // fallback to address if token not found
```
steveluscher: You can write this in async/await style.
```ts
try {
  const tokenList = await getTokenList(parsedCluster, Strategy.Solana);
  if (typeof tokenList.get(address)?.name === 'undefined') {
    throw new Error('invalid token address');
  }
  return `Token | ${(tokenList.get(address) as TokenInfo).name}`;
} catch {
  return address;
}
```

steveluscher (changes requested):
Left a few drive-by comments. Feel free to take another pass and I'll review it more closely on the next one!

app/utils/get-readable-title-from-address.ts
jdubpark: Looks good!

jdubpark:
Is there any (similar) way to fetch NFT names from addresses?

jdubpark:
Range slice has been added, e.g. USD Coin (EP…1v).

Unrelated, `getTokenList()` returns an empty list when suing `custom` cluster and RPC.
https://github.com/solana-labs/explorer/blob/17f644efa1f6f174e1dd7bc46bc397ca55f0a004/app/providers/mints/token-registry.tsx#L17-L19

jdubpark (commented):
Fixed lint issues and changed try-catch

steveluscher (approved):
thank-you

PR #253, opened 2023-06-02
Parse feature activation accounts

ripatel-fd:
Adds https://github.com/solana-labs/solana/wiki/Feature-Gate-Activation-Schedule support to UI

### DIFF
--- a/app/components/account/FeatureAccountSection.tsx
+++ b/app/components/account/FeatureAccountSection.tsx
@@ -1,0 +1,104 @@
+import { Address } from '@components/common/Address';
+import { TableCardBody } from '@components/common/TableCardBody';
+import { Account } from '@providers/accounts';
+import * as BufferLayout from '@solana/buffer-layout';
+import { PublicKey } from "@solana/web3.js";
+
+import { UnknownAccountCard } from './UnknownAccountCard';
+
+export const FEATURE_PROGRAM_ID = new PublicKey('Feature111111111111111111111111111111111111');
+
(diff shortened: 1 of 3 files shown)


app/components/account/FeatureAccountSection.tsx
```diff
@@ -0,0 +1,104 @@
...
+        activatedAt = (
+            <tr>
+                <td>Activated At Slot</td>
+                <td className="text-lg-end">
+                    <code>{feature.activatedAt}</code>
+                </td>
+            </tr>
+        );
+    }
+    console.log(feature);
```
michaelh-laine, line 77: is this intentionally left?

ngundotra:
Can you please follow the format for PRs + issues next time? @ripatel-jump 

Just add a screenshot + a link or two to test the feature you've added would help

steveluscher:
Thanks for this! I'm concerned to the degree that this UI would only work in two of these three cases:

1. ✅ When a feature has been activated at the epoch boundary for which it was scheduled, we can read out the activation timestamp and render it in the UI
2. ✅ When a feature has been scheduled for activation but the activation epoch boundary hasn't been reached, the timestamp is none and we can render ‘not active’ in the UI
3. ❌ When a feature has not been scheduled at all, its feature flag account doesn't exist so there's no UI we can show one way or another

I worry that this will confuse folks coming to the Explorer in case 3, expecting to see information about a feature, and seeing nothing instead.

PR #256, opened 2023-06-05
Use experimental web3js for homepage data

mcintyre94:
This PR is a first effort to start proving out the experimental web3js on Explorer. All data displayed on the homepage is fetched using functions of the experimental web3js.

Notes:

- Annoyingly this doesn't eliminate the old web3js dependency from the homepage. There's a big tree of providers with various web3.js (and other) dependencies in [the app layout](https://github.com/solana-labs/explorer/blob/master/app/layout.tsx).

- The main difference for the functions covered here is using bigint in the new web3js. I've tried to keep processing of bigint as bigint, for example calculating percentages of bigints is done without converting them to numbers. I've mostly left display components as number, for example `<Slot>` is used in many places and is left unchanged, and `<CountUp>` is a 3rd party dependency which expects numbers.

- I've added the experimental web3js using an alias `web3js-experimental`. pnpm has an issue where peerDependencies (some libs have web3js as a peer dependency) don't work correctly, [see issue + workaround used here](https://github.com/pnpm/pnpm/issues/6588#issuecomment-1563850988)

- We mostly used the exported return types from web3js as our data structures. The experimental web3js doesn't export return types, so I've created minimal types that just contain the fields that we need and then passed these around as needed

### DIFF
--- a/app/components/LiveTransactionStatsCard.tsx
+++ b/app/components/LiveTransactionStatsCard.tsx
@@ -224,7 +224,8 @@ function AnimatedTransactionCount({ info }: { info: PerformanceInfo }) {
     const countUpRef = React.useRef({ lastUpdate: 0, period: 0, start: 0 });
     const countUp = countUpRef.current;
 
-    const { transactionCount: txCount, avgTps } = info;
+    const { transactionCount, avgTps } = info;
+    const txCount = Number(transactionCount);
 
     // Track last tx count to reset count up options
     if (txCount !== txCountRef.current) {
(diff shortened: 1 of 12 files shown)


steveluscher (approved):
This PR doesn't completely remove `@solana/web3.js` from the dependency graph of the homepage (eg. it's still in things like `SolanaPingProvider`) but it's an awesome start!

> The experimental web3js doesn't export return types, so I've created minimal types that just contain the fields that we need and then passed these around as needed

That's probably the right thing to do, but consider doing `ReturnType<foo.getAccountInfo>` or whatever if it's really cumbersome.

PR #258, opened 2023-06-06
Remove genesis hash call from cluster provider

mcintyre94:
It turns out that nothing uses the `genesisHash` that we fetch on every page as part of the cluster provider. Deleting code is the best way to refactor it :)


steveluscher (approved):
Oh man, sorry for the delay; you don't need my review for this sort of stuff :)
giphy-1

PR #267, opened 2023-07-11
Refactor ClusterProvider to remove legacy web3js dependency

mcintyre94:
This provider makes 3 requests, and makes their responses available to child components:

- `getFirstAvailableBlock`
- `getEpochInfo`
- `getEpochSchedule`

The first 2 are reasonably straightforward, we have basically the same structure as the legacy web3js, with just number/bigint changes.

`getEpochSchedule` is a bit more complex. The existing API exposes a class with a bunch of functionality for finding the epoch for a slot, and the first/last slot for an epoch. None of this is RPC functionality, it's all baked into the legacy web3js code. Since the experimental web3js doesn't do any of that, I've copied these functions into the Explorer codebase, as pure functions that take an `EpochSchedule` (pure data returned by the new RPC method) and a slot/epoch (bigint). 

See existing web3js code here: https://github.com/solana-labs/solana-web3.js/blob/9232d2b1019dc50f852ad70aa81624e751d76161/packages/library-legacy/src/epoch-schedule.ts 

Also note that I hit a bug in experimental web3js where some of these functions are incorrectly typed as unknown: https://github.com/solana-labs/solana-web3.js/issues/1389
This was easy enough to work around for now

I've also moved `localStorageIsAvailable` from `utils/index.ts` to its own `utils/local-storage`. This lets us import it without pulling in the web3js dependency in `utils/index.ts`

The result of this PR is that the `ClusterProvider` in the root layout no longer pulls in the legacy web3js dependency.

### DIFF
--- a/app/components/SearchBar.tsx
+++ b/app/components/SearchBar.tsx
@@ -242,7 +242,7 @@ async function buildDomainOptions(connection: Connection, search: string, option
 }
 
 // builds local search options
-function buildOptions(rawSearch: string, cluster: Cluster, tokenRegistry: TokenInfoMap, currentEpoch?: number) {
+function buildOptions(rawSearch: string, cluster: Cluster, tokenRegistry: TokenInfoMap, currentEpoch?: bigint) {
     const search = rawSearch.trim();
     if (search.length === 0) return [];
 
@@ -285,7 +285,8 @@ function buildOptions(rawSearch: string, cluster: Cluster, tokenRegistry: TokenI
             ],
         });
 
-        if (currentEpoch !== undefined && Number(search) <= currentEpoch + 1) {
+        // Parse as BigInt but not if it starts eg 0x or 0b
+        if (currentEpoch !== undefined && !(/^0\w/.test(search)) && BigInt(search) <= currentEpoch + 1n) {
             options.push({
                 label: 'Epoch',
                 options: [
--- a/app/epoch/[epoch]/page-client.tsx
+++ b/app/epoch/[epoch]/page-client.tsx
@@ -12,6 +12,8 @@ import { ClusterStatus } from '@utils/cluster';
 import { displayTimestampUtc } from '@utils/date';
 import React from 'react';
 
+import { getFirstSlotInEpoch, getLastSlotInEpoch } from '@/app/utils/epoch-schedule';
+
 type Props = {
     params: {
         epoch: string;
@@ -71,8 +73,8 @@ function EpochOverviewCard({ epoch }: OverviewProps) {
         return <LoadingCard message="Loading epoch" />;
     }
 
-    const firstSlot = epochSchedule.getFirstSlotInEpoch(epoch);
-    const lastSlot = epochSchedule.getLastSlotInEpoch(epoch);
+    const firstSlot = getFirstSlotInEpoch(epochSchedule, BigInt(epoch));
+    const lastSlot = getLastSlotInEpoch(epochSchedule, BigInt(epoch));
 
     return (
         <>
@@ -108,13 +110,13 @@ function EpochOverviewCard({ epoch }: OverviewProps) {
                     <tr>
                         <td className="w-100">First Slot</td>
                         <td className="text-lg-end font-monospace">
-                            <Slot slot={firstSlot} />
+                            <Slot slot={Number(firstSlot)} />
                         </td>
                     </tr>
                     <tr>
                         <td className="w-100">Last Slot</td>
                         <td className="text-lg-end font-monospace">
-                            <Slot slot={lastSlot} />
+                            <Slot slot={Number(lastSlot)} />
                         </td>
                     </tr>
                     {epochState.data.firstTimestamp && (
--- a/app/epoch/[epoch]/page-client.tsx
+++ b/app/epoch/[epoch]/page-client.tsx
@@ -110,13 +110,13 @@ function EpochOverviewCard({ epoch }: OverviewProps) {
                     <tr>
                         <td className="w-100">First Slot</td>
                         <td className="text-lg-end font-monospace">
-                            <Slot slot={Number(firstSlot)} />
+                            <Slot slot={firstSlot} />
                         </td>
                     </tr>
                     <tr>
                         <td className="w-100">Last Slot</td>
                         <td className="text-lg-end font-monospace">
-                            <Slot slot={Number(lastSlot)} />
+                            <Slot slot={lastSlot} />
                         </td>
                     </tr>
                     {epochState.data.firstTimestamp && (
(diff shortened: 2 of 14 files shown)


app/components/SearchBar.tsx
```diff
@@ -285,7 +285,7 @@ function buildOptions(rawSearch: string, cluster: Cluster, tokenRegistry: TokenI
             ],
         });
 
-        if (currentEpoch !== undefined && Number(search) <= currentEpoch + 1) {
+        if (currentEpoch !== undefined && Number(search) <= currentEpoch + BigInt(1)) {
```
steveluscher: I guess theoretically we could parse the input as a BigInt too, but being careful not to allow fancy integer literal strings that start with `0x`, `0b` and `0o`.

```ts
if (search.length === 0 || /^0\w/.test(search)) return [];

if (currentEpoch !== undefined && BigInt(search) <= currentEpoch + 1n) { /* ... */ }
```

(Also, I'd just use BigInt literals like `1n`)

app/providers/cluster.tsx
```diff
@@ -115,19 +124,23 @@ async function updateCluster(dispatch: Dispatch, cluster: Cluster, customUrl: st
...
+            rpc.getEpochInfo().send(),
         ]);
 
         dispatch({
             cluster,
             clusterInfo: {
                 epochInfo,
                 epochSchedule,
-                firstAvailableBlock,
+                firstAvailableBlock: firstAvailableBlock as bigint, // todo why am I getting unknown here?
```
steveluscher: 🤔

app/utils/cluster.ts
```diff
@@ -41,26 +39,18 @@ export function clusterName(cluster: Cluster): string {
...
     switch (cluster) {
         case Cluster.Devnet:
-            return process.env.NEXT_PUBLIC_DEVNET_RPC_URL ?? modifyUrl(DEVNET_URL);
+            return process.env.NEXT_PUBLIC_DEVNET_RPC_URL ?? DEVNET_URL;
         case Cluster.MainnetBeta:
-            return process.env.NEXT_PUBLIC_MAINNET_RPC_URL ?? modifyUrl(MAINNET_BETA_URL);
+            return process.env.NEXT_PUBLIC_MAINNET_RPC_URL ?? MAINNET_BETA_URL;
         case Cluster.Testnet:
-            return process.env.NEXT_PUBLIC_TESTNET_RPC_URL ?? modifyUrl(TESTNET_URL);
+            return process.env.NEXT_PUBLIC_TESTNET_RPC_URL ?? TESTNET_URL;
```
steveluscher: Was this accidentally committed?

app/epoch/[epoch]/page-client.tsx
```diff
@@ -108,13 +110,13 @@ function EpochOverviewCard({ epoch }: OverviewProps) {
                     <tr>
                         <td className="w-100">First Slot</td>
                         <td className="text-lg-end font-monospace">
-                            <Slot slot={firstSlot} />
+                            <Slot slot={Number(firstSlot)} />
```
Comment (steveluscher) [inline app/epoch/[epoch]/page-client.tsx]: Can we update `Slot` to take a BigInt instead?

app/providers/cluster.tsx
```diff
@@ -128,19 +129,20 @@ async function updateCluster(dispatch: Dispatch, cluster: Cluster, customUrl: st
...
         dispatch({
             cluster,
             clusterInfo: {
                 epochInfo,
-                epochSchedule,
-                firstAvailableBlock: firstAvailableBlock as bigint, // todo why am I getting unknown here?
+                // These are incorrectly typed as unknown
+                // See https://github.com/solana-labs/solana-web3.js/issues/1389
+                epochSchedule: epochSchedule as EpochSchedule,
+                firstAvailableBlock: firstAvailableBlock as bigint,
```
steveluscher, line 145: Oh no. I commented on https://github.com/solana-labs/solana-web3.js/issues/1389

app/utils/local-storage.ts
```diff
@@ -0,0 +1,10 @@
+export function localStorageIsAvailable() {
+  const test = 'test';
+  try {
+    localStorage.setItem(test, test);
+    localStorage.removeItem(test);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
```
steveluscher: <3

app/utils/local-storage.ts
```diff
@@ -0,0 +1,10 @@
+export function localStorageIsAvailable() {
+  const test = 'test';
+  try {
+    localStorage.setItem(test, test);
+    localStorage.removeItem(test);
```
steveluscher: Consider doing this once, ever.

```ts
let localStorageIsAvailableDecision: boolean | undefined;
export function localStorageIsAvailable() {
  if (localStorageIsAvailableDecision === undefined) {
    const test = 'test';
    try {
      localStorage.setItem(test, test);
      localStorage.removeItem(test);
      localStorageIsAvailableDecision = true;
    } catch (e) {
      localStorageIsAvailableDecision = false;
    }
  }
  return localStorageIsAvailableDecision;
}
```

steveluscher (approved):
giphy

app/utils/cluster.ts
mcintyre94: It's an intentional simplification. Previously we used the `clusterApiUrl` function to get eg. `https://api.mainnet-beta.solana.com` and then replaced the `api` with `explorer-api` to get our custom explorer RPC. Since the new web3js doesn't expose default URLs and I don't want `clusterApiUrl` pulling in the web3js dependency I just replaced it by hardcoding the `explorer-api` URLs instead. 

Before this change we'd also use the `api` URLs instead of `explorer-api` on localhost, but they seem to behave the same for me (namely I get 403s on mainnet with both URLs, both work fine on dev/testnet). So I'm not sure there's any benefit to what we were doing before.

app/epoch/[epoch]/page-client.tsx
Comment (mcintyre94) [reply app/epoch/[epoch]/page-client.tsx]: We can but it'll blow the changes up a bunch because it's used in lots of places. Might be worth it though, I'll have another look

app/components/SearchBar.tsx
mcintyre94: Updated to use this. I moved the regex to just before the BigInt parse because there are tokens with names that start `0x` and I don't want to break search for them

app/epoch/[epoch]/page-client.tsx
Comment (mcintyre94) [reply app/epoch/[epoch]/page-client.tsx]: Oh I'm being dumb, just changed Slot to take either. All we do with it is `toString` and `toLocaleString` and that works fine for both!

app/utils/cluster.ts
mcintyre94: Changed this back, I've just replaced the `clusterApiUrl` calls with their responses now.

PR #268, opened 2023-07-12
Refactor MintsProvider to remove web3js dependency

mcintyre94:
Note: Merge after #267 (this PR builds on that)
Actual diff: https://github.com/solana-labs/explorer/compare/1f5128679ff182b4d1c75ca5d6062b8523286ff2...mcintyre94:explorer:mints-provider-refactor 

This PR refactors out the `MintsProvider` provider that was used on all pages. This provider was just a wrapper for 2 other providers:

- `TokenRegistryProvider` remains in the root layout. It didn't have any web3js dependency so is just copied as-is to a new provider
- `LargestAccountsProvider` is only used for the `address/[address]/largest` route. It's removed from the root layout and now only used on that page. Refactoring it to use experimental web3js is blocked by https://github.com/solana-labs/solana-web3.js/pull/1390 but as it's no longer on the homepage path there's no rush there 

The PR also refactors `ClusterModal` to remove its web3js dependency, by inlining a utils function that only it was using

### DIFF
--- a/app/components/ClusterModal.tsx
+++ b/app/components/ClusterModal.tsx
@@ -3,7 +3,6 @@
 import { useCluster, useClusterModal, useUpdateCustomUrl } from '@providers/cluster';
 import { useDebounceCallback } from '@react-hook/debounce';
 import { Cluster, clusterName, CLUSTERS, clusterSlug, ClusterStatus } from '@utils/cluster';
-import { assertUnreachable } from '@utils/index';
 import dynamic from 'next/dynamic';
 import Link from 'next/link';
 import { usePathname, useRouter, useSearchParams } from 'next/navigation';
@@ -80,6 +79,10 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
             )}
         </>
     );
+}
+
+function assertUnreachable(_x: never): never {
+    throw new Error('Unreachable!');
 }
 
 function ClusterToggle() {
--- a/app/utils/index.ts
+++ b/app/utils/index.ts
@@ -16,10 +16,6 @@ export type SignatureProps = {
     signature: TransactionSignature;
 };
 
-export function assertUnreachable(_x: never): never {
-    throw new Error('Unreachable!');
-}
-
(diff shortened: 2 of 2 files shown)


mcintyre94:
This needs a bit more work on the distribution/largest page

On current Explorer there's an error on USDC mainnet, the RPC returns:
```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32010,
        "message": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA excluded from account secondary indexes; this RPC method unavailable for key"
    },
    "id": "8240185a-6b40-4c59-bbc7-babc334ee899"
}
```

[This displays an error message currently](https://explorer.solana.com/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v/largest):
image

[On this branch](https://explorer-pdtlc15ek-solana-labs.vercel.app/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v/largest) it doesn't display the error:
image

I think it's also not displaying the loading state the way it currently does, but the error state is clearer and probably the same root cause.

steveluscher:
Yeah, Triton nerfed that call on purpose. It's too expensive.

app/address/[address]/largest/layout.tsx
```diff
@@ -0,0 +1,10 @@
+import { LargestAccountsProvider } from "@providers/largest-accounts";
+
+type Props = { children: React.ReactNode };
+export default function AddressLayout({ children }: Props) {
+  return (
+    <LargestAccountsProvider>
+      {children}
+    </LargestAccountsProvider>
+  );
+}
```
Comment (steveluscher) [inline app/address/[address]/largest/layout.tsx]: EOF linebreak.

mcintyre94:
I'll rework this - most of it is covered by #270 which gets rid of that token distribution page and completely removes the `LargestAccountsProvider` I moved around here

mcintyre94:
This PR is now just the small change to `ClusterModal` which removes the web3js dependency there. Everything else disappeared after #270

PR #275, opened 2023-07-18
Refactor search to remove legacy web3js dependency

mcintyre94:
This PR refactors the search bar to remove the client-side legacy web3js dependency. This means that the core app layout, shared by every page, no longer pulls in legacy web3js

- The client-side call to `getDomainInfo` is replaced with a GET request to a new API `/api/domain-info/[domain]`
- This API just calls `getDomainInfo` and returns the response as JSON. This function and everything else in `utils/name-service` that can be run on the server are extracted to a new `utils/domain-info`. This is necessary because `utils/domain-info` has `use client` (needed for its client-side hooks)
- The API has a 24h cache-response header. I haven't done any manual client-side caching, Next's fetch de-dupe should be sufficient. 
- For now I haven't changed how we get the domain info. Longer term I'd like to support multiple domains: https://github.com/solana-labs/explorer/issues/271 
- I've extracted everything used by the search bar from `utils/tx` to a new `utils/programs`. I've removed the legacy web3js dependency, which was just used to pull in program public keys

The result of this is that some pages load ~100kb less JS. I'm not exactly sure how some of these numbers work though, eg the homepage is already not including web3js despite including the search bar. 

Screenshot 2023-07-18 at 15 23 19

We can also inspect the bundle before and after this change, this is the app/layout endpoint. On master we have the legacy web3js 
Screenshot 2023-07-18 at 15 01 10

After we've stripped many of these dependencies out:
Screenshot 2023-07-18 at 15 02 33

Mostly it shows that we should stop including that token list though... https://github.com/solana-labs/explorer/issues/201


mcintyre94:
> Are we sure about this? I see code like this in the homepage bundles.

I'm not sure TBH! But last week the homepage was at ~90kb more JS first load than it is on master atm. See https://twitter.com/callum_codes/status/1679124485218226176 where I was first experimenting with the bundle analyzer. And I expected this change to be the one that caused that drop, not an earlier one. I think I'm just not really understanding exactly how these bundles are being put together and where those numbers are coming from though.

PR #281, opened 2023-07-25
Remove token registry dependency

### LINKED ISSUE #201: explorer: Use Unified Token-list API
#### Problem

Currently the explorer queries the following to get token-metadata:

1. Check if Token is in the legacy token-list
2. If not in token-list try on-chain via Metaplex Fungible token metadata

This has the following problems:

1. Wallets do metaplex fungible token metadata first, then legacy token-list. This allows someone with a token to update their metadata by creating the on-chain account
2. We have to load the legacy token-list on the front end to manage searching, querying, etc
3. Searching for on-chain metadata via the searchbar is not done today, and is not feasible via `getProgramAccounts` (nor recommended)

#### Proposed Solution

Solflare created a [Unified Token API](https://github.com/solflare-wallet/utl-api) that manages caching, searching, and verifying tokens via coingecko, legacy token-list, and Metaplex Fungible token metadata.

Proposal is to use it to create a faster experience for users on the explorer searching and viewing their tokens.

We have two options:

1. Deploy our own instance and use it
2. Use Solflare's public API https://token-list-api.solana.cloud (The team is fine with the explorer using it)

Proposing we use option 2 for speed to ship. Happy to work on it if consensus is met.

cc: @steveluscher @jstarry

mcintyre94:
This PR removes and replaces the `@solana/spl-token-registry` dependency, which was by far the largest dependency and was pulling in an >3MB outdated JSON file on every page. It should improve performance across the board, especially on slower networks. Token data will also be much more up to date, and tokens created since we deprecated the token registry will be represented correctly. 

The main replacement for this dependency is [Solflare's Unified Token List (UTL) SDK](https://github.com/solflare-wallet/utl-sdk). This is a typescript SDK that provides information about a token. It's got great performance and covers most use cases we have really well.

It has some tradeoffs:

- It's quite a heavy dependency, mostly because of its `metaplex-foundation/js` dependency which allows it to fetch on-chain data for tokens that aren't in its API/CDN already. For this reason I've used the underlying API directly for the search bar, to avoid pulling the SDK into the root layout. This is actually the same behaviour as the SDK has for search anyway, since falling back to an on-chain query can't be done for search.
- It doesn't have all the per-token data from the old token registry, specifically for extensions it only has the coingecko one, and is missing eg bridge contracts and websites. We only use these on the token address page. Longer term we might want to remove these since we don't have a way to get this data for new tokens. But for now I'm using a CDN version of our last token list to fetch the token info including these legacy fields. For tokens that aren't in that list I use the SDK as a fallback, and just return the data without those extensions.

Other than these cases, we now use the SDK for all token info. 

Some specific implementation decisions:

- The address/[address] page fetches token info once (using the legacy CDN, then API fallback), and uses prop drilling for child components.
- `useAccountOwnedTokens` now returns tokens with the fetched name/logo/symbol for each, fetched from the UTL SDK. 
- The `<Address>` component used to search the token registry in all cases, like it does with programs etc. This is no longer the case since we no longer have the entire token list on every page. A new prop `tokenListInfo` can be used to pass known token info in. This is used for example when rendering the user's owned tokens. Alternatively `fetchTokenLabelInfo` can be used to asynchronously fetch the token for the address, and update  the address to display it after fetching. This should be used for addresses that might be a token, for example transaction accounts. This is the equivalent of `useMetadata` which fetches NFT metadata. 
- I've used SWR in a few places where we fetch token info as an enhancement, usually just to display the symbol for a token as a suffix for a count, eg 5 USDC. In these cases I use the loading state of SWR to distinguish between data not fetched yet (we display no suffix) and no data available (we typically add the 'tokens' suffix)

The result of all this is that the huge tokenlist file which dominated our bundle is gone:

Master:
master-all-min

This branch:
rtr-all-min

Looking at just the layout shared by all pages, the effect is even more drastic.

Master:
master-layout-min

This branch:
rtr-layout-min

Closes #201


app/components/SearchBar.tsx
```diff
@@ -57,10 +55,17 @@ export function SearchBar() {
...
-
+        const options = buildOptions(search, cluster, clusterInfo?.epochInfo.epoch);
         setSearchOptions(options);
 
+        if (search.length > 0) {
+            buildTokenOptions(search, cluster).then(tokenOptions => {
+                if (tokenOptions) {
+                    setSearchOptions(options => [...options, tokenOptions])
+                }
+            })
```
steveluscher: This is a sure way to get annoying race conditions.

* effect runs
* `setSearchOptions(A)`
* effect runs again
* `setSearchOptions(B)`
* `setSearchOptions(buildTokenOptions(B))`
* `setSearchOptions(buildTokenOptions(A))` <== shit

Use `useAsycEffect()` and honour the cancellation thingy.

steveluscher:
```ts
useAsyncEffect(async isMounted => {
  const options = buildOptions(…);
  if (search.length > 0) {
    buildTokenOptions(...).then(tokenOptions => {
      if (isMounted()) {
        setSearchOptions(...);
      }
    });
  }
}, [search]);
```

app/utils/token-search.ts
```diff
@@ -0,0 +1,69 @@
...
+    label: string;
+    value: string[];
+    pathname: string;
+}
+
+export async function searchTokens(
+    search: string,
+    cluster: Cluster
+): Promise<SearchElement[]> {
+    if (process.env.NEXT_PUBLIC_DISABLE_TOKEN_SEARCH || !search) {
```
steveluscher, line 40: smart-thinking

app/utils/token-search.ts
```diff
@@ -0,0 +1,69 @@
...
+        return Promise.resolve([]);
```
steveluscher: You don't have to do this. You can just return `[]` because the function itself is `async`;

app/utils/token-search.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    // See https://github.com/solflare-wallet/utl-sdk/blob/master/src/types.ts#L5
+    let chainId: number;
+    if (cluster === Cluster.MainnetBeta) chainId = 101;
+    else if (cluster === Cluster.Testnet) chainId = 102;
+    else if (cluster === Cluster.Devnet) chainId = 103;
+    else { return Promise.resolve([]); }
```
steveluscher: `return [];`

app/utils/token-search.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const apiResponse = await fetch(`https://token-list-api.solana.cloud/v1/search?query=${search}&chainId=${chainId}&start=0&limit=20`);
```
steveluscher: I don't have an attack in mind, but I just have a general sense that we should [`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) these user-generated parts.

app/address/[address]/layout.tsx
```diff
@@ -189,12 +192,25 @@ export default function AddressLayout({ children, params }: Props) {
...
+        if (!isToken) return;
+
+        assertIsBase58EncodedAddress(address);
+        let isSubscribed = true;
+        getTokenInfo(address, cluster, url).then(token => {
+            if (isSubscribed) setTokenDetails(token)
+        })
+
+        return () => { isSubscribed = false };
+    }, [cluster, url, isToken]);
```
Comment (steveluscher) [inline app/address/[address]/layout.tsx]:
```ts
    useAsyncEffect(async isMounted => {
        if (!isToken) return;
        assertIsBase58EncodedAddress(address);
        const token = await getTokenInfo(address, cluster, url);
        if (!isMounted()) {
          return;
        }
        setTokenDetails(token);
    }, [cluster, url, isToken]);
```

app/utils/token-info.ts
```diff
@@ -0,0 +1,34 @@
...
+    return new Client(config);
+}
+
+export async function getTokenInfo(
+    address: Base58EncodedAddress,
+    cluster: Cluster,
+    connectionString: string
+): Promise<Token | undefined> {
+    const client = makeUtlClient(cluster, connectionString);
+    if (!client) return Promise.resolve(undefined);
```
steveluscher:
```ts
    if (!client) return undefined;
```

app/utils/token-info.ts
```diff
@@ -32,3 +32,14 @@ export async function getTokenInfo(
...
     return token;
 }
+
+export async function getTokenInfos(
+    addresses: PublicKey[],
+    cluster: Cluster,
+    connectionString: string
+): Promise<Token[] | undefined> {
+    const client = makeUtlClient(cluster, connectionString);
+    if (!client) return Promise.resolve(undefined);
```
steveluscher:
```ts
    if (!client) return undefined;
```

next.config.mjs
```diff
@@ -12,8 +12,8 @@ const nextConfig = {
   images: {
     remotePatterns: [
       {
-        hostname: 'raw.githubusercontent.com',
-        pathname: '/solana-labs/token-list/main/assets/**',
+        hostname: '**',
+        pathname: '**',
```
steveluscher: Ooh. This is going to blow our Vercel image optimization credits immediately, I've seen it happen before and it's not pretty.

What's the issue here? There's no pattern we can match on for tokens from this thing?

https://vercel.com/docs/concepts/image-optimization/managing-image-optimization-costs#how-to-minimize-image-optimization-costs

app/address/[address]/layout.tsx
```diff
@@ -165,17 +165,34 @@ function AddressLayoutInner({ children, params: { address } }: Props) {
...
+        if (infoStatus === FetchStatus.Fetched && infoProgram === "spl-token" && pubkey) {
+            getFullLegacyTokenInfoUsingCdn(pubkey, cluster, url).then(tokenInfo => {
+                if (isSubscribed) {
+                    setFullLegacyTokenInfo(tokenInfo)
+                }
+            })
+        }
+
+        return () => { isSubscribed = false; }
+    }, [infoStatus, infoProgram, pubkey?.toString(), cluster, url])
```
Comment (steveluscher) [inline app/address/[address]/layout.tsx]:
```ts
    useAsyncEffect(async isMounted => {
        if (infoStatus === FetchStatus.Fetched && infoProgram === "spl-token" && pubkey) {
            const tokenInfo = await getFullLegacyTokenInfoUsingCdn(pubkey, cluster, url)
            if (!isMounted()) {
              return;
            }
            setFullLegacyTokenInfo(tokenInfo);
        }
    }, [infoStatus, infoProgram, pubkey?.toString(), cluster, url])
```

app/utils/tx.ts
```diff
@@ -47,6 +47,16 @@ export function addressLabel(address: string, cluster: Cluster, tokenRegistry?:
     );
 }
 
+export function addressLabel_(address: string, cluster: Cluster, tokenName?: string): string | undefined {
```
steveluscher: Underscore?

app/components/account/TokenAccountSection.tsx
```diff
@@ -343,22 +344,25 @@ function NonFungibleTokenMintAccountCard({
...
-                ◎<span className="font-monospace">{new BigNumber(info.tokenAmount.uiAmountString).toFormat(9)}</span>
-            </>
-        );
-    } else {
-        balance = <>{info.tokenAmount.uiAmountString}</>;
-        unit = tokenRegistry.get(info.mint.toBase58())?.symbol || 'tokens';
-    }
+    const balance = info.isNative ? (
+        <>
+            ◎<span className="font-monospace">{new BigNumber(info.tokenAmount.uiAmountString).toFormat(9)}</span>
```
steveluscher: This is probably just me, but I have a slight preference against Unicode in source code. Alternative: `{'\u25ce'}`

But don't change it because I said so; feel free to ignore.

steveluscher: OMG what an opportunity to nuke `BigNumber` in favor of `Intl.NumberFormat`. Not now, but later.

app/components/account/history/TokenTransfersCard.tsx
```diff
@@ -29,16 +31,14 @@ type IndexedTransfer = {
...
     const history = useAccountHistory(address);
     const fetchAccountHistory = useFetchAccountHistory();
     const refresh = () => fetchAccountHistory(pubkey, true, true);
     const loadMore = () => fetchAccountHistory(pubkey, true);
-
-    const { tokenRegistry } = useTokenRegistry();
-
-    const mintDetails = React.useMemo(() => tokenRegistry.get(address), [address, tokenRegistry]);
+    const swrKey = ['get-token-info', address];
+    const { data: tokenInfo, isLoading: tokenInfoLoading } = useSWR(swrKey, () => getTokenInfo(pubkey, cluster, url))
```
steveluscher: Does passing a new fetching function on every render pass not break memoization? I think you can do this instead:

```ts
async function fetchTokenInfo([_, address, cluster, url]: ['get-token-info', string, string, string]) {
  return await getTokenInfo(new PublicKey(address), cluster, url);
}

export function TokenTransfersCard({ address }: { address: string }) {
    const { cluster, url } = useCluster();
    const { data: tokenInfo, isLoading: tokenInfoLoading } = useSWR(
        ['get-token-info', address, cluster, url],
        fetchTokenInfo,
     );
    /* ... */
}
```

app/utils/token-info.ts
```diff
@@ -58,6 +58,10 @@ function makeUtlClient(
     return new Client(config);
 }
 
+export function getTokenInfoSwrKey(address: string) {
+    return ['get-token-info', address];
```
steveluscher: See comments on previous commit.

app/components/transaction/TokenBalancesCard.tsx
```diff
@@ -38,9 +39,30 @@ export function TokenBalancesCard({ signature }: SignatureProps) {
...
+    const [tokenInfosLoading, setTokenInfosLoading] = useState(true);
+    const [tokenSymbols, setTokenSymbols] = useState<Map<string, string>>(new Map());
+
+    useEffect(() => {
+        const mints = rows.map(r => new PublicKey(r.mint));
+        getTokenInfos(mints, cluster, url).then(tokens => {
+            setTokenSymbols(new Map(tokens?.map(t => [t.address, t.symbol])));
+            setTokenInfosLoading(false);
+        });
+    }, [])
```
steveluscher:
```ts
    useAsyncEffect(async isMounted => {
        const mints = rows.map(r => new PublicKey(r.mint));
        const tokens = await getTokenInfos(mints, cluster, url);
        if (!isMounted()) {
            return;
        }
        setTokenSymbols(new Map(tokens?.map(t => [t.address, t.symbol])));
        setTokenInfosLoading(false);
    }, [])
```

app/components/common/Address.tsx
```diff
@@ -103,3 +117,23 @@ export const useTokenMetadata = (useMetadata: boolean | undefined, pubkey: strin
...
+        if (!info) {
+            getTokenInfo(new PublicKey(pubkey), cluster, url)
+                .then(token => {
+                    setInfo(token);
+                })
+                .catch(() => {
+                    setInfo(undefined);
+                });
+        }
+    }, [fetchTokenLabelInfo, pubkey, cluster, url, info, setInfo]);
```
steveluscher:
```ts
    useAsyncEffect(async isMounted => {
        if (!fetchTokenLabelInfo) return;
        if (!info) {
            try {
                const token = await getTokenInfo(new PublicKey(pubkey), cluster, url);
                if (!isMounted()) {
                    return;
                }
                setInfo(token);
            } catch {
                if (!isMounted()) {
                    return;
                }
                setInfo(undefined);
            }
        }
    }, [fetchTokenLabelInfo, pubkey, cluster, url, info, setInfo]);
```

next.config.mjs
mcintyre94: Ahh, damn. There's no single domain that the token logos returned by the solflare API are hosted at. It looks like the place this is used is the token holdings table, shall I just change that to ``? We already use plain `` for the token address page logo for example. 

 FWIW the current list of unique hosts returned by https://token-list-api.solana.cloud/v1/list is:
```
["assets.coingecko.com",
 "pscwbp7c3jss2hzt34p5y2atycb2vopk2x43iapvbr367pezd2nq.arweave.net",
 "gopromoted.com", "ipfs.io",
 "qdpw5kh2suzanv4jpry3jevawh7hmzqdewjfvg7ofuladhc7poaa.arweave.net",
 "imagedelivery.net", "www.vivendell.com", "user-images.githubusercontent.com",
 "pbs.twimg.com", "raw.githubusercontent.com", "www.fora.co", "pcarvalho.com",
 "storage.googleapis.com", "website-bucket.ams3.digitaloceanspaces.com",
 "api.fgraph.io", "cannabiznft.io", "static.particle.network",
 "www.dropbox.com", "nikkoagustino.my.id", "cdn.kaizen.finance",
 "solnetwebapistorage.blob.core.windows.net",
 "themetaversemuseum.s3.eu-west-3.amazonaws.com", "metaseer.io",
 "avatars.githubusercontent.com", "imgur.com",
 "bafkreifpmmgm3wp7e5u5hhnbxbq5djoh6zhickapb6e3yq2x473pn2w66q.ipfs.nftstorage.link",
 "i.redd.it", "assets.blockstars.gg", "discordapp.com",
 "dd6jvny7gn5elomvxk6uqibfvsdtpfiid7rxvparqrqaq4kz.arweave.net", "youngtech.fr",
 "bafkreidhiegzjjqnikoevl4fbuuocdrnvmxyy2bnnoeoha6kwykhgscuwm.ipfs.dweb.link",
 "bafybeib2znxf6whdtq354t6jsyejs7nlnevqd2pjxatr5ft2tgsgyrazb4.ipfs.nftstorage.link",
 "solthirsty.io", "www.solgoats.io", "komoverse.io",
 "bafybeidcqhtql7u6ppdgttwrx2e2qhb64mrvhqyilgks43msuhaisz3juu.ipfs.dweb.link",
 "cdn.eternaldragons.com", "metakitchen.io", "metaversol.com",
 "res.cloudinary.com", "www.protorealitygames.com", "rawcdn.githack.com",
 "www.arweave.net", "666serials.xyz", "w7.pngwing.com",
 "chiiepdxwskdrdtt3mqrnde3xh7hib4go22qacb5myxdj34xwwqq.arweave.net",
 "files.solcasino.io", "breezybunnies.com", "i.imgur.com", "arweave.net",
 "bafybeiav3sa2tcy4wmnju5hf7r3nskovfffqb5ffz4g4pshuwuwkpabmsu.ipfs.dweb.link",
 "ri6fdpqu6asrle3koykcepmdyd6m5eu2sfwb34vjfoxl2cvb.arweave.net",
 "www.panzerdogs.io", "mongomons.com", "github.com", "raw.githack.com",
 "mint.thearmadillos.io", "shop.alienchickenfarm.com", "gateway.pinata.cloud",
 "bafkreiefof3cjzgsd2xvfcijcf3r3az2rgycad7ycpveghjlmo6r5surh4.ipfs.dweb.link",
 "lizardtoken.xyz",
 "c5kh5ocamc7ov5pchbjdj5pikijb7toenkhpy345svfir6d7naoa.arweave.net", "i.ibb.co",
 "secureservercdn.net",
 "rxvwsth6k2wkjbpgqixi4db3ekvlailhvm2xa4pcmqzl6dtctqwa.arweave.net",
 "cdn.discordapp.com",
 "bafkreif2643iapwypek5ifmzwosd5uv3dtognfxy42br6d4c2k6nl3g6ci.ipfs.dweb.link",
 "artiftakt3d.s3.amazonaws.com", "dl.airtable.com",
 "theblockchainapi-website.s3.amazonaws.com", "proleague.gg",
 "zplphm7g437efyuli2psx6tmh4nf7ux6zf3iw65xnk5zjltaomca.arweave.net",
 "radrugs.io", "lootheroes.io",
 "bafybeie4ijaerva5fjqa4ojdy356jzf4xhhggybpwxf2ekn5tyg7xiedzu.ipfs.infura-ipfs.io",
 "www.synthetify.io", "static1.squarespace.com", "media.discordapp.net",
 "www.codegrepper.com", "neome.com", "www.turtles.com", "metadrago.art",
 "www.spookeletons.com", "devbeta.human.ag", "solgats.io",
 "static.wixstatic.com", "cdn.jsdelivr.net", "metavillage.app",
 "metakitchendev.onrender.com",
 "hv4gxzchk24cqfezebn3ujjz6oy2kbtztv5vghn6kpbkjc3vg4rq.arweave.net",
 "i.postimg.cc", "cdn.hngr.co",
 "toc24s2sbu3nkr65vejwjh2bmer3x3vo4xbhzcacaxr55ant3o2q.arweave.net",
 "wtcaeyantzogvbrzbwcgqfesfnz7mekovkyybs3ae54fptigq5za.arweave.net",
 "bafybeieax25iipx7k67ul3l5eeuzoisx5gohsxddrltuftuogktnrpbzgy.ipfs.infura-ipfs.io",
 "creator-hub-prod.s3.us-east-2.amazonaws.com", "assets.solmice.com",
 "media.zoker.com", "tokenhunters.app",
 "resource-branding-assets.s3.us-west-2.amazonaws.com", "media.kaijucards.io",
 "bafkreibp4wzfh5lrqr3gqlcduk4hk64qwmosyryhwrqjaa7qqcggyrylqe.ipfs.nftstorage.link"]
```

app/utils/tx.ts
mcintyre94: This gets cleaned up at the end and just becomes `addressLabel`, I just needed to keep both versions around as I refactored piece by piece

next.config.mjs
mcintyre94: I've just changed this to `` for now and reverted the change to config. I don't think we have any better options than that?

app/components/account/history/TokenTransfersCard.tsx
mcintyre94: Thanks - didn't realise you could get SWR to pass the key in like that!

mcintyre94:
On the topic of react-select, I've pushed a commit which refactors the searchto use their `<AsyncSelect>` component. This removes most of our state handling code around search, and instead we just have a simple function that async returns a list of options.

The main benefit of this for users is that it avoids an annoying race condition introduced by the token search where if the local search returned no results, then we showed either no results or loading, as we loaded the tokens. This meant that there was an annoying flash as you typed, as it flicked between token results and a loading state. I couldn't find an easy way to fix this in the current code and what we had for handling options was already getting pretty complex. `AsyncSelect` automatically keeps the current list of tokens displayed as it loads more, and handles the loading state for us. 

The downside is that local results are no longer shown immediately, they wait for the token search to resolve. But because the results are kept as you type, I think that in practice this is a better UX overall.

I did also find one annoying regression, if you start typing in search, blur it and then re-focus it, currently we will show the options for what you previously typed but `AsyncSelect` won't. I think this is a react-select bug, see https://github.com/JedWatson/react-select/issues/5714 

Also unrelated to this change, something in our `search-bar` CSS is breaking keyboard navigation of options which is annoying! It works nicely by default in all the react-select examples. I haven't figured out what yet though.

steveluscher:
Oooh. Is it possible, at all, to break the change from `react-select` to `AsyncSelect` into a separate PR, maybe _before_ this one? This is already a _lot_ of change, and now the changes are unrelated.

steveluscher (approved):
giphy

steveluscher:
> Also unrelated to this change, something in our search-bar CSS is breaking keyboard navigation of options which is annoying!

FTFY: e4aec85

mcintyre94:
I took another look at the original issue and one of the problems was that the Kin logo is wrong on our Explorer. I noticed that I still had this bug, because the token page was using the old CDN content if it existed (to keep extensions). I've updated that a bit so that it now fetches both old and new info in parallel, and if both exist then it uses as much data from the SDK as it can. The result is that [Kin has the correct logo](https://explorer-git-fork-mcintyre94-remove-token-registry-solana-labs.vercel.app/address/kinXdEcpDQeHPEuQnqmUgtYykqKGVFq6CeVX5iAHJq6) and [stuff with extensions still show them](https://explorer.solana.com/address/7vfCXTUXx5WJV5JADk17DUJ4ksgau7utNKj4b963voxs).

I also tweaked the loading of full token info to use SWR and pass its loading state down to the child components, which hold rendering until it's loaded. That avoids the flash of 'unknown token' you could get after the token account loaded and before the token info was fetched.

app/utils/token-info.ts
```diff
@@ -73,44 +73,57 @@ export async function getTokenInfo(
...
-    if (tokenListResponse.status >= 400) {
-        reportError(new Error('Error fetching token list from CDN'));
-        return undefined;
+    const [legacyCdnTokenInfo, sdkTokenInfo] = await Promise.all([
+        getFullLegacyTokenInfoUsingCdn(address, chainId),
+        getTokenInfo(address, cluster, connectionString)
+    ]);
+
+    if (!sdkTokenInfo) {
+        return legacyCdnTokenInfo;
```
steveluscher, line 110: You can actually get a little bit more perf out of this if you:

* Kick both off at the same time
* Only `await` the legacy one if the other one for sure fails.

```ts
// Both of these will be kicked off synchronously, because of the way promises work.
const legacyCdnTokenInfoPromise = getFullLegacyTokenInfoUsingCdn(address, chainId);
const sdkTokenInfo = await getTokenInfo(address, cluster, connectionString);

if (!sdkTokenInfo) {
  // Only ever await this if the sdkTokenInfo for sure turned up falsey.
  return await legacyCdnTokenInfoPromise;
}
```

mcintyre94: Hmm we use the legacy data on the other path as well though:

```ts
    return {
        address: sdkTokenInfo.address,
        chainId,
        decimals: sdkTokenInfo.decimals ?? legacyCdnTokenInfo?.decimals ?? 0,
        extensions: legacyCdnTokenInfo?.extensions,
        logoURI: sdkTokenInfo.logoURI ?? legacyCdnTokenInfo?.logoURI,
        name: sdkTokenInfo.name,
        symbol: sdkTokenInfo.symbol,
        tags
    }
```

Thinking about it more we could remove the fallback for decimals/logo, if you unset them on-chain, maybe we should respect that instead of using a legacy value. But while we use these extensions I think we're going to end up awaiting the `legacyCdnTokenInfo` on both return paths.

steveluscher: Oh, haha.

giphy-1

PR #286, opened 2023-08-21
Show token balance changes on a simulated transaction

brittcyr:
Currently, the simulation only shows whether or not the transaction succeeds and the logs.

It would be very useful for this to include token balance changes since the token program doesnt log the accounts or amounts. This should bring the displayed information of a simulation up to parity with what the explorer does for completed transactions.

The difficult part of this is that the simulateTransaction RPC spec does not include the pre/post token balances similar to asking for a completed transaction. To address this, we request the accountData for all accounts immediately before simulating and record the values of all token accounts. Then the simulateTransaction allows optional arguments for accounts to get their post simulation data. Parse that to get the post transaction token balance. Then convert it to the same format that the existing token balances component from the transaction page.

Example with token balances (before, the simulate was the same thing without the token changes component):
image


asktree (changes requested):
commented on telegram. this approach doesn't smell right to me, getting the explorer to show the right info should not be very complicated.

EDIT: 
OK, the problem is that I thought this was a governance-ui PR. now I understand my confusion. please ignore.

DonDuala:
@ngundotra can you review for us pls

mcintyre94 (changes requested):
Hey @brittcyr, thanks for contributing this, this is a great idea to add!

I do have one general concern/question about this approach, but I'm not sure there's a better alternative. In general for past transactions, the current balance isn't going to be an accurate pre-transaction balance. Do you have any thoughts about the impact of that?

I think this is going to be much more useful for a proposed transaction than for inspecting a past transaction. Do you think it'd make sense to disable this for the `/tx/<id>/inspect` page, and only use it when inspecting a transaction from base64? I'm curious what you're finding this most useful for though! And obviously nothing stops someone dumping the base64 of a past transaction, but it might be more obvious that it's not going to be accurate then. 

I've done some testing of this locally and really like this! I have found a few issues though:

- I think we should hide this when the simulation fails, because it looks like we get `null` accounts. Eg http://localhost:3000/tx/22urWLL1EntXLdjXqi9MYTCyxi3dVzxVhMPpgJ2fXP9t7FfyPX5zeTD5W9bZaqQmhPSPw6vRXXTCUujAyPir4nPG?cluster=devnet, if you inspect + simulate (http://localhost:3000/tx/22urWLL1EntXLdjXqi9MYTCyxi3dVzxVhMPpgJ2fXP9t7FfyPX5zeTD5W9bZaqQmhPSPw6vRXXTCUujAyPir4nPG/inspect?cluster=devnet) it’ll show a -5 change which isn’t correct.

- When I test with a tx with address table lookups (eg T3Bo3gmGMkpmPUEVKiQ7CWqU6HFNPce63TaPe8k59aDYa9p376VkyKjej1QCHE49m8h9erbhzeZtqX1171Wnw1K on mainnet) I’m getting an error: “Failed to get account keys because address table lookups were not resolved” I don’t get this on current explorer. Caused by the call to `getAccountKeys`: https://github.com/solana-labs/solana-web3.js/blob/8f5ba1b0fce3d4b60fee4a4aeb78a83d2c7b2b93/src/message/v0.ts#L78

- Please run `pnpm lint --fix` to fix the CI lint errors

app/components/inspector/SimulatorCard.tsx
```diff
@@ -81,11 +92,97 @@ function useSimulator(message: VersionedMessage) {
...
+                const tokenAccountKeys: ParsedMessageAccount[] = [];
+
+                for (let index = 0; index < accountKeys.length; index++) {
+                    const key = accountKeys[index];
+                    const parsedAccountPre = parsedAccountsPre.value[index];
+                    const accountDataPost = resp.value.accounts?.at(index)?.data[0];
+                    const accountOwnerPost = resp.value.accounts?.at(index)?.owner;
+
+                    if (
+                        parsedAccountPre?.owner.toBase58() == TOKEN_PROGRAM_ID.toBase58() &&
```
mcintyre94: I think we should include the token22 program here too. Eg test with a BERN transaction (mint is CKfatsPMUf8SkiURsDXs7eK6GWb4Jsd6UDbs7twMCWxo).

But note these all seem to use address table lookups, so we probably won't be able to test one yet

app/components/inspector/SimulatorCard.tsx
```diff
@@ -81,11 +92,97 @@ function useSimulator(message: VersionedMessage) {
...
+                            mint: mint,
+                            owner: owner,
+                            uiTokenAmount: tokenAmount,
+                        };
+                        preTokenBalances.push(preTokenBalance);
+                    }
+
+                    if (
+                        accountOwnerPost === TOKEN_PROGRAM_ID.toBase58() &&
+                        Buffer.from(accountDataPost!, 'base64').length === 165
```
mcintyre94: Same deal with token22 program, and I think it'd be >= 165 length then

brittcyr:
> Hey @brittcyr, thanks for contributing this, this is a great idea to add!
> 
> I do have one general concern/question about this approach, but I'm not sure there's a better alternative. In general for past transactions, the current balance isn't going to be an accurate pre-transaction balance. Do you have any thoughts about the impact of that?
> 
> I think this is going to be much more useful for a proposed transaction than for inspecting a past transaction. Do you think it'd make sense to disable this for the `/tx/<id>/inspect` page, and only use it when inspecting a transaction from base64? I'm curious what you're finding this most useful for though! And obviously nothing stops someone dumping the base64 of a past transaction, but it might be more obvious that it's not going to be accurate then.
> 
> I've done some testing of this locally and really like this! I have found a few issues though:
> 
> * I think we should hide this when the simulation fails, because it looks like we get `null` accounts. Eg http://localhost:3000/tx/22urWLL1EntXLdjXqi9MYTCyxi3dVzxVhMPpgJ2fXP9t7FfyPX5zeTD5W9bZaqQmhPSPw6vRXXTCUujAyPir4nPG?cluster=devnet, if you inspect + simulate (http://localhost:3000/tx/22urWLL1EntXLdjXqi9MYTCyxi3dVzxVhMPpgJ2fXP9t7FfyPX5zeTD5W9bZaqQmhPSPw6vRXXTCUujAyPir4nPG/inspect?cluster=devnet) it’ll show a -5 change which isn’t correct.
> * When I test with a tx with address table lookups (eg T3Bo3gmGMkpmPUEVKiQ7CWqU6HFNPce63TaPe8k59aDYa9p376VkyKjej1QCHE49m8h9erbhzeZtqX1171Wnw1K on mainnet) I’m getting an error: “Failed to get account keys because address table lookups were not resolved” I don’t get this on current explorer. Caused by the call to `getAccountKeys`: https://github.com/solana-labs/solana-web3.js/blob/8f5ba1b0fce3d4b60fee4a4aeb78a83d2c7b2b93/src/message/v0.ts#L78
> * Please run `pnpm lint --fix` to fix the CI lint errors

I am not targeting past transactions. I am not actually aware of anyone using the simulate other than linking from realms proposal which is the use case I am desiring. Disabling it on past tx inspect seems right. Would be a lot of work and outside the scope of this to get the pre-balance for that use case.

Other comments seem doable, will address them.

brittcyr:
Not sure what to do about the address lookup tables since I need getAccountKeys before simulation. Should I just skip the token balance changes in that case?

mcintyre94:
> I am not targeting past transactions. I am not actually aware of anyone using the simulate other than linking from realms proposal which is the use case I am desiring. 

Got you, makes a lot of sense for a proposal for sure! I think just hide it when we know it's a past transaction, or when simulation fails. This should work really nicely then!

> Not sure what to do about the address lookup tables since I need getAccountKeys before simulation. Should I just skip the token balance changes in that case?

I'm not really familiar with this web3js API TBH. But can you use `message.resolveAddressLookupTables`? https://solana-labs.github.io/solana-web3.js/classes/MessageV0.html#resolveAddressTableLookups Might need to look more into this though!

brittcyr:
> * T3Bo3gmGMkpmPUEVKiQ7CWqU6HFNPce63TaPe8k59aDYa9p376VkyKjej1QCHE49m8h9erbhzeZtqX1171Wnw1K

do you have an example of address table lookups on a base64 inspector tx?

mcintyre94:
You can use that transaction as base64 on mainnet:

ARZ0M+KuIaQSPS6j1Kf4KfPLVV/Xu7Dl1CCThwPzUdpkYC6EGcTkKKc8CTatRgqYcUkbPAstWzU47Xr7yL/ZowKAAQAHC3maBMLkV5LzuqX/7ToBxoQwBelW83WQqHLXCeu5fYQFCKtORSAAGe3iGWbVj87K0JVp8eHr1TqLwMywg6BcjtwKBSRBoVhOV4xwBfanvh4FdBl3V3APJBkA8l9K/YoZzg1kTL4ts5oE1wHY1u+a8zoDushoiLFircTVAC+fC6vwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADBkZv5SEXMv/srbpyw5vnvIzlu8X3EmssQ5s6QAAAAAR51VvyMcBu7nTFbs5oFQf9sbLeo/SOUQKxzaJWvBOPBt324ddloZPZy+FGzut5rBy0he1fWzeROoz1hX7/AKmMlyWPTiSJ8bs9ECkUjg2DC1oTmdr/EIQEjnvY2+n4WbQ/+if11/ZKdMCbHylYed5LCas238ndUUsyGqezjOXo4vh36D2uxoiPQ6u1rnsxG7eMNJYIl/zlp7gsFuIP8Ze+60iy4dQ9F7P+H/r1Zn6taV0QqzcndA08lQfcZuuSNAQFAAUCwFwVAAgGAAEADwQHAQEGHAcKAAMDAgERDwYQCQYTEhQAAw4NAgsMEQ8QBxAkwSCbM0HWnIEAAQAAAB5kAAEkBwhFAAAAACENnBAAAAAAMgAABwMBAAABCQH9QKy7pCQfVqHZ8aPpnAIxlPWQsH0NTGkkpgg1woyRXgTGx8XEBjWuyMLBww==

brittcyr:
Handled address lookup tables and made the token balances not show up with an error.

DonDuala:
Bump @mcintyre94

app/components/inspector/SimulatorCard.tsx
```diff
@@ -49,15 +53,22 @@ export function SimulatorCard({ message }: { message: VersionedMessage }) {
...
+                </div>
+                <ProgramLogsCardBody message={message} logs={logs} cluster={cluster} url={url} />
             </div>
-            <ProgramLogsCardBody message={message} logs={logs} cluster={cluster} url={url} />
-        </div>
+            {showTokenBalanceChanges && simulationTokenBalanceRows && !simulationError && simulationTokenBalanceRows.rows.length && (
+                <div className="card">
+                    <TokenBalancesCardInner rows={simulationTokenBalanceRows.rows} />
+                </div>
+            )}
```
mcintyre94:
```suggestion
            {showTokenBalanceChanges && simulationTokenBalanceRows && !simulationError && simulationTokenBalanceRows.rows.length ? (
                <div className="card">
                    <TokenBalancesCardInner rows={simulationTokenBalanceRows.rows} />
                </div>
            ) : null}
```

I think this needs to be using a ternary. On a transaction with no simulation rows (but also no error), for example `Afs79yHxq/0meJcx8g8Ii5TgxqqzKiRfpn7PFKZjKK82h0zUovJyqjlCQ4LZlY1qyJxUiHfrk+PWlmR+0n1cRAyAAQADB+Stj1FYmqtHlZI/QGmQeYoR1irFeTLBI2sp7Gc6ug7MuCuUoNYwatIHcVDwZphNsSdPcn3cWj7QDruFzkL8aV0PqJUEtyjqPaTVYdCQwoy3TlS7gGgNtvZHVS5ljgMmfg7CBa/RYA94N/8Dm0+91HOmbMuq3k5VX9oBUUprCZcLAwZGb+UhFzL/7K26csOb57yM5bvF9xJrLEObOkAAAAAIQcuVuAJV1WUsDbUNQYARXuUf1y8xSDmekMHgzfF4Tozl/UQw28pzeUpQzoB+d1FwIc8XGJJMTrf86rvH9XywCMULIemm/r4y+W1vlqddqaEK5+j2ZJjP3MHekr417mUMBAAJA+gDAAAAAAAABQoAEgETFAIHCAkKCYu+5vlNoM4EAQUZEhUBABMWFAIXBwsKCAkMDQ4PDBgZGhAGAy1b9mAHNRbq4WipWwUGAAAACwAAAAAAAAABBAFu9OSOigEAAAEDAAAAU0RLAAEFGRIVAQATFhQCFwcLCggJDw0ODwwYGRoRBgMtW/ZgBzUW6uFwj000BgAAAAsAAAAAAAAAAgQBb/TkjooBAAABAwAAAFNESwABBRkSFQEAExYUAhcHCwoICQwNDg8MGBkaEAYDLVv2YAc1FurhGLum/QUAAAAPAAAAAAAAAAEEAXD05I6KAQAAAQMAAABTREsAAQUZEhUBABMWFAIXBwsKCAkPDQ4PDBgZGhEGAy1b9mAHNRbq4XiYPjwGAAAADwAAAAAAAAACBAFx9OSOigEAAAEDAAAAU0RLAAEFGRIVAQATFhQCFwcLCggJDA0ODwwYGRoQBgMtW/ZgBzUW6uEIQMz5BQAAABMAAAAAAAAAAQQBcvTkjooBAAABAwAAAFNESwABBRkSFQEAExYUAhcHCwoICQ8NDg8MGBkaEQYDLVv2YAc1FurhCBs3QAYAAAATAAAAAAAAAAIEAXP05I6KAQAAAQMAAABTREsAAQUZEhUBABMWFAIXBwsKCAkMDQ4PDBgZGhAGAy1b9mAHNRbq4eDI8fUFAAAAFgAAAAAAAAABBAF09OSOigEAAAEDAAAAU0RLAAEFGRIVAQATFhQCFwcLCggJDw0ODwwYGRoRBgMtW/ZgBzUW6uGYnS9EBgAAABYAAAAAAAAAAgQBdfTkjooBAAABAwAAAFNESwABBRkSFQEAExYUAhcHCwoICQwNDg8MGBkaEAYDLVv2YAc1FurhuFEX8gUAAAAaAAAAAAAAAAEEAXb05I6KAQAAAQMAAABTREsAAQUZEhUBABMWFAIXBwsKCAkPDQ4PDBgZGhEGAy1b9mAHNRbq4SggKEgGAAAAGgAAAAAAAAACBAF39OSOigEAAAEDAAAAU0RLAAEBYQWxzthEkoKua2gEhVWJDudkqzKq8gXk22w8tDHQoVQLHB8gHh0kISIjJSYJAAIEAQMFCQoG` on mainnet, I'm getting a 0 rendered:

Screenshot 2023-09-13 at 16 55 07

mcintyre94 (commented):
Sorry for the delay @brittcyr and thanks for the bump @DonDuala! This is really close now :)

I've made a suggestion where I think we need to use a ternary

In addition I'm seeing a slight styling issue at the bottom of the Token Balances card - do you know where that might be coming from?
Screenshot 2023-09-13 at 16 49 09

brittcyr:
> Sorry for the delay @brittcyr and thanks for the bump @DonDuala! This is really close now :)
> 
> I've made a suggestion where I think we need to use a ternary
> 
> In addition I'm seeing a slight styling issue at the bottom of the Token Balances card - do you know where that might be coming from? Screenshot 2023-09-13 at 16 49 09

Issue was a card inside a card. Fixed

mcintyre94:
This looks great! Thanks for the contribution and for your patience, @brittcyr!

mcintyre94 (approved):

PR #288, opened 2023-08-24
fix: check UTL verifiedness of tokens before showing spoof warning banner

steveluscher:
fix: check UTL verifiedness of tokens before showing spoof warning banner
  
  # Summary
  
  The UTL provides a `verfied` property. Let's not show a token name/icon spoof warning banner unless that property is `false`.
  
  
  # Test Plan
  
  image
  
  http://localhost:3000/address/2VhjJ9WxaGC3EZFwJG9BDUs9KxKCAjQY4vgd1qxgYWVg/transfers
  
  image
  
  http://localhost:3000/address/DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263
  
  image
  
  http://localhost:3000/address/FASTAr1in6u53vykPbCuFTUZnGdiLJrXaTEQ7UAArdBm

### DIFF
--- a/app/utils/token-info.ts
+++ b/app/utils/token-info.ts
@@ -31,6 +31,9 @@ export type FullLegacyTokenInfo = {
     readonly tags?: string[],
     readonly extensions?: TokenExtensions,
 }
+export type FullTokenInfo = FullLegacyTokenInfo & {
+    readonly verified: boolean;
+};
 
 type FullLegacyTokenInfoList = {
     tokens: FullLegacyTokenInfo[]
@@ -97,7 +100,7 @@ export async function getFullTokenInfo(
     address: PublicKey,
     cluster: Cluster,
     connectionString: string
-): Promise<FullLegacyTokenInfo | undefined> {
+): Promise<FullTokenInfo | undefined> {
     const chainId = getChainId(cluster);
     if (!chainId) return undefined;
 
@@ -107,7 +110,12 @@ export async function getFullTokenInfo(
     ]);
 
     if (!sdkTokenInfo) {
-        return legacyCdnTokenInfo;
+        return legacyCdnTokenInfo
+            ? {
+                  ...legacyCdnTokenInfo,
+                  verified: true,
+              }
+            : undefined;
     }
 
     // Merge the fields, prioritising the sdk ones which are more up to date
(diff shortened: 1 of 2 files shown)


app/utils/token-info.ts
```diff
@@ -107,7 +110,12 @@ export async function getFullTokenInfo(
     ]);
 
     if (!sdkTokenInfo) {
-        return legacyCdnTokenInfo;
+        return legacyCdnTokenInfo
+            ? {
+                  ...legacyCdnTokenInfo,
+                  verified: true,
```
steveluscher, line 116: @mcintyre94, I'm presuming that _everything_ that comes from the legacy token list CDN can be presumed to be ‘verified.’ Does that make sense?

app/utils/token-info.ts
```diff
@@ -123,8 +131,9 @@ export async function getFullTokenInfo(
         logoURI: sdkTokenInfo.logoURI ?? undefined,
         name: sdkTokenInfo.name,
         symbol: sdkTokenInfo.symbol,
-        tags
-    }
+        tags,
+        verified: !!(legacyCdnTokenInfo || sdkTokenInfo.verified),
```
steveluscher, line 135: This means ‘if you're on the legacy token list, or the UTL registry said you're verified, you're verified.’ Is that sound, @mcintyre94?

askibin (approved):

app/utils/token-info.ts
mcintyre94: Sorry, I missed this PR before! I think this is reasonable, but if the UTL registry contains the same token as unverified maybe that should take precedence?

So:
- Not on the legacy token list at all, UTL verified determines it
- On legacy token list and UTL, UTL verified determines it
- On legacy token list and not UTL, verified

steveluscher: So, if I understand correctly, that basically implies that we should do this?
```suggestion
        verified: sdkTokenInfo.verified ?? false,
```

1. By this line, we know that `sdkTokenInfo` is non-null.
2. The ‘not on UTL but on legacy list’ case is handled above.

steveluscher: Building a Vercel preview for that now: #289.

PR #292, opened 2023-09-15
chore: delete Sentry

steveluscher:
chore: delete Sentry

# Summary

One of the central tenets of measurement is this question: are you prepared to change what you work on based on the outcome of the data.

Explorer is not under the kind of active development that requires deep insights into runtime data at this time. Put another way, I personally have never performed work on the Explorer based on a result that I observed in Sentry.

Until this changes, let's get rid of it.


steveluscher:
This seems to have taken first-load JS on the homepage from 217 kB to 171kB.

Lighthouse score is approaching 90%.
image

PR #299, opened 2023-09-28
Updating the canonical reference for the Memo program

jnwng:
The legacy version of the Memo program (`Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo`) is referred to as Memo v1,
whereas the latest version of the Memo program (`MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr`) is more canonically known simply as `Memo`: https://spl.solana.com/memo

came up in the context of an upcoming wallet integration which noted that the newer version of the Memo program isn't prototypically referred to as Memo v2 but just Memo


steveluscher (approved):
aidan-gallagher-friends

PR #311, opened 2023-12-13
Integrate basic token-2022 support

joncinque:
#### Problem

Token-2022 has been around for awhile, but the explorer still has no knowledge of it!

#### Summary of changes

This is really just basic support, and it was done in a pretty simple manner:

* look for instances of `"spl-token"`, which is the program name returned from JSON-parsed encoding, and make sure that it also supports `"spl-token-2022"`
* look for instances of `TOKEN_PROGRAM_ID`, and make sure that it also supports `TOKEN_2022_PROGRAM_ID`

So in the end, accounts and mints are properly displayed, and the additional token-2022 accounts are fetched too.

There is no parsing of extensions included in this PR, but it can certainly be done in follow-up work.

### DIFF
--- a/app/address/[address]/attributes/page-client.tsx
+++ b/app/address/[address]/attributes/page-client.tsx
@@ -2,6 +2,7 @@
 
 import { MetaplexNFTAttributesCard } from '@components/account/MetaplexNFTAttributesCard';
 import { ParsedAccountRenderer } from '@components/account/ParsedAccountRenderer';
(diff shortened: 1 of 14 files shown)


mcintyre94 (approved):
Looks good! We need to figure out how to get these token-2022 tokens into search and our token lookup though!

joncinque:
Yeah for sure! The emphasis for this PR should be on the word "basic" :sweat_smile:

PR #312, opened 2023-12-13
token-account: Show delegate and delegated amount

joncinque:
#### Problem

Token accounts can have delegates with delegated amounts, but the explorer doesn't show that information.

#### Solution

Also show them!

Here's a wrapped SOL version:

image

And a normal token version:

image


mcintyre94 (approved):
Nice, looks good!

PR #315, opened 2024-01-19
token-2022: Show all extensions

joncinque:
#### Problem

Although the explorer has some support of token-2022, it doesn't show any mint or account extensions, which leads to a lot of confusion for people who create mints.

#### Solution

Everything is provided with JSON-parsed encoding, so:

* create types for all of the JSON-parsed returns
* show each extension

For the most part, each commit corresponds to an extension, except otherwise noted. I wasn't sure about the best way to display all of these, especially those with many fields, so I went with a little header line before showing each field. Let me know what you think of that.

Here are some screenshots of a mint with every extension (minus token group, group member, group pointer, and member pointer):

image

image

image

And an account with all extensions:

image

image

And in case you care, here's how you can create your own mega-token:

```
set -ex
solana-keygen -o mint.json --no-passphrase
mint=$(solana-keygen pubkey mint.json)
solana-keygen -o source.json --no-passphrase
source_pubkey=$(solana-keygen pubkey source.json)
spl-token -ul \
  create-token \
  --enable-close \
  --enable-freeze \
  --enable-metadata \
  --enable-non-transferable \
  --enable-permanent-delegate \
  --enable-confidential-transfers auto \
  --default-account-state initialized \
  --interest-rate 5 \
  --transfer-fee 10 11 \
  --transfer-hook $mint \
  --program-id TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb \
  mint.json
spl-token -ul initialize-metadata $mint MegaToken MT https://spl.solana.com/token-2022
spl-token -ul update-metadata $mint Mega Token

spl-token -ul create-account --immutable $mint source.json
spl-token -ul enable-required-transfer-memos "$source_pubkey"
spl-token -ul enable-cpi-guard "$source_pubkey"
spl-token -ul configure-confidential-transfer-account --address "$source_pubkey"
```

### DIFF
--- a/app/components/account/TokenAccountSection.tsx
+++ b/app/components/account/TokenAccountSection.tsx
@@ -1017,7 +1017,7 @@ function TokenExtensionRows(decimals: number, tokenExtension: TokenExtension) {
             const extension = create(tokenExtension.state, TransferHookAccount);
             return (
                 <tr>
-                    <td>Transfer Hook Account</td>
+                    <td>Transfer Hook Status</td>
                     <td className="text-lg-end">{!extension.transferring && 'not '}transferring</td>
                 </tr>
             );
--- a/app/components/account/TokenAccountSection.tsx
+++ b/app/components/account/TokenAccountSection.tsx
@@ -13,7 +13,7 @@ import { displayTimestampWithoutDate } from '@utils/date';
 import { abbreviatedNumber, normalizeTokenAmount } from '@utils/index';
 import { addressLabel } from '@utils/tx';
 import { MintAccountInfo, MultisigAccountInfo, TokenAccount, TokenAccountInfo } from '@validators/accounts/token';
-import { MintCloseAuthority, TokenExtension } from '@validators/accounts/token-extension';
+import { MintCloseAuthority, TokenExtension, TransferFeeAmount } from '@validators/accounts/token-extension';
 import { BigNumber } from 'bignumber.js';
 import { useEffect, useMemo, useState } from 'react';
 import { ExternalLink, RefreshCw } from 'react-feather';
@@ -549,7 +549,19 @@ function TokenExtensionRows(mintInfo: MintAccountInfo, tokenExtension: TokenExte
                 return <></>;
             }
         }
-        case 'transferFeeAmount':
+        case 'transferFeeAmount': {
+            const extension = create(tokenExtension.state, TransferFeeAmount);
+            return (
+                <tr>
+                    <td>Withheld Amount</td>
+                    <td className="text-lg-end">
+                        {normalizeTokenAmount(extension.withheldAmount, mintInfo.decimals).toLocaleString('en-US', {
+                            maximumFractionDigits: 20,
+                        })}
+                    </td>
+                </tr>
+            );
+        }
         case 'transferFeeConfig':
         case 'confidentialTransferMint':
         case 'confidentialTransferAccount':
--- a/app/components/account/TokenAccountSection.tsx
+++ b/app/components/account/TokenAccountSection.tsx
@@ -13,7 +13,12 @@ import { displayTimestampWithoutDate } from '@utils/date';
 import { abbreviatedNumber, normalizeTokenAmount } from '@utils/index';
 import { addressLabel } from '@utils/tx';
 import { MintAccountInfo, MultisigAccountInfo, TokenAccount, TokenAccountInfo } from '@validators/accounts/token';
-import { MintCloseAuthority, TokenExtension, TransferFeeAmount } from '@validators/accounts/token-extension';
+import {
+    MintCloseAuthority,
+    TokenExtension,
+    TransferFeeConfig,
+    TransferFeeAmount,
+} from '@validators/accounts/token-extension';
 import { BigNumber } from 'bignumber.js';
 import { useEffect, useMemo, useState } from 'react';
 import { ExternalLink, RefreshCw } from 'react-feather';
@@ -562,7 +567,78 @@ function TokenExtensionRows(mintInfo: MintAccountInfo, tokenExtension: TokenExte
                 </tr>
             );
         }
-        case 'transferFeeConfig':
+        case 'transferFeeConfig': {
+            const extension = create(tokenExtension.state, TransferFeeConfig);
+            return (
+                <>
+                    <tr>
+                        <h4>Transfer Fee Config</h4>
+                    </tr>
+                    {extension.transferFeeConfigAuthority && (
+                        <tr>
+                            <td>Transfer Fee Authority</td>
+                            <td className="text-lg-end">
+                                <Address pubkey={extension.transferFeeConfigAuthority} alignRight link />
+                            </td>
+                        </tr>
+                    )}
+                    <tr>
+                        <td>Older Fee Epoch</td>
+                        <td className="text-lg-end">{extension.olderTransferFee.epoch}</td>
+                    </tr>
+                    <tr>
+                        <td>Older Maximum Fee</td>
+                        <td className="text-lg-end">
+                            {normalizeTokenAmount(
+                                extension.olderTransferFee.maximumFee,
+                                mintInfo.decimals
+                            ).toLocaleString('en-US', {
+                                maximumFractionDigits: 20,
+                            })}
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Older Fee Basis Points</td>
+                        <td className="text-lg-end">{extension.olderTransferFee.transferFeeBasisPoints}</td>
+                    </tr>
+                    <tr>
+                        <td>Newer Fee Epoch</td>
+                        <td className="text-lg-end">{extension.newerTransferFee.epoch}</td>
+                    </tr>
+                    <tr>
+                        <td>Newer Maximum Fee</td>
+                        <td className="text-lg-end">
+                            {normalizeTokenAmount(
+                                extension.newerTransferFee.maximumFee,
+                                mintInfo.decimals
+                            ).toLocaleString('en-US', {
+                                maximumFractionDigits: 20,
+                            })}
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Newer Fee Basis Points</td>
+                        <td className="text-lg-end">{extension.newerTransferFee.transferFeeBasisPoints}</td>
+                    </tr>
+                    {extension.withdrawWithheldAuthority && (
+                        <tr>
+                            <td>Withdraw Withheld Fees Authority</td>
+                            <td className="text-lg-end">
+                                <Address pubkey={extension.withdrawWithheldAuthority} alignRight link />
+                            </td>
+                        </tr>
+                    )}
+                    <tr>
+                        <td>Withheld Amount</td>
+                        <td className="text-lg-end">
+                            {normalizeTokenAmount(extension.withheldAmount, mintInfo.decimals).toLocaleString('en-US', {
+                                maximumFractionDigits: 20,
+                            })}
+                        </td>
+                    </tr>
+                </>
+            );
+        }
         case 'confidentialTransferMint':
         case 'confidentialTransferAccount':
         case 'defaultAccountState':
--- a/app/components/account/TokenAccountSection.tsx
+++ b/app/components/account/TokenAccountSection.tsx
@@ -13,6 +13,7 @@ import { displayTimestampWithoutDate } from '@utils/date';
 import { abbreviatedNumber, normalizeTokenAmount } from '@utils/index';
 import { addressLabel } from '@utils/tx';
 import { MintAccountInfo, MultisigAccountInfo, TokenAccount, TokenAccountInfo } from '@validators/accounts/token';
+import { TokenExtension } from '@validators/accounts/token-extension';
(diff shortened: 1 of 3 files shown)


luke-truitt:
pls merge

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+        }
+        case 'transferFeeAmount': {
+            const extension = create(tokenExtension.state, TransferFeeAmount);
+            return (
+                <tr>
+                    <td>Withheld Amount</td>
+                    <td className="text-lg-end">
+                        {normalizeTokenAmount(extension.withheldAmount, decimals).toLocaleString('en-US', {
+                            maximumFractionDigits: 20,
+                        })}
```
mcintyre94, line 633: I think we should use the token symbol here to show the unit is in tokens, and not eg SOL. Is `withheldAmount` a hardcoded fee per transfer?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                    {extension.transferFeeConfigAuthority && (
+                        <tr>
+                            <td>Transfer Fee Authority</td>
+                            <td className="text-lg-end">
+                                <Address pubkey={extension.transferFeeConfigAuthority} alignRight link />
+                            </td>
+                        </tr>
+                    )}
+                    <tr>
+                        <td>Older Fee Epoch</td>
```
mcintyre94: I think it'd be useful to highlight which of the old and new fees is active somehow, since the account page doesn't display the epoch and you need that to know which is used. Maybe just labelling one of them (current)?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                        <td className="text-lg-end">{extension.olderTransferFee.epoch}</td>
+                    </tr>
+                    <tr>
+                        <td>Older Maximum Fee</td>
+                        <td className="text-lg-end">
+                            {normalizeTokenAmount(extension.olderTransferFee.maximumFee, decimals).toLocaleString(
```
mcintyre94, line 663: Again I think this is in token units, so adding the symbol would help

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                                'en-US',
+                                {
+                                    maximumFractionDigits: 20,
+                                }
+                            )}
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Older Fee Basis Points</td>
+                        <td className="text-lg-end">{extension.olderTransferFee.transferFeeBasisPoints}</td>
```
mcintyre94: Should we convert basis points to %?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                            <td className="text-lg-end">{extension.withdrawWithheldAuthorityElgamalPubkey}</td>
+                        </tr>
+                    )}
+                    <tr>
+                        <td>Harvest to Mint</td>
+                        <td className="text-lg-end">{extension.harvestToMintEnabled ? 'enabled' : 'disabled'}</td>
+                    </tr>
+                    <tr>
+                        <td>Withheld Amount</td>
+                        <td className="text-lg-end">{extension.withheldAmount}</td>
```
mcintyre94, line 771: Append symbol?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                    <td>DefaultAccountState</td>
+                    <td className="text-lg-end">{extension.accountState}</td>
+                </tr>
+            );
+        }
+        case 'nonTransferable': {
+            return (
+                <tr>
+                    <td>Non-Transferable</td>
+                    <td className="text-lg-end">enabled</td>
```
mcintyre94, line 789: These rows without a heading are a bit confusing because they end up under the last heading added even though they're not related to it. Maybe move them all to come first so they're before any headings/groups?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                        <tr>
+                            <td>Authority</td>
+                            <td className="text-lg-end">
+                                <Address pubkey={extension.rateAuthority} alignRight link />
+                            </td>
+                        </tr>
+                    )}
+                    <tr>
+                        <td>Current Rate Basis Points</td>
+                        <td className="text-lg-end">{extension.currentRate}</td>
```
mcintyre94: Convert to %?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                    </tr>
+                    <tr>
+                        <td>Pre-Current Average Rate</td>
+                        <td className="text-lg-end">{extension.preUpdateAverageRate}</td>
```
mcintyre94: Is this a %?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                    </tr>
+                    <tr>
+                        <td>Last Update Timestamp</td>
+                        <td className="text-lg-end">{extension.lastUpdateTimestamp}</td>
+                    </tr>
+                    <tr>
+                        <td>Initialization Timestamp</td>
+                        <td className="text-lg-end">{extension.initializationTimestamp}</td>
```
mcintyre94: For this + lastUpdateTimestamp it'd be nice to convert to a readable date. We probably have a helper for that already

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                return (
+                    <tr>
+                        <td>Permanent Delegate</td>
+                        <td className="text-lg-end">
+                            <Address pubkey={extension.delegate} alignRight link />
+                        </td>
+                    </tr>
+                );
+            } else {
+                return <></>;
```
mcintyre94, line 839: Would it be worth putting a row in with Permanent Delegate and then Not Set or similar? Might be useful to know the extension is there but the delegate isn't set. Similar to what you have with CPI guard

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                                <ExternalLink className="align-text-top ms-2" size={13} />
+                            </a>
+                        </td>
+                    </tr>
+                    {extension.additionalMetadata?.map(keyValuePair => (
+                        <tr key="{keyValuePair[0]}">
+                            <td>{keyValuePair[0]}</td>
+                            <td className="text-lg-end">{keyValuePair[1]}</td>
+                        </tr>
+                    ))}
```
mcintyre94: I'm a bit concerned about using a field set by the token as a top-level field, could we put a label "Additional Metadata" or similar above these? I'm thinking about someone creating a token with fields like "Official" or "Jupiter Airdrop" on scam tokens and us displaying them the same way we do other fields with no indication.

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+                    <td className="text-lg-end">{extension.requireIncomingTransferMemos ? 'enabled' : 'disabled'}</td>
+                </tr>
+            );
+        }
+        case 'transferHookAccount': {
+            const extension = create(tokenExtension.state, TransferHookAccount);
+            return (
+                <tr>
+                    <td>Transfer Hook Account</td>
+                    <td className="text-lg-end">{!extension.transferring && 'not '}transferring</td>
```
mcintyre94, line 1079: More of a question, how come this is labelled "Account" when it just seems to be a boolean?

app/components/account/TokenAccountSection.tsx
```diff
@@ -497,3 +552,509 @@ function MultisigAccountCard({ account, info }: { account: Account; info: Multis
...
+            );
+        }
+        case 'tokenGroupMember': {
+            return (
+                <tr>
+                    <td>Group Member</td>
+                    <td className="text-lg-end">unparseable</td>
+                </tr>
+            );
+        }
```
mcintyre94, line 1157: How come this and `tokenGroup` are unparseable? The types in token-extension.ts look displayable

mcintyre94 (approved):
Looks good!

I think longer term we might want to change how these are displayed because I think this is pushing the limits of the table really. Maybe an extensions tab or something?

I don't think any of my comments are blockers, they're mostly just nitpicks about how we're displaying certain values

I would flag the 'Additional Metadata' though because I think there's a real risk there of malicious tokens abusing it. Maybe the easiest fix for now is to make metadata the last extension displayed, and add a header above Additional Metadata?

app/components/account/TokenAccountSection.tsx
joncinque: Sure! Done in 04f8981

app/components/account/TokenAccountSection.tsx
joncinque: Yeah that makes sense, done in 0210f70

app/components/account/TokenAccountSection.tsx
joncinque: Done in 04f8981

app/components/account/TokenAccountSection.tsx
joncinque: Done in ed26eadc48e75324c7cc612930908cc78560d9aa

app/components/account/TokenAccountSection.tsx
joncinque: Done in 04f8981

app/components/account/TokenAccountSection.tsx
joncinque: Sure thing, done in de3404c

app/components/account/TokenAccountSection.tsx
joncinque: Done in ed26eadc48e75324c7cc612930908cc78560d9aa

app/components/account/TokenAccountSection.tsx
joncinque: Yep, updated to percentage, done in ed26eadc48e75324c7cc612930908cc78560d9aa

app/components/account/TokenAccountSection.tsx
joncinque: Good call, done in 64920d6

app/components/account/TokenAccountSection.tsx
joncinque: If the permanent delegate is unset, it becomes impossible to ever set it again, similar to what we do with mint authority or freeze authority, so for consistency with everything else, I think it would be better to not show it.

CPI Guard is different because you can always toggle it

app/components/account/TokenAccountSection.tsx
joncinque: Yeah sure, great idea! Done in 82c8ba4

app/components/account/TokenAccountSection.tsx
joncinque: Good point, this was a bit lazy. How about "Transfer Hook Status"? It's just a flag saying if the account is in the middle of a transfer or not. So unless there's a big bug, it should always show it's not transferring.

I can just remove showing it totally, but people probably want to see if their account has a transfer hook. Anyway, I changed it to "Transfer Hook Status" in 4b6a068

app/components/account/TokenAccountSection.tsx
joncinque: Entirely out of laziness, done in 62f21de

app/components/account/TokenAccountSection.tsx
mcintyre94: Transfer Hook Status sounds good!

app/components/account/TokenAccountSection.tsx
mcintyre94: Not sure why these aren't working for me on the preview, but not a blocker

joncinque: I think it's because there's no symbol defined in the `tokenInfo`

mcintyre94: Ah I was confused because it shows symbol: MT somewhere else
But this looks like it's working on other tokens now it's deployed, so all good!

PR #320, opened 2024-02-05
[anchor] display bytes as hex

guibescos:
Current the Anchor `bytes` type (`Vec<u8>` in rust) gets represented very poorly as it's interpreted as a utf-8 string. 
Instead it should be represented as a hex payload similar to serialized instruction data.

Current : image
https://explorer.solana.com/tx/32dxbGfqTo7H56aZ9rNJWUy2tBEHgbnojVsBBJ6RiG17f4vNpZF6zzsWHojXA5943MxxTVobvjegURdVxeHAUqMy?cluster=devnet

After:
image


ngundotra:
Thanks for your PR! I am closing this in favor of showing the `bytes` type as `base64` since it is a little more Solana friendly

PR #339, opened 2024-05-15
Added address lookup table instruction renderers

nasjuice:
This PR handles instruction renderers for address lookup table program

https://github.com/solana-labs/explorer/issues/248


ngundotra (approved):
Looks great. Thank you for this PR!

ngundotra (changes requested):
Actually, this should be using the AddressLookupTableInstruction and various methods in @solana/web3.js instead of using `superstruct`. Apologies for earlier confusion!

ngundotra:
Closing because stale.

PR #343, opened 2024-05-29
fix: accountHeader name 

pkxro:
Uses the metadata pointer to get the token name and renders in the AccountHeader

### DIFF
--- a/app/address/[address]/layout.tsx
+++ b/app/address/[address]/layout.tsx
@@ -19,6 +19,8 @@ import { VoteAccountSection } from '@components/account/VoteAccountSection';
 import { ErrorCard } from '@components/common/ErrorCard';
 import { Identicon } from '@components/common/Identicon';
 import { LoadingCard } from '@components/common/LoadingCard';
+import { create } from 'superstruct';
+import { MetadataPointer, TokenMetadata } from '@validators/accounts/token-extension'
 import {
     Account,
     AccountsProvider,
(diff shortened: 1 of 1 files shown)


app/address/[address]/layout.tsx
```diff
@@ -245,11 +247,25 @@ function AccountHeader({ address, account, tokenInfo, isTokenInfoLoading }: { ad
...
+        if (metadataPointerExtension && metadataExtension) {
+            const tokenMetadata = create(metadataExtension.state, TokenMetadata);
+            const { metadataAddress } = create(metadataPointerExtension.state, MetadataPointer);
+
+            // Handles the basic case where MetadataPointer is reference the Token Metadata extension directly
+            // Does not handle the case where MetadataPointer is pointing at a separate account.
+            if (metadataAddress?.toString() === address) {
+                token.name = tokenMetadata.name
+            }
+    }
```
Comment (ngundotra) [inline app/address/[address]/layout.tsx:290]: This is only partially correct.

Technically you're supposed to simulate the `emit` instruction & deserialize the return data.


Here's example code. There's an edge case not covered in this too, which is when the return data is greater than 1024 bytes. But we can get to that later.

For now, I'd recommend just adding this.
```ts
async function getTokenMetadata(connection: any, programId: PublicKey, metadataPointer: PublicKey) {
    const ix = createEmitInstruction({ metadata: metadataPointer, programId });
    const message = MessageV0.compile({
        instructions: [ix],
        // Use toly.sol's key as the payer key for the simulated transaction
        payerKey: new PublicKey('86xCnPeV69n6t3DnyGvkKobf9FdN2H9oiVDdaMpo2MMY'),
        recentBlockhash: (await connection.getLatestBlockhashAndContext()).value.blockhash,
    });

    const tx = new VersionedTransaction(message);
    const result = await connection.simulateTransaction(tx, {
        commitment: 'confirmed',
        replaceRecentBlockhash: true,
        sigVerify: false,
    });
    
    if (result.value.returnData) {
        const buffer = base64.decode(result.value.returnData.data[0]);
        return deserializeTokenMetadata(buffer);
    }
    return null;
}
```

ngundotra (approved):
Left comments. What you have works, but it is a special case when t22 metadata pointer points to the mint account itself.

app/address/[address]/layout.tsx
Comment (pkxro) [reply app/address/[address]/layout.tsx:290]: Noted, will merge this in as a stopgap and merge in separate pr

PR #347, opened 2024-06-11
fix: add logo support for PYUSD & USDP

catalinred:
Based on https://github.com/solana-labs/explorer/pull/343, this PR adds support for the PYUSD & USDP tokens.

image

image


app/address/[address]/layout.tsx
```diff
@@ -275,20 +280,32 @@ function AccountHeader({
...
             const tokenMetadata = create(metadataExtension.state, TokenMetadata);
             const { metadataAddress } = create(metadataPointerExtension.state, MetadataPointer);
 
+            (async () => {
+                // Avoid re-renders once tokenData has been set
+                if (!tokenData.name && !tokenData.image) {
+                    const response = await fetch(tokenMetadata.uri);
+                    setTokenData(await response.json());
+                }
+            })();
```
Comment (jnwng) [inline app/address/[address]/layout.tsx]: instead of using the iife here i think a more canonical React hook would make sure the state works as expected:

```
import React, { useEffect, useState } from 'react';

const MyComponent: React.FC<{ tokenMetadata: { uri: string } }> = ({ tokenMetadata }) => {
    const [tokenData, setTokenData] = useState<{ name?: string, image?: string }>({});

    useEffect(() => {
        const fetchTokenData = async () => {
            if (!tokenData.name && !tokenData.image) {
                const response = await fetch(tokenMetadata.uri);
                const data = await response.json();
                setTokenData(data);
            }
        };

        fetchTokenData();
    }, [tokenMetadata.uri, tokenData.name, tokenData.image]);

    return (
        <div>
            {/* Render token data here */}
            <p>Name: {tokenData.name}</p>
            <img src={tokenData.image} alt={tokenData.name} />
        </div>
    );
};
export default MyComponent;
```

jnwng (changes requested):

app/address/[address]/layout.tsx
```diff
@@ -275,20 +280,32 @@ function AccountHeader({
...
-
+        const metadataExtension = mintInfo?.extensions?.find(
+            ({ extension }: { extension: string }) => extension === 'tokenMetadata'
+        );
+        const metadataPointerExtension = mintInfo?.extensions?.find(
+            ({ extension }: { extension: string }) => extension === 'metadataPointer'
+        );
 
         if (metadataPointerExtension && metadataExtension) {
             const tokenMetadata = create(metadataExtension.state, TokenMetadata);
```
Comment (catalinred) [inline app/address/[address]/layout.tsx]: @jnwng  I tried to implement your code suggestion but unfortunately didn't succeed. It seems that `tokenMetadata` is available only within this conditional, which is why I added the IIFE function within `AccountHeader`.

From my testing, I haven't noticed any inconsistencies with the state, but please let me know if you observe otherwise.

Comment (jnwng) [reply app/address/[address]/layout.tsx]: hmm okay, gotcha. i think this is okay for now

can you do some validation on the response / some error handling?

Comment (catalinred) [reply app/address/[address]/layout.tsx]: @jnwng I just added a basic error handling block.

app/address/[address]/layout.tsx
```diff
@@ -286,10 +291,23 @@ function AccountHeader({
...
             const { metadataAddress } = create(metadataPointerExtension.state, MetadataPointer);
 
+            (async () => {
+                try {
+                    // Avoid re-renders once tokenData has been set
+                    if (!tokenData.name && !tokenData.image) {
+                        const response = await fetch(tokenMetadata.uri);
+                        setTokenData(await response.json());
```
Comment (jnwng) [inline app/address/[address]/layout.tsx]: since we don't trust this URI, probably makes sense to unpack the response and only take the fields we want

Comment (ngundotra) [reply app/address/[address]/layout.tsx]: This layout is called in every single page load of the Explorer. If there's an error fetching the URI it will brick loading the rest of the page. 

Better way to do this is to refactor all of this in the same pattern as the `<CompressedNftAccountHeader/>` component. Once we're inside this `IIFE` block, then return a `T22AccountHeader` component.

In the `T22AccountHeader` component, instead of executing a naked fetch, try using `useMetadataJsonLink` from `@/app/providers/compressed-nft`.

ngundotra (changes requested):

PR #353, opened 2024-07-18
Fix account history

Krishnacore:
This small change will allow transaction history to be displayed on both the current version and on agave


steveluscher:
giphy-2114359118

PR #361, opened 2024-08-17
Removed undefined mixins

njtan142:
these mixins are variants, which causes errors as they are simply undefined yet


ngundotra:
Hey @njtan142 you're getting these errors because instead of pnpm you are using yarn

njtan142:
Thank you for the clarification!

PR #383, opened 2024-10-08
feat: add a simple message signing and verification page to the explorer

sagar-a16z:
There doesn't seem to be an easy way to quickly sign a message, verify, and share it. I built https://simple-message-signer.vercel.app/ earlier this year. Source: https://github.com/sagar-a16z/simple-message-signer. 

I thought it would be really cool to have this be part of the Solana Explorer. It never stores anything about users or their wallets and also supports rudimentary sharing of messages via url params. 

image


ngundotra:
Taking a look this week. Thanks for the PR!

ngundotra:
Review from @mcintyre94 offline:


* I'd strongly prefer this to be implemented without wallet-adapter, because wallet-adapter is heavy and not tree shakable and probably makes the Explorer bundle horrible. It'll also be a problem for any web3js 2.x continued integration into the Explorer. This would be much better using `@solana/react` https://github.com/solana-labs/solana-web3.js/tree/master/examples/react-app 
 
* I haven't checked if this is there, but I'd block signing a message that serializes to a transaction. Just to avoid introducing a trusted [solana.com](http://solana.com/) domain people can convince people to paste things they shouldn't in and sign them

* I'd also consider adding some UI warning about sharing the signature, because there might be people using signed transactions to verify ownership in dumb ways

sagar-a16z:
Thanks for taking a look! 

> Review from @mcintyre94 offline:
> 
> * I'd strongly prefer this to be implemented without wallet-adapter, because wallet-adapter is heavy and not tree shakable and probably makes the Explorer bundle horrible. It'll also be a problem for any web3js 2.x continued integration into the Explorer. This would be much better using `@solana/react` https://github.com/solana-labs/solana-web3.js/tree/master/examples/react-app

Might be a while before I have cycles to get to this. If this is a blocker for this I'd appreciate some help. @solana/react is a lot less dev friendly since it doesn't provide much out of the box.  

> * I haven't checked if this is there, but I'd block signing a message that serializes to a transaction. Just to avoid introducing a trusted [solana.com](http://solana.com/) domain people can convince people to paste things they shouldn't in and sign them

Good call. I used to append some bytes to the signed message to prevent that but removed that. I'll check that the message isn't a BS58 encoded Transaction or a Transaction Message. That should prevent this.

> * I'd also consider adding some UI warning about sharing the signature, because there might be people using signed transactions to verify ownership in dumb ways

Could you be more specific? I think transaction signing makes this safe but maybe I'm missing something? I'll add a little message just in case.

mcintyre94:
> Could you be more specific? I think disabling transaction signing makes this safe but maybe I'm missing something? I'll add a little message just in case.

I'm basically worried about there being some site that isn't being careful with message signing, that isn't including things like domains or timestamps in the message to sign.

Suppose there's some site that just uses "please add my address" as the message to sign. If I can get you to sign that and send me the signature, then I have everything I need to verify as you on that site. It's not obvious to users that a signature can potentially be used like that.

sagar-a16z:
> I'm basically worried about there being some site that isn't being careful with message signing, that isn't including things like domains or timestamps in the message to sign.
> 
> Suppose there's some site that just uses "please add my address" as the message to sign. If I can get you to sign that and send me the signature, then I have everything I need to verify as you on that site. It's not obvious to users that a signature can potentially be used like that.

Ah yes that's a good point. I added a message https://github.com/solana-labs/explorer/pull/383/files#diff-531a904ebfefe89fc7d323072e85a1d1e0a22aea646c9cd0916047dbd063d38bR159

image

The safest possible thing to do here is to add a pre-fix so that it can't be used anywhere else.

sagar-a16z:
Added a `_sign offchain_` prefix. I tried using the scheme proposed here https://docs.solanalabs.com/cli/examples/sign-offchain-message which is also used by the solana cli's `sign-offchain-message` command but that prefix is illegal on some wallets (Phantom for example will refuse to `signMessage` if the message containes any non UTF-8 bytes).

This prefix is not visible in the webpage but the wallet will display it.

PR #388, opened 2024-10-13
Add support for raw txn messages in search options

sahilkakwani9:
Adds functionality to the search feature to recognize and handle base64 encoded raw transaction messages. When a valid base64 encoded transaction message is entered in the search, it now resolves to the transaction inspector path.

### DIFF
--- a/app/components/SearchBar.tsx
+++ b/app/components/SearchBar.tsx
@@ -298,10 +298,31 @@ function buildOptions(rawSearch: string, cluster: Cluster, currentEpoch?: bigint
             });
         }
     } catch (err) {
-        /* empty */
+        // If bs58 decoding fails, check if it's a valid base64 string
+        if (isValidBase64(search)) {
+            options.push({
+                label: 'Transaction Message',
+                options: [
+                    {
+                        label: 'Decoded Transaction Message',
+                        pathname: `/tx/inspector?message=${search}`,
+                        value: [search],
+                    },
+                ],
+            });
+        }
     }
 
     return options;
+}
+
+function isValidBase64(str: string): boolean {
+    try {
+        Buffer.from(str, 'base64');
+        return true;
+    } catch (err) {
+        return false;
+    }
 }
 
 function DropdownIndicator() {


ngundotra:
Hey this is an awesome PR! Thanks for sending. Will try out later today!

app/components/SearchBar.tsx
```diff
@@ -298,12 +298,33 @@ function buildOptions(rawSearch: string, cluster: Cluster, currentEpoch?: bigint
             });
         }
     } catch (err) {
-        /* empty */
+        // If bs58 decoding fails, check if it's a valid base64 string
+        if (isValidBase64(search)) {
+            options.push({
+                label: 'Transaction Message',
```
nickfrosty: Can you make the label be more clearly related to the Inspector since that is where the user will be taken to? 
```suggestion
                label: 'Transaction Inspector',
```

app/components/SearchBar.tsx
```diff
@@ -298,12 +298,33 @@ function buildOptions(rawSearch: string, cluster: Cluster, currentEpoch?: bigint
...
+                options: [
+                    {
+                        label: 'Decoded Transaction Message',
```
nickfrosty: Similar comment as the other:
```suggestion
                        label: 'Inspect Decoded Transaction',
```

app/components/SearchBar.tsx
```diff
@@ -298,12 +298,33 @@ function buildOptions(rawSearch: string, cluster: Cluster, currentEpoch?: bigint
...
+                        pathname: `/tx/inspector?message=${search}`,
```
nickfrosty: Since the inspector can parse a base64 encoded full transaction (message and signatures), the search box should handle that as well.

Can you fix this to support both?

Here is an example full transaction (single memo instruction) that is created by serializing an entire transaction:

```
AQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAID5P9GJXpSnreZM8/18ftjMR6Y3g6/OQm+Ax7miUsjUhEDBkZv5SEXMv/srbpyw5vnvIzlu8X3EmssQ5s6QAAAAAVKU1qZKSEGTSTocWDaOHx8NbXdvJK7geQfqEBBBUSNBPY+6MoJeAOVWbvgg6ALv7vjwNbnziJMJUlhpmJZREYCAQAJA+gDAAAAAAAAAgAedGhpcyBpcyBhIHNpbXBsZSBtZW1vIG1lc3NhZ2Uy
```

Reference: Web3.js serialize [Legacy Transaction](https://solana-labs.github.io/solana-web3.js/classes/Transaction.html#serialize) and [Version Transaction](https://solana-labs.github.io/solana-web3.js/classes/VersionedTransaction.html#serialize)

nickfrosty (changes requested):
Thanks for adding this @sahilkakwani9

I just tested this out locally and it does not seem to fully work out of the box.

Also, this being merged should resolve the following issue:

Fixes: https://github.com/solana-labs/explorer/issues/385

app/components/SearchBar.tsx
sahilkakwani9: thanks for the review. I have added the support now and it should work accordingly for full serialized transactions as well as serialized messages.

nickfrosty:
@sahilkakwani9 is this now ready for review again?

sahilkakwani9:
> @sahilkakwani9 is this now ready for review again?

Yes ser

nickfrosty (approved):
Sorry for the delay on this, I totally dropped the ball. This looks great, thanks again :)

PR #396, opened 2024-10-24
Add verify command to verified build tab

Woody4618:
- Deriving the PDA from the upgrade authority and the program id, then getting the parameter from the PDA and composing it into a solana-verify comand

Like this people can easily copy the command from the explorer and verify for themselfs. 

Related Issue: 
https://github.com/solana-labs/explorer/issues/393 

Looks like this: 
image

And like this when there is no Verify PDA:
image

### DIFF
--- a/app/utils/verified-builds.tsx
+++ b/app/utils/verified-builds.tsx
@@ -1,10 +1,13 @@
 import { sha256 } from '@noble/hashes/sha256';
-import { PublicKey } from '@solana/web3.js';
+import { Connection, PublicKey } from '@solana/web3.js';
 import useSWRImmutable from 'swr/immutable';
 
+import { useAnchorProgram } from '../providers/anchor';
+import { useCluster } from '../providers/cluster';
 import { ProgramDataAccountInfo } from '../validators/accounts/upgradeable-program';
 
 const OSEC_REGISTRY_URL = 'https://verify.osec.io';
+const VERIFY_PROGRAM_ID = 'verifycLy8mB96wd9wqq3WDXQwM4oU6r42Th37Db9fC';
 
 export type OsecRegistryInfo = {
     is_verified: boolean;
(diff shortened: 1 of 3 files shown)


app/utils/verified-builds.tsx
```diff
@@ -13,29 +16,90 @@ export type OsecRegistryInfo = {
...
+
+    if (registryError || pdaError) {
+        return { data: null, error: registryError || pdaError, isLoading };
+    }
+
+    // Create command from the args of the verify PDA
+    if (registryData && pdaData && !isLoading) {
+        const verifiedData = registryData as OsecRegistryInfo;
+        verifiedData.verify_command = `solana-verify verify-from-repo -um --program-id ${programId.toBase58()} ${
+            pdaData.gitUrl
```
ngundotra, line 83: Should we add a marker somewhere telling people it's only possible to verify mainnet programs right now?

app/utils/verified-builds.tsx
```diff
@@ -13,29 +16,90 @@ export type OsecRegistryInfo = {
...
+    programAuthority,
     options,
 }: {
     programId: PublicKey;
+    programAuthority: PublicKey | null;
     options?: { suspense: boolean };
 }) {
-    const { data, error, isLoading } = useSWRImmutable(
+    const { url: clusterUrl } = useCluster();
+    const connection = new Connection(clusterUrl);
```
ngundotra, line 33: As of right now this will fetch PDA args uploaded to devnet, but the command will only work for mainnet. 

I think the command should only be shown for mainnet, wdyt?

ngundotra (changes requested):

app/utils/verified-builds.tsx
Woody4618: Sure can do. Any idea on how we should show it? Is there some component for hints or so?

app/utils/verified-builds.tsx
Woody4618: Ah just showing it for mainnet is probably the easiest solution then we dont need an addtional hint right?

Woody4618: Added a mainnet check

app/utils/verified-builds.tsx
```diff
@@ -13,29 +17,97 @@ export type OsecRegistryInfo = {
...
+    if (registryError || pdaError) {
+        return { data: null, error: registryError || pdaError, isLoading };
+    }
+
+    // Create command from the args of the verify PDA
+    if (registryData && pdaData && !isLoading) {
+        const verifiedData = registryData as OsecRegistryInfo;
+        verifiedData.verify_command = `solana-verify verify-from-repo -um --program-id ${programId.toBase58()} ${
+            pdaData.gitUrl
+        } --commit-hash ${pdaData.commit}`;
```
ngundotra, line 84: isn't it possible that we are omitting other general args like `-b` ? Do you think it's worth adding now that we understand it's an ok security surface?

Woody4618: Yep probably makes sense. Sounds dangerous

Woody4618: Yes also makes sense. Ill add a sentence and a link to the docs 👍

Woody4618: Updated. Looks nice i think: 
image
Also filtering out the -b and --base-image flag

PR #398, opened 2024-10-29
Address icon consistency

catalinred:
This PR adds new PNG and SVG favicons, similar to [solana-com](https://github.com/solana-foundation/solana-com/tree/main/public) for a better consistency.  The [PNG](https://caniuse.com/link-icon-png) fallback is used for legacy browsers while the [SVG ](https://caniuse.com/link-icon-svg) is added for supporting higher resolutions.

---

The explorer app was previously using an obsolete Solana icon (right), visible in the address bar below when compared to existing solana.com favicon (left). 

image


ngundotra:
Thanks Catalin!

PR #406, opened 2024-11-11
Add source code url and fix --commit-hash bug

Woody4618:
Showing the repository URL in that case the on chain PDA was uploaded. 

Fix: 
The -- commit hash was always shown even when there was no commit hash used in the verify args

### DIFF
--- a/app/components/account/VerifiedBuildCard.tsx
+++ b/app/components/account/VerifiedBuildCard.tsx
@@ -29,10 +29,10 @@ export function VerifiedBuildCard({ data, pubkey }: { data: UpgradeableLoaderAcc
 
     // Define the message based on the verification status
     let verificationMessage = 'Information provided by osec.io';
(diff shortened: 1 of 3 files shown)


Woody4618:
These are the different states: 
Osec Verified 
image

Self verified: 
image

Not verified: 
image

Woody4618:
These are the three different states: 
http://localhost:3002/address/MkabCfyUD6rBTaYHpgKBBpBo5qzWA2pK2hrGGKMurJt/verified-build
http://localhost:3002/address/kGYz2q2WUYCXhKpgUF4AMR3seDA9eg8sbirP5dhbyhy/verified-build
http://localhost:3002/address/oreV2ZymfyeXgNgBdqMkumTqqAprVqgBWQfoYkrtKWQ/verified-build

ngundotra (approved):

PR #419, opened 2024-11-22
Add commit hash to URL

Woody4618:
Problem: 
The URL from the verified on chain data did not have the commit hash attached. 

Solution: 
Added the commit with the /tree/ formate same way the osec API does it. 
Also changed it so that the Osec api is prefered instead of overwritten from the PDA when available.

When commit hash available: 
image
 
When there is no commit hash: 
image


ngundotra:
LGTM, will fix and merge soon

ngundotra:
This was included in a different PR. Closing

PR #421, opened 2024-11-26
added /tx/inspector route to return transaction data

YaTut1901:


ngundotra:
Hey Ivan, thanks for putting this together.

Some questions

1) It looks like the InspectorPage never actually uses the api route. Can you update this to use the api route you added?

2) Can you provide a few tests that assert api responses for different protocols on both mainnet & devnet?

YaTut1901:
@ngundotra you mean make InspectorPage to use api that I created to fetch data and display?

YaTut1901:
@ngundotra could you please clarify?

ngundotra:
Yes can you please make the InspectorPage use the API route & add some tests?

YaTut1901:
@ngundotra it is quite a big task to refactor all that code in InspectorPage to use fetching from API. I would say this project is better to be migrated to monorepo structure to have API version and UI. But if to talk about refactoring just InspectorPage, may we finish this task with just making test for API and open another to refactor the UI?

scutuatua-crypto (approved):

PR #426, opened 2024-12-19
Only show Trusted Verified Build information

ngundotra:
Verified Build information was susceptible to a "griefing" attack disclosed by security firm Accretion. The vulnerability stemmed from the fact that anyone could overwrite verified build information in the OSec API by providing a more recent verified build information. This allowed malicious actors to overwrite protocol's GitHub repo url in the OSec API.

This has now been patched by OtterSec. The fix requires that program developers write their verified build arguments on-chain using their program authority. 

For previously verified programs, their verification information was uploaded by the OSec team. However, to update previously verified information, teams must deploy their verified build information onchain.

For teams using multisig program authorities, we've described a process to prepare this transaction using `solana-verify` `v0.4.0` here: https://solana.com/developers/guides/advanced/verified-builds#how-to-verify-your-program-when-its-controlled-by-a-multisig-like-squads

### DIFF
--- a/app/utils/verified-builds.tsx
+++ b/app/utils/verified-builds.tsx
@@ -211,22 +211,8 @@ function coalesceCommandFromPda(verifiedData: OsecRegistryInfo, programId: Publi
 
     // Add additional args if available, for example mount-path and --library-name
     if (pdaData.args && pdaData.args.length > 0) {
-        const filteredArgs = [];
-
-        for (let i = 0; i < pdaData.args.length; i++) {
-            const arg = pdaData.args[i];
-
-            if (arg === '-b' || arg === '--base-image') {
-                i++; // Also skip the parameter
-                continue;
-            }
-            filteredArgs.push(arg);
-        }
-
-        if (filteredArgs.length > 0) {
-            const argsString = filteredArgs.join(' ');
-            verifiedData.verify_command += ` ${argsString}`;
-        }
+        const argsString = pdaData.args.join(' ');
+        verifiedData.verify_command += ` ${argsString}`;
     }
     verifiedData.repo_url = pdaData.gitUrl;
     if (verifiedData.verification_status === VerificationStatus.NotVerified) {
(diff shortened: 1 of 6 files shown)


app/utils/verified-builds.tsx
```diff
@@ -39,111 +55,171 @@ export function useVerifiedProgramRegistry({
...
+    });
+
+    const mappedBySigner: Record<string, OsecInfo> = {};
+
+    // Map the registryData by signer in order to enforce hierarchy of trust
+    registryData?.forEach(entry => {
+        mappedBySigner[entry.signer] = entry;
+    });
+
+    // Get the program authority's entry first, then the trusted signers
```
Woody4618, line 116: Would it make sense to get these already sorted from the API? Would make this easier to implement for other explorers

app/utils/verified-builds.tsx
```diff
@@ -39,111 +55,171 @@ export function useVerifiedProgramRegistry({
...
-        verifiedData.verify_command = isMainnet(cluster)
+    const message = TRUSTED_SIGNERS[osecInfo?.signer || '']
+        ? 'Verification information provided by a trusted signer.'
+        : 'Verification information provided by the program authority.';
+
+    const enrichedOsecInfo: OsecRegistryInfo = {
+        ...osecInfo,
+        message,
+        signer: osecInfo?.signer || '',
+        verification_status: VerificationStatus.Verified,
```
Woody4618: Is this always VerificationStatus.Verified or could it also not be verified?  I think this would need to come from the osecInfo?

Woody4618 (approved):

app/utils/verified-builds.tsx
ngundotra: I think it's fair to ask explorers to sort the results because each one is responsible for deciding what "verified" means to them.

app/utils/verified-builds.tsx
ngundotra: Great catch! Patched!

PR #442, opened 2025-01-23
Automatic Feature Gate Ingest

ngundotra:
Refactor feature gate data into standalone, generated file called `app/utils/feature-gate/featureGates.json`

Github CI actions to update nightly


scripts/parse_feature_gates.py
```diff
@@ -0,0 +1,90 @@
...
+            new_feature = features_by_key[existing['key']]
+            existing_features[i]['devnetActivationEpoch'] = new_feature['devnetActivationEpoch']
+            existing_features[i]['testnetActivationEpoch'] = new_feature['testnetActivationEpoch']
+            del features_by_key[existing['key']]
+    
+    new_features = list(features_by_key.values())
+    if len(new_features) > 0:
+        print("New features:")
+        for f in new_features:
+            print(f"{f['key']} - {f['title']}")
```
gitteri, line 81: Was this for debugging or is this so we have these logs in the github action?

ngundotra: So we can have logs in the github action

PR #459, opened 2025-02-10
Sigverify Ed25519 Payload Decoding Standard

ngundotra:
TBD - sRFC number (waiting on Drift)

This PR adds support for decoding Ed25519 payloads that reference another program's instructions.

Example transaction on devnet - [for local testing](http://localhost:3000/tx/32jTSg2bLQLytXTrA7rgRTEZGE5uU2mMEMZFrnoZK6t4DsUtjVr1nZYUmypMwSrWcgPXG2rKH7MbCGDBPPbbWwP8?cluster=devnet)

Future to-do:
- When the Ed25519 precompile data is not externally referenced, meaning it is self-contained in the ed25519 ix data, then attempt to decode using the following outer-instruction's anchor idl.

### DIFF
--- a/app/components/instruction/ed25519/Ed25519DetailsCard.tsx
+++ b/app/components/instruction/ed25519/Ed25519DetailsCard.tsx
@@ -1,5 +1,6 @@
-import { BorshTypesCoder } from '@coral-xyz/anchor/dist/cjs/coder/borsh/types';
-import { hash } from '@coral-xyz/anchor/dist/cjs/utils/sha256';
+import { Program } from '@coral-xyz/anchor';
+import { IdlType } from '@coral-xyz/anchor/dist/cjs/idl';
+import { sha256 } from '@noble/hashes/sha256';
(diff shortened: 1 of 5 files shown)


app/components/instruction/ed25519/Ed25519DetailsCard.tsx
```diff
@@ -77,6 +84,178 @@ const extractData = (
...
+                    </td>
+                </tr>
+            );
+        }
+
+        const name = decodedMessage.name;
+        const data = decodedMessage.data;
+
+        if (!name || !data) {
+            return null;
```
LukasDeco, line 166: Should you instead just fallback to rendering the message as a base64 string like in the clause above?

ngundotra: yea you were right

scutuatua-crypto (approved):

PR #461, opened 2025-02-13
feat: show feature info

rogaldh:
Here is how new functionality looks like:
image
image

Variant for the feature without `simd_link`:
image


app/components/common/TableCardBody.tsx
```diff
@@ -1,9 +1,24 @@
...
+        layout: {
+            "compact": ["table-nowrap"],
+            "expanded": []
+        }
+    }
+})
+
+export interface TableCardBodyProps extends VariantProps<typeof tableVariants>, React.PropsWithChildren {}
+
+export function TableCardBody({ children, ...props }: TableCardBodyProps) {
```
ngundotra: What is this doing?

app/components/account/FeatureAccountSection.tsx
```diff
@@ -20,42 +27,157 @@ type Props = Readonly<{
...
+                <td className="text-lg-end">
+                    {featureInfo.simd && featureInfo.simd_link ? (
+                        <a
+                            href={featureInfo.simd_link}
+                            target="_blank"
+                            rel="noopener noreferrer"
+                            className=""
+                        >
+                            See SIMD {featureInfo.simd}
+                        </a>
```
ngundotra: nit: remove "See SIMD" and just have "SIMD {featureInfo.simd}" and the little logo that tells people it will open a new tab/window

app/address/[address]/feature-gate/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const feature = getFeatureInfo(address);
+    const data = await fetchFeatureGateInformation(feature);
+
+    // remark-gfm won't handle github-flavoured-markdown with a table present at it
+    // TODO: figure out a configuration to render GFM table correctly
+    return (
+        <FeatureGateCard>
+            <ReactMarkdown remarkPlugins={[remarkGFM, remarkFrontmatter]}>
+                {data}
+            </ReactMarkdown>
```
Comment (ngundotra) [inline app/address/[address]/feature-gate/page.tsx]: is there a way to get the text color to match the existing default text color (white)? It's a little bit hard to read with the current gray color

ngundotra (changes requested):
Looks great! Thank you for putting this together. Few nits

app/components/common/TableCardBody.tsx
rogaldh: That code adds "variants" for the table. Original one contains `table-nowrap` class that adds `no-wrap` to all the cells.
For the feature we do not need this, as description might be quite long and with `no-wrap` there will be a scroll. It's not very convenient.

That's why I've added additional variant (`expanded`), which allows to not use `no-wrap` and use it for all the needed cells.

All the previous tables won't be affected

app/components/account/FeatureAccountSection.tsx
rogaldh: Updated

rogaldh: image

app/address/[address]/feature-gate/page.tsx
Comment (rogaldh) [reply app/address/[address]/feature-gate/page.tsx]: There is, but it looks a bit overwhelming. Check it out:
image

app/components/account/FeatureAccountSection.tsx
```diff
@@ -20,14 +29,43 @@ type Props = Readonly<{
 }>;
 
 const FeatureCard = ({ account }: Props) => {
+    const { cluster } = useCluster();
     const feature = parseFeatureAccount(account);
```
ngundotra, line 46: Can you make this FeatureCard show even when the account does not exist on-chain?

I think it is vastly more informative current feature card.

ngundotra (changes requested):
Everything looks great. You were right about text-color: white -- too strong. I reset it to muted like you had. 

I also made a modification so that the feature gate tab always shows for the feature gate. Before it would only show if the key was initialized on the current cluster. Can you just help update the feature account card so it always renders, even when the account doesn't exist?

app/components/account/FeatureAccountSection.tsx
rogaldh: Implemented.

There was exception thrown upon invalid data. I've adjusted logic to allow showing the Feature card for accounts that present at featureGates.json (even them do not owned by Feature program)

ngundotra (approved):
Looks fantastic. Thank you for the changes

PR #462, opened 2025-02-13
chore: improve semicolon rule

rogaldh:
- [x] Add semicolon rule for `eslint`
    - Current configuration allows to be inconsistent with semicolon
- [x] Add script to check for format errors (it uses current configuration for `prettier`)

Next thing that might be good to implement: check for prettier' errors at lint step. This allows to prevent passing CI checks for codebase formatted incorrectly.


ngundotra:
need to rebase

ngundotra (approved):

PR #469, opened 2025-02-20
fix: catch errors at Solflare's api to render layout properly

rogaldh:
Address that is default for solflare-wallet/utl-sdk's api requests returns 404 at this moment for mints.

[Link](https://token-list-api.solana.cloud/v1/mints?chainId=101)


rogaldh:
https://github.com/solflare-wallet/utl-api/issues/5

PR #603, opened 2025-06-05
Scaled ui amount support for balances, supply, transfers

gitteri:
# Scaled ui amount support for balances, supply, transfers

## Description

This PR adds support for the Token-2022 Scaled UI Amount extension, which allows tokens to display different UI amounts than their raw on-chain amounts. This is particularly useful for tokens that want to adjust their display values without requiring a migration of existing balances (stock splits, dividend payments, yield etc).

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [X] Protocol integration
-   [ ] Documentation update

## Screenshots

Screenshot 2025-06-04 at 7 21 39 PM
Screenshot 2025-06-04 at 7 22 22 PM
Screenshot 2025-06-04 at 7 22 55 PM

## Changes
- Added `getCurrentTokenScaledUiAmountMultiplier` to calculate the current multiplier based on the token's extension data and timestamp
- Added `calculateCurrentTokenScaledUiAmountMultiplier` to derive the multiplier from raw and UI amounts
- Added comprehensive test coverage for both functions
- Added `ScaledUiAmountMultiplierTooltip` to show an info icon with details about the multiplier
- Updated UI in all places that have summary amounts (balances, transfer amounts, supply) to add a tooltip if the scaled ui multiplier is in effect

## Technical Note
Currently the "getTransaction" RPC endpoint does not return the scaled amount of the transfer in the pre and post balances section. This PR works around this by manually calculating this amount. In the future, Anza is likely to update the RPC to return the scaled amount but this works for now.

## Testing
Added test suites covering:
- Token extensions with and without scaling configuration
- Timestamp-based multiplier selection
- Various decimal places and amount scenarios
- Edge cases like fractional multipliers and large numbers

## Checklist

-   [x] CI/CD checks pass

## Additional Notes

----

> [!IMPORTANT]
> Adds support for Token-2022 Scaled UI Amounts, updating UI components and adding functions to calculate and display scaled amounts.
> 
>   - **Behavior**:
>     - Adds support for Token-2022 Scaled UI Amount extension to display different UI amounts than raw on-chain amounts.
>     - Updates UI components (`OwnedTokensCard.tsx`, `TokenAccountSection.tsx`, `TokenTransfersCard.tsx`) to show tooltips with scaling details.
>     - Manually calculates scaled amounts due to current RPC limitations.
>   - **Functions**:
>     - Adds `getCurrentTokenScaledUiAmountMultiplier()` and `calculateCurrentTokenScaledUiAmountMultiplier()` in `token-info.ts`.
>     - Adds `uiAmountToRawAmount()` in `index.ts`.
>   - **Components**:
>     - Adds `ScaledUiAmountMultiplierTooltip` component in `ScaledUiAmountMultiplierTooltip.tsx`.
>   - **Tests**:
>     - Adds tests for `calculateCurrentTokenScaledUiAmountMultiplier` and `getCurrentTokenScaledUiAmountMultiplier` in `calculate-scaled-ui-multiplier-test.ts` and `get-scaled-ui-multiplier-test.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fexplorer&utm_source=github&utm_medium=referral)<sup> for 90d259bb31cb4a6e7bcc26ce6b55a641a8be327e. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


app/components/account/TokenAccountSection.tsx
```diff
@@ -157,9 +160,12 @@ function FungibleTokenMintAccountCard({
                 <tr>
                     <td>{mintInfo.mintAuthority === null ? 'Fixed Supply' : 'Current Supply'}</td>
                     <td className="text-lg-end">
-                        {normalizeTokenAmount(mintInfo.supply, mintInfo.decimals).toLocaleString('en-US', {
-                            maximumFractionDigits: 20,
-                        })}
+                        <span>
```
rogaldh, line 163: image

I think that icon at the main card is not very well aligned. The simplest solution would be to adjust it's position via `transalte(-2px, -1px)`

rogaldh:
image

I think that data on that page is misleading. I see current supply (20000) and down there transfers are displayed, but it amount is not altered.
It becomes harder to do the math as I have to divide 20000 by the multiplicator (what it multiplicator will have fractions?).
Maybe it will be better to display oridinal amout and altered value inside the tooltip.

app/utils/token-info.ts
```diff
@@ -175,3 +176,28 @@ export async function getTokenInfos(
...
+export function calculateCurrentTokenScaledUiAmountMultiplier({
+    amount,
+    decimals,
+    uiAmount,
+}: {
+    amount: string;
+    decimals: number;
+    uiAmount: number;
+}): number {
+    const rawAmount = Number(amount);
```
rogaldh: Might need to check for NaN as it will lead to division errors

rogaldh:
image
Amount is displayed with modificator included and to get original amount you have to count in head. I think it might be helpful to display original value to not count in head every time (imagine you have fractions for the modifier)

rogaldh:
I've also found issue with calculation. You can see NaN inside the tooltip
image

http://localhost:3000/address/DmatrEXUzjvGRQeQAUwefktx78f5a8ZU2gTUvDv6Zmca/tokens?cluster=devnet

ngundotra (approved):
Clean change! Ty for this

PR #661, opened 2025-07-28
Adding Transaction Cost (costUnits) to the Explorer

waldrobi:
## Description

Added `Transaction Cost` to block overview and transaction overview
Utilization is computed using `Transaction Cost` rather than `Compute Units Consumed`

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [x] Protocol integration
-   [ ] Documentation update

## Screenshots

Block Overview:
image

Signature Overview:
image

## Testing

There are no existing tests for these classes.
I did however double check again the RPC APIs with the screenshots above. 

## Related Issues

PR adding `costUnits` in `solana/web3.js` 
https://github.com/solana-foundation/solana-web3.js/pull/3748

## Checklist

-   [ ] My code follows the project's style guidelines
-   [ ] I have added tests that prove my fix/feature works
-   [ ] CI/CD checks pass

## Additional Notes

----

> [!IMPORTANT]
> Adds transaction cost (costUnits) to block and transaction overviews, updating utilization calculations and UI components.
> 
>   - **Behavior**:
>     - Adds `Transaction Cost` to block and transaction overviews in `layout.tsx` and `page-client.tsx`.
>     - Utilization now computed using `Transaction Cost` instead of `Compute Units Consumed` in `layout.tsx`.
>   - **UI Changes**:
>     - Adds `Txn Cost` column to transaction tables in `BlockHistoryCard.tsx`.
>     - Allows sorting by `txnCost` in `BlockHistoryCard.tsx`.
>   - **Data Handling**:
>     - Introduces `costUnits` in transaction data structures in `BlockHistoryCard.tsx` and `page-client.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fexplorer&utm_source=github&utm_medium=referral)<sup> for fb05cc773b14228edc6658faaae1c980e9d59f92. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app/block/[slot]/layout.tsx
+++ b/app/block/[slot]/layout.tsx
@@ -48,11 +48,14 @@ function BlockLayoutInner({ children, params: { slot } }: Props) {
 
         let totalCUs = 0;
         let totalRequestedCUs = 0;
+        let totalCostUnits = 0;
         for (const tx of block.transactions) {
             const requestedCUs = estimateRequestedComputeUnits(tx, epoch, cluster);
             const cus = tx.meta?.computeUnitsConsumed ?? 0;
+            const costUnits = tx.meta?.costUnits ?? 0;
             totalRequestedCUs += requestedCUs;
             totalCUs += cus;
+            totalCostUnits += costUnits;
         }
 
         const showSuccessfulCount = block.transactions.every(tx => tx.meta !== null);
@@ -170,11 +173,17 @@ function BlockLayoutInner({ children, params: { slot } }: Props) {
                             </tr>
                         )}
                         <tr>
-                            <td className="w-100">Compute Unit Utilization</td>
+                            <td className="w-100">Total Compute Units Consumed</td>
+                            <td className="text-lg-end font-monospace">
+                                <span>{totalCUs.toLocaleString()}</span>
+                            </td>
+                        </tr>
+                        <tr>
+                            <td className="w-100">Transaction Cost Utilization</td>
                             <td className="text-lg-end font-monospace">
                                 <span>
-                                    {totalCUs.toLocaleString()} / {maxComputeUnits.toLocaleString()} (
-                                    {Math.round((totalCUs / maxComputeUnits) * 100)}%)
+                                    {totalCostUnits.toLocaleString()} / {maxComputeUnits.toLocaleString()} (
+                                    {Math.round((totalCostUnits / maxComputeUnits) * 100)}%)
                                 </span>
                             </td>
                         </tr>
--- a/app/components/block/BlockHistoryCard.tsx
+++ b/app/components/block/BlockHistoryCard.tsx
@@ -172,7 +172,7 @@ export function BlockHistoryCard({ block, epoch }: { block: VersionedBlockRespon
         if (sortMode === 'compute' && showComputeUnits) {
             filteredTxs.sort((a, b) => b.computeUnits! - a.computeUnits!);
         } else if (sortMode === 'txnCost') {
-            filteredTxs.sort((a, b) => (b.costUnits! - a.costUnits!));
+            filteredTxs.sort((a, b) => b.costUnits! - a.costUnits!);
         } else if (sortMode === 'fee') {
             filteredTxs.sort((a, b) => (b.meta?.fee || 0) - (a.meta?.fee || 0));
         } else if (sortMode === 'reservedCUs') {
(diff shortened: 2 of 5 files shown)


app/components/block/BlockHistoryCard.tsx
```diff
@@ -318,6 +340,11 @@ export function BlockHistoryCard({ block, epoch }: { block: VersionedBlockRespon
                                                     : 'Unknown'}
                                             </td>
                                         )}
+                                        <td>
+                                            {tx.costUnits !== undefined
+                                                        ? new Intl.NumberFormat('en-US').format(tx.costUnits)
+                                                        : 'Unknown'}
```
jacobcreech: It's interesting that we put `unknown` on the block page when the costUnits error, but just don't show on the individual signature page when the costUnits error. I don't think this is an issue for this PR, but noting in https://github.com/solana-foundation/explorer/issues/662

app/block/[slot]/layout.tsx
```diff
@@ -170,11 +173,17 @@ function BlockLayoutInner({ children, params: { slot } }: Props) {
...
+                            </td>
+                        </tr>
+                        <tr>
+                            <td className="w-100">Transaction Cost Utilization</td>
                             <td className="text-lg-end font-monospace">
                                 <span>
-                                    {totalCUs.toLocaleString()} / {maxComputeUnits.toLocaleString()} (
-                                    {Math.round((totalCUs / maxComputeUnits) * 100)}%)
+                                    {totalCostUnits.toLocaleString()} / {maxComputeUnits.toLocaleString()} (
+                                    {Math.round((totalCostUnits / maxComputeUnits) * 100)}%)
```
Comment (ngundotra) [inline app/block/[slot]/layout.tsx:186]: This should just be Transaction Cost, don't need utilization, since ComputeBudget limit does not apply to transaction cost (only applies to execution CUs)

app/components/block/BlockHistoryCard.tsx
```diff
@@ -40,10 +40,11 @@ const useQueryAccountFilter = (query: ReadonlyURLSearchParams): PublicKey | null
     return null;
 };
 
-type SortMode = 'index' | 'compute' | 'fee' | 'reservedCUs';
+type SortMode = 'index' | 'compute' | 'txn cost' | 'fee' | 'reservedCUs';
```
ngundotra: this key should be one word, since it goes into the `UrlSearchParam` as `sortBy=txn%20cost`. simpler if key was just `txnCost`

app/components/block/BlockHistoryCard.tsx
ngundotra: Maybe `Not Available` is better?

jacobcreech: That's also confusing. I'd rather just not show the component to avoid people wondering what it is.

app/block/[slot]/layout.tsx
Comment (waldrobi) [reply app/block/[slot]/layout.tsx:186]: Are you sure of that?  In `agave` the limit applies to Transaction cost. 

The check against the limit is [here](https://github.com/anza-xyz/agave/blob/7e7ad5b46740db8a8cc5f45d52dfb83a73c7d1f4/cost-model/src/cost_tracker.rs#L291-L295)

Which uses this [sum function ](https://github.com/anza-xyz/agave/blob/7e7ad5b46740db8a8cc5f45d52dfb83a73c7d1f4/cost-model/src/cost_tracker.rs#L282)

The sum is [defined here](https://github.com/anza-xyz/agave/blob/7e7ad5b46740db8a8cc5f45d52dfb83a73c7d1f4/cost-model/src/transaction_cost.rs#L167-L175) which equates to `Transaction Cost`

app/components/block/BlockHistoryCard.tsx
waldrobi: Agreed, that I found the inconsistency odd. 
Moving the discussion to the hyperlinked issue.

app/block/[slot]/layout.tsx
Comment (waldrobi) [reply app/block/[slot]/layout.tsx:186]: This was part of the reason I chose to move utilisation to TC from CU

Comment (ngundotra) [reply app/block/[slot]/layout.tsx:186]: @waldrobi Yup you're right. This makes sense. I mistakenly thought this was the transaction page.

Good to go on this

PR #759, opened 2025-11-21
feat: Improvements for InteractiveIDL UI

askov:
## Description

Rework the existing IDL tab components for consistency according to the design.

### ⚠️ Stories review 
Reviewing the stories is optional. This code does not affect production and is intended for developers only.

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Protocol integration
-   [ ] Documentation update

## Screenshots
| Before | After |
| -------- | ------- |
|explorer solana com_address_ProgM6JCCvbYkfKqJYHePx4xxSUSqJp7rh8Lyv7nk7S_idl| localhost_3000_address_ProgM6JCCvbYkfKqJYHePx4xxSUSqJp7rh8Lyv7nk7S_idl   |

## Testing
- Visit an IDL page, e.g. http://localhost:3000/address/ProgM6JCCvbYkfKqJYHePx4xxSUSqJp7rh8Lyv7nk7S/idl

## Related Issues
N/A

## Checklist

-   [x] CI/CD checks pass
-   [ ] I have included screenshots for protocol screens (if applicable)

----

> [!IMPORTANT]
> Enhance InteractiveIDL UI by refactoring IDL components, adding structured rendering, and improving styling and testing.
> 
>   - **UI Enhancements**:
>     - Refactor `IdlCard` in `IdlCard.tsx` to use tabs for `program-metadata` and `anchor` IDLs.
>     - Add `IdlSection` in `IdlSection.tsx` for rendering IDL with search and download features.
>     - Introduce `BaseFormattedIdl` and related components for structured IDL rendering.
>   - **Components**:
>     - Add `BaseIdlAccounts`, `BaseIdlConstants`, `BaseIdlErrors`, `BaseIdlEvents`, `BaseIdlFields`, `BaseIdlInstructions`, `BaseIdlPdas`, `BaseIdlTypes` for detailed IDL sections.
>     - Implement `AnchorFormattedIdl` and `CodamaFormattedIdl` for specific IDL formats.
>   - **Utilities**:
>     - Add `triggerDownload` in `triggerDownload.ts` for downloading IDL files.
>     - Update `tailwind.config.ts` and `styles.css` for styling consistency.
>   - **Testing**:
>     - Add tests for `IdlCard` in `IdlCard.spec.tsx` to verify rendering logic.
>     - Add Storybook stories for new components to demonstrate usage and appearance.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fexplorer&utm_source=github&utm_medium=referral)<sup> for f1550481621d2c5e4e787ef5070922d50a062102. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app/shared/lib/triggerDownload.ts
+++ b/app/shared/lib/triggerDownload.ts
@@ -1,9 +1,98 @@
-export const triggerDownload = async (data: string, filename: string, type?: string): Promise<void> => {
-    const blob = new Blob([Buffer.from(data, 'base64')], type ? { type } : {});
-    const fileDownloadUrl = URL.createObjectURL(blob);
-    const tempLink = document.createElement('a');
-    tempLink.href = fileDownloadUrl;
-    tempLink.setAttribute('download', filename);
-    tempLink.click();
-    URL.revokeObjectURL(fileDownloadUrl);
+export interface DownloadOptions {
+    /** MIME type for the file */
+    type?: string;
+    /** Maximum file size in bytes. Prevents DoS by limiting memory usage. */
+    maxSize?: number;
+}
+
+const DEFAULT_MAX_SIZE = 10 * 1024 * 1024; // 10MB in bytes
+
+/**
+ * Triggers a file download from base64 data
+ *
+ * @param data - Base64 encoded string
+ * @param filename - Name for the downloaded file (browsers automatically sanitize this)
+ * @param options - Optional configuration object with type and maxSize
+ * @throws Error if data is invalid base64, file size exceeds maxSize, or if download fails
+ */
+export const triggerDownload = async (data: string, filename: string, options?: DownloadOptions): Promise<void> => {
+    if (!data || typeof data !== 'string') {
+        throw new Error('Invalid data: must be a non-empty string');
+    }
+
+    if (!isValidBase64(data)) {
+        throw new Error('Invalid data: not a valid base64 string');
+    }
+
+    const type = options?.type;
+    const maxSize = options?.maxSize ?? DEFAULT_MAX_SIZE;
+
+    // Check size BEFORE decoding to prevent DoS via memory exhaustion
+    // Base64 encoding increases size by ~33%, so we estimate: base64Length * 3/4
+    // This prevents creating expensive Buffer/Blob objects for huge files
+    const estimatedDecodedSize = Math.ceil((data.length * 3) / 4);
+    if (estimatedDecodedSize > maxSize) {
+        throw new Error(
+            `File size (estimated ${formatBytes(estimatedDecodedSize)}) exceeds maximum allowed size (${formatBytes(
+                maxSize
+            )})`
+        );
+    }
+
+    let fileDownloadUrl: string | null = null;
+    let tempLink: HTMLAnchorElement | null = null;
+
+    try {
+        const decodedData = Buffer.from(data, 'base64');
+
+        // MIME type is not validated because:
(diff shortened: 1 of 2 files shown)


app/shared/lib/triggerDownload.ts
```diff
@@ -0,0 +1,9 @@
+export const triggerDownload = async (data: string, filename: string, type?: string): Promise<void> => {
+    const blob = new Blob([Buffer.from(data, 'base64')], type ? { type } : {});
```
notEduardo: I'm not familiar with the type of data being downloaded, but is it possible that the data or filename could be malicious? if so, we could update this method with some validations 
```suggestion
export const triggerDownload = async (data: string, filename: string, type?: string): Promise<void> => {
    // Validate base64 format
    if (!/^[A-Za-z0-9+/]*={0,2}$/.test(data)) {
        throw new Error('Invalid base64 data');
    }

    // Sanitize filename to prevent directory traversal
    const sanitizedFilename = filename.replace(/[^a-z0-9._-]/gi, '_');

    // Limit file size to prevent DoS (e.g., 10MB)
    const decoded = Buffer.from(data, 'base64');
    if (decoded.length > 10 * 1024 * 1024) {
        throw new Error('File too large to download');
    }

    const blob = new Blob([Buffer.from(data, 'base64')], type ? { type } : {});
```

app/features/idl/ui/IdlCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }, [idl, programMetadataIdl]);
+
+    useEffect(() => {
+        // wait until both data are ready and then activate first available in the array
+        if (tabs.every(tab => tab.idl !== undefined)) {
+            setActiveTabIndex(tabs.findIndex(tab => tab.idl));
+        }
+    }, [tabs]);
+
+    if ((!idl && !programMetadataIdl) || activeTabIndex == undefined) {
```
notEduardo: `==` treats `null` and `undefined` as equal, which could hide bugs. Use strict equality for type safety.
```suggestion
    if ((!idl && !programMetadataIdl) || activeTabIndex === undefined) {
```

askov: In this case it doesn't matter, but thanks for the suggestion. Maybe we should change eslint rules to forbid this usage?

app/shared/lib/triggerDownload.ts
askov: Good catch! In our case, all these parameters aren't related to user input (we define them with code, and if something bad happens, it occurs before this function call). Moreover, this function is just a copy and paste from the already existing production code (see Downloadable.tsx)

I've improved the function to be more robust and secure.

I haven't included any sanitization because modern browsers already do this job well; see more details in the comments. Also, to prevent memory issues, we should validate the size before decoding, as it creates even more problems (see the comments).

app/features/idl/ui/IdlCard.tsx
notEduardo: not sure, up to @Woody4618 @jacobcreech  but this would be the line to add. [More context](https://eslint.org/docs/latest/rules/eqeqeq)
```eslint
"eqeqeq": ["error", "always"]
```

Woody4618: yeah im fine with changing the rule.

PR #793, opened 2025-12-18
chore: Mock test data for the Instruction Cards

shazzzam:
## Description

Mock Solana RPC calls in tests  to avoid 429s in the pipeline and make tests run faster in CI

This PR adds new .env variable  `TEST_USE_REAL_RPC=true` enables real RPC calls and overwrites existing fixtures, whiles false/unset runs tests mocked with data from fixtures

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Protocol integration
-   [ ] Documentation update
-   [x] Other (please describe):

## Checklist

-   [x] CI/CD checks pass

## Additional Notes

----

> [!IMPORTANT]
> Add mock system for Solana RPC calls to improve test reliability and speed by avoiding rate limits.
> 
>   - **Behavior**:
>     - Introduces mock system for Solana RPC calls in `mock-solana-rpc.ts` to avoid 429 errors and speed up tests.
>     - Adds `.env` variable `TEST_USE_REAL_RPC` to toggle real RPC calls.
>   - **Testing**:
>     - Updates tests in `TokenExtensionRow.spec.tsx`, `BaseInstructionCard.spec.tsx`, and `AssociatedTokenDetailsCard.spec.tsx` to use mock connections.
>     - Uses `createMockConnection` and `createRealConnection` to manage RPC call behavior.
>   - **Fixtures**:
>     - Adds JSON fixtures for RPC responses in `fixtures/rpc-responses/` to simulate RPC call data.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fexplorer&utm_source=github&utm_medium=referral)<sup> for 33be33ca7b8d61160a7a0781ad299cab7f35d5b6. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


scripts/record-cluster-fixtures.ts
```diff
@@ -0,0 +1,158 @@
...
+        console.log(`   ✨ Success!`);
+    } catch (error) {
+        console.error(`   ❌ Failed:`, error);
+        throw error;
+    }
+}
+
+// Common test configurations
+const TEST_CONFIGS = {
+    'default-test': {
```
rogaldh, line 118: I think that it is better to name it explicitly the mainnet. Cluster slug from the `/clusters` will fit

scripts/record-cluster-fixtures.ts
```diff
@@ -0,0 +1,158 @@
...
+        cluster: 'mainnet-beta',
+        url: 'https://api.mainnet-beta.solana.com',
```
rogaldh, line 120: Let's use RPC addresses from cluster as well to. Otherwise this may drift upon changes into main RPCs

scripts/record-cluster-fixtures.ts
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env tsx
```
rogaldh, line 1: I like the feature personally, but could you move this into a separate PR, please? That is to narrow the context for changes. Already existing fixtures are good to go

scripts/record-cluster-fixtures.ts
```diff
@@ -0,0 +1,158 @@
...
+
+        // Check for errors
+        if (epochInfoData.error) throw new Error(`getEpochInfo error: ${JSON.stringify(epochInfoData.error)}`);
+        if (epochScheduleData.error)
+            throw new Error(`getEpochSchedule error: ${JSON.stringify(epochScheduleData.error)}`);
+        if (firstBlockData.error)
+            throw new Error(`getFirstAvailableBlock error: ${JSON.stringify(firstBlockData.error)}`);
+
+        // Build fixture data
+        const fixtureData = {
```
rogaldh, line 89: Let's use existing types for epoch and related ones. That is to ensure we write down proper data into fixture

package.json
```diff
@@ -15,8 +15,10 @@
         "sb": "storybook dev -p 6006",
         "scan": "next dev & react-scan localhost:3000",
         "start": "next start",
+        "writeclusterfixtures": "tsx scripts/record-cluster-fixtures.ts all",
```
rogaldh, line 18: Let's move docs how to run this script into `docs/development.md`, please. Small guide how to work with fixtures would be a good addition

app/__tests__/fixtures/cluster-updates/cluster-updates-devnet.json
```diff
@@ -0,0 +1,47 @@
...
+      "__type": "bigint",
+      "__value": "432000"
+    },
+    "warmup": false
+  },
+  "firstAvailableBlock": {
+    "__type": "bigint",
+    "__value": "116113408"
+  }
+}
```
rogaldh, line 47: Could you, please, ensure that all fixtures are created with proper symbol in the end. Looks like fixtures are not the target for prettier atm

app/__tests__/mock-solana-rpc.ts
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+    return result;
+}
+
+function saveFixture(filename: string, data: any): void {
+    const filepath = join(FIXTURES_DIR, `${filename}.json`);
+    // Preprocess to catch PublicKey and other special types before JSON.stringify
+    const preprocessed = preprocessForSerialization(data);
+    writeFileSync(filepath, JSON.stringify(preprocessed, null, 2), 'utf-8');
```
rogaldh, line 128: Could we also minify the fixture?

app/__tests__/mock-solana-rpc.ts
```diff
@@ -0,0 +1,360 @@
...
+        })
+        .join('|');
+
+    let hash = 0;
+    for (let i = 0; i < paramsStr.length; i++) {
+        const char = paramsStr.charCodeAt(i);
+        hash = (hash << 5) - hash + char;
+    }
+
+    return `${method}-${Math.abs(hash).toString(36)}`;
```
rogaldh, line 73: That logic is quite complex. I'd suggest using WEB Crypto API to create hashes. Something like that:

```
    const data = new TextEncoder().encode(JSON.stringify(params));
    const hashBuffer = await crypto.subtle.digest('SHA-256', data);
    const hash = Buffer.from(hashBuffer).toString('hex').slice(0, 12); // 12 to make it more resilient to the overlap
    return `${method}-${hash}`;
```

app/__tests__/fixtures/rpc-responses/common-base-instruction-card-getAddressLookupTable-cuo2mz.json
```diff
@@ -0,0 +1,1010 @@
+{
```
rogaldh, line 1: Is there a reson we write the type for each value into fixture? I think that it is more reasonable to keep it as close to API response as possible and convert into proper type upon reading the fixture

app/components/account/__tests__/TokenExtensionRow.spec.tsx
```diff
@@ -13,6 +14,12 @@ import { TokenExtension } from '@/app/validators/accounts/token-extension';
 
 import { TokenExtensionRow } from '../TokenAccountSection';
 
+vi.mock('@solana/web3.js', async () => {
+    const actual = await vi.importActual<typeof import('@solana/web3.js')>('@solana/web3.js');
```
rogaldh, line 18: Could we import the module via * wildcard at the top and use it as `const actual = await vi.importActual<typeof SolanaKit>();`?

app/__tests__/mock-solana-rpc.ts
```diff
@@ -0,0 +1,360 @@
...
+}
+export const mockPresets = {
+    empty: (testName: string) => createMockConnection(testName),
+    forIdlTests: (testName: string, realConnection?: Connection) => createMockConnection(testName, realConnection),
+    forInstructionTests: (testName: string, realConnection?: Connection) =>
+        createMockConnection(testName, realConnection),
+};
+
+export function createRealConnection(
+    cluster: 'mainnet-beta' | 'testnet' | 'devnet' = 'mainnet-beta'
```
rogaldh, line 349: I think it's better to use already existing aliases for clusters.

app/__tests__/mock-solana-rpc.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    throw new Error(
+        `No fixture found for ${method} with params: ${JSON.stringify(params)}\n` +
+            `Expected fixture file: ${filename}.json\n` +
+            `To record this fixture, run tests with: TEST_USE_REAL_RPC=true pnpm test`
+    );
+}
+
+export function createMockConnection(testName: string, realConnection?: Connection): Connection {
+    if (USE_REAL_RPC && !realConnection) {
```
rogaldh, line 186: Could we, please, move the detection layer for RPC to a top-level component to separate core logic from the check whether is it possible to use it?

app/components/common/__tests__/BaseInstructionCard.spec.tsx
```diff
@@ -13,7 +14,8 @@ describe('BaseInstructionCard', () => {
     test('should render "BaseInstructionCard"', async () => {
         const index = 1;
         const m = mock.deserializeMessageV0(stubs.aTokenCreateIdempotentMsg);
-        const connection = new Connection(clusterApiUrl('mainnet-beta'));
+        const realConnection = createRealConnection('mainnet-beta');
+        const connection = createMockConnection('common-base-instruction-card', realConnection);
```
rogaldh, line 18: What do you think we change drift from the <%test-case-name%> to <%epoch_name%> to create fixtures?
In case we have two different tests created at the same time, they will produce separate fixtures. Correct me if I am wrong here.
That moves the focus from the current cluster state (epoch, for example) to the component and produces extra fixtures likely.

rogaldh (changes requested):
Let's mock responses for Instruciton cards first and add fixture creation within separate PR

app/__tests__/fixtures/rpc-responses/common-base-instruction-card-getAddressLookupTable-cuo2mz.json
shazzzam: Types are stored only for BigInt and PublicKeys, they are needed to restore object using corresponding constructors

rogaldh:
Implemented by https://github.com/solana-foundation/explorer/pull/896

scripts/record-cluster-fixtures.ts
riazulhoque: Done

PR #802, opened 2025-12-30
fix: Improve derivation for PDA at Interactive IDLs and configure analytics

rogaldh:
## Description

PR introduces various fixes into "Interactive IDL" feature to improve it for production usage.
- Fix pda derivation for programs that are specified at seeds.
- Improve UI for Instruction form to allow changing autogenerated fields
- Improve the list of autosuggestions
- Send events to GA to track "Interactive IDL" usage

## Type of change

-   [x] Bug fix
-   [x] New feature

## Screenshots

n/a

## Testing

`pnpm t`

## Related Issues

n/a

## Checklist

-   [x] CI/CD checks pass

----

> [!IMPORTANT]
> Change `sampleRate` to 1 in `createSentryConfig` to enable full issue tracking in Sentry.
> 
>   - **Behavior**:
>     - Change `sampleRate` from 0.1 to 1 in `createSentryConfig` in `config.mjs`, enabling full issue tracking instead of 10%.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fexplorer&utm_source=github&utm_medium=referral)<sup> for 6f24e9cf55b86bec59346a6c6c10bb9fd2fbaf18. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app/features/idl/interactive-idl/model/use-instruction.ts
+++ b/app/features/idl/interactive-idl/model/use-instruction.ts
@@ -469,29 +469,38 @@ function useInvocationState({
 
     return {
         handleSimulatedTxResult,
-
         handleTxEnd,
         handleTxError,
         handleTxStart,
         handleTxSuccess,
-
         isExecuting,
-
         lastResult,
-
         parseLogs,
     };
 }
 
+/**
+ * Extracts custom error code from InstructionError variant.
+ */
 function extractCustomErrorCode(error: TransactionError | null): number | undefined {
-    if (!error || typeof error !== 'object' || !('InstructionError' in error)) {
+    if (!error || typeof error !== 'object') {
         return undefined;
     }
-    const innerError = error['InstructionError'] as [number, unknown];
+
+    if (!('InstructionError' in error)) {
+        return undefined;
+    }
+
+    const innerError = error['InstructionError'];
+    if (!Array.isArray(innerError) || innerError.length < 2) {
+        return undefined;
+    }
+
     const instructionError = innerError[1];
     if (typeof instructionError === 'object' && instructionError !== null && 'Custom' in instructionError) {
         return (instructionError as { Custom: number })['Custom'];
     }
+
     return undefined;
 }
 


Woody4618 (approved):

app/features/idl/interactive-idl/model/pda-generator/program-resolver.ts
```diff
@@ -0,0 +1,49 @@
...
+function isIdlSeedAccount(seed: IdlSeed): seed is IdlSeedAccount {
+    return seed.kind === 'account';
+}
+
+interface ResolverContext {
+    args: Record<string, string | undefined>;
+    accounts: Record<string, string | Record<string, string | undefined> | undefined>;
+}
+
+export function resolveProgramId(
```
Woody4618, line 23: I dont understand this. What is a pdaProgram? Why is pdaProgram of type IdlSeed?

app/features/idl/interactive-idl/model/use-instruction.ts
```diff
@@ -417,7 +434,7 @@ function useInvocationState() {
             if (typeof simulatedTx.value.err === 'string') {
                 errorMessage = simulatedTx.value.err;
             }
-            throw new Error(errorMessage ?? 'Could not simulate transaction');
```
Woody4618, line 420: How can this actually happen? What else can simulatedTx.value.err be if its not null? Is it an object then? Can we handle that properly? Please change it to: Simulated with errors. See logs for details.

app/shared/lib/analytics.ts
```diff
@@ -0,0 +1,95 @@
...
+    return 'dataLayer' in win && Array.isArray(win.dataLayer);
+}
+
+function hasGtag(win: Window): win is Window & { gtag: (...args: unknown[]) => void } {
+    return typeof (win as unknown as { gtag?: unknown }).gtag === 'function';
+}
+
+function isAnalyticsEnabled(): boolean {
+    return (
+        isEnvEnabled(process.env.NEXT_PUBLIC_GOOGLE_TAG_ID) ||
```
Woody4618: We dont have TAG_ID configured. Why is that one needed for tacking?

rogaldh: There is TAG already at analytics script [here](https://github.com/solana-foundation/explorer/blob/master/app/%40analytics/default.js#L5). It is done so to make implementation symmetrical and allow migrating to tag manager in the future

Woody4618: I mean we dont have the tag env variable set in vercel. Do you have it?

app/features/idl/interactive-idl/model/use-instruction.ts
rogaldh: Right. I've reused existing helper to display errors and changed the wording.
Here is the result:

image

app/features/idl/interactive-idl/model/pda-generator/program-resolver.ts
rogaldh: That is about handling program address presented at the [Anchor's IDL](https://www.anchor-lang.com/docs/basics/pda#pda-seeds-in-the-idl).
By default PDA is derived with seeds for the current program, but if program is present, we need to derive the address for this specific program instead.

So the name `pdaProgam` describes tha program that is provided by IDL inside `pda.program` property together with `pda.seeds`. [Here](https://github.com/solana-foundation/anchor/tree/master/ts/packages/anchor/src)
image

This program, for example.
image

app/shared/lib/analytics.ts
rogaldh: Locally, yes. It appeared when we tried to use tag-manager. But it was not enabled on production.
But I see, let's remove it and return to this problem later. Sounds good?

app/features/idl/interactive-idl/model/use-instruction.ts
rogaldh: changed to:
image

Woody4618: awesome thanks

PR #811, opened 2026-01-16
feat: Provide `robots.txt`, `sitemap.xml` and setup BotID

rogaldh:
## Description

⚠️ To enable the feature, set `NEXT_PUBLIC_BOTID_ENABLED=true` into envs for production.

PR introduces several improvements:
- Provide info for crawlers: `robots.txt`
- Generate a `sitemap.xml` for crawlers
- Configure `botid` to disallow several areas for bots
- Fix small style issue when header overlaps the main content on the 404 page

## Type of change

-   [x] Bug fix
-   [x] Other (please describe): chore improvement

## Screenshots

image

## Testing

`pnpm t`

## Related Issues

[HOO-248](https://linear.app/solana-fndn/issue/HOO-248)

## Checklist

-   [x] My code follows the project's style guidelines
-   [x] I have added tests that prove my fix/feature works
-   [x] All tests pass locally and in CI
-   [x] I have updated documentation as needed
-   [x] I have run `build:info` script to update build information
-   [x] CI/CD checks pass
-   [x] I have included screenshots for protocol screens (if applicable)

PR #815, opened 2026-01-20
fix: Improve handling for bots with BotID

rogaldh:
## Description

PR fixes the logic to handle requests detected as bot requests with BotId

## Type of change

-   [x] Bug fix

## Screenshots

n/a

## Testing

- Direct visit to tx page works
- Direct visit to /api route works
- curl request to /api without `x-is-human` header works
- curl request to /api with `x-is-human` header works

If `NEXT_PUBLIC_BOTID_ENABLED=true` (feature is enabled):
- Direct visit to tx page works
- Direct visit to /api route works
- curl request to /api without `x-is-human` header works
- curl request to /api with `x-is-human` header works

if `NEXT_PUBLIC_BOTID_CHALLENGE_MODE_ENABLED=true` (challenge mode is enabled):
- Human requests to /api allowed
- Bot requests to /api blocked with 401
- curl request to /api with `x-is-human` header works if user is not a bot

## Related Issues

#811 

## Checklist

-   [x] CI/CD checks pass

### DIFF
--- a/__tests__/middleware.spec.ts
+++ b/__tests__/middleware.spec.ts
@@ -141,7 +141,9 @@ describe('middleware', () => {
                 expect(body).toEqual({ error: 'Access denied: request identified as automated bot' });
                 expect(loggerWarnSpy).toHaveBeenCalledWith(expect.stringContaining('[middleware] Bot detected'));
                 expect(loggerErrorSpy).toHaveBeenCalledWith(
-                    expect.stringContaining('[middleware] Challenge mode enabled, blocking')
+                    expect.objectContaining({
+                        message: expect.stringContaining('[middleware] Challenge mode enabled, blocking'),
+                    })
                 );
             });
 
@@ -158,7 +160,9 @@ describe('middleware', () => {
 
                 expect(response.status).toBe(401);
                 expect(loggerErrorSpy).toHaveBeenCalledWith(
-                    expect.stringContaining('[middleware] Challenge mode enabled, blocking')
+                    expect.objectContaining({
+                        message: expect.stringContaining('[middleware] Challenge mode enabled, blocking'),
+                    })
                 );
             });
 
@@ -215,7 +219,9 @@ describe('middleware', () => {
 
             expect(response.status).toBe(401);
             expect(loggerErrorSpy).toHaveBeenCalledWith(
-                expect.stringContaining('[middleware] Challenge mode enabled, blocking')
+                expect.objectContaining({
+                    message: expect.stringContaining('[middleware] Challenge mode enabled, blocking'),
+                })
             );
         });
     });


middleware.ts
```diff
@@ -2,44 +2,53 @@ import { isEnvEnabled } from '@utils/env';
...
-    // /tx/* - block bad bots, allow verified crawlers (Googlebot, Bingbot, etc.)
-    if (pathname.startsWith('/tx/')) {
-        if (verification.isBot && !verification.isVerifiedBot) {
-            return new NextResponse('Forbidden', { status: 403 });
+        if (isEnvEnabled(process.env.NEXT_PUBLIC_BOTID_CHALLENGE_MODE_ENABLED)) {
+            console.log(`[middleware] Challenge mode enabled, blocking: ${pathname}`);
+            return NextResponse.json(BOT_RESPONSE.body, { status: BOT_RESPONSE.status });
         }
+    } else {
+        console.log(`[middleware] Human verified, allowing: ${pathname}`);
```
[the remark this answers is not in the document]
rogaldh: console is replaced with Logger that respects project's LOG_LEVEL

app/utils/logger.ts
```diff
@@ -44,6 +44,12 @@ export default class StraightforwardLogger {
         }
         isLoggable(0) && console.error(error, ...other);
```
Woody4618, line 45: Is this correct? I would assume we always want to log all errors. Also not only when its a specific log level

.env.example
```diff
@@ -21,6 +21,8 @@ NEXT_PUBLIC_BAD_TOKENS=
 NEXT_PUBLIC_BOTID_ENABLED=true
 ## Simulate bot behavior locally for testing (dev only)
 NEXT_PUBLIC_BOTID_SIMULATE_BOT=false
+## Enable challenge mode to return status code for detected bots (default: log only)
+NEXT_PUBLIC_BOTID_CHALLENGE_MODE_ENABLED=false
```
Woody4618, line 25: Why do we have another env variable. What does it do. Isnt this a vercel setting and we already have NEXT_PUBLIC_BOTID_ENABLED

middleware.ts
```diff
@@ -1,45 +1,55 @@
...
-    // Feature flag to disable bot protection
+    const { pathname } = request.nextUrl;
+
     if (!isEnvEnabled(process.env.NEXT_PUBLIC_BOTID_ENABLED)) {
         return NextResponse.next();
     }
 
-    const { pathname } = request.nextUrl;
+    // Allow requests without x-is-human header (direct API calls)
+    if (!request.headers.has('x-is-human')) {
```
Woody4618, line 16: Whats this header? Cant any one set their own headers?

middleware.ts
```diff
@@ -1,45 +1,55 @@
...
         }
+    } else {
+        Logger.info(`[middleware] Human verified, allowing: ${pathname}`);
     }
 
     return NextResponse.next();
 }
 
 export const config = {
-    matcher: ['/api/:path*', '/tx/:path*'],
```
Woody4618, line 38: Why only api?

middleware.ts
```diff
@@ -1,45 +1,55 @@
...
         },
     });
 
-    // /api/* - block all bots (verified crawlers shouldn't hit API per robots.txt)
-    if (pathname.startsWith('/api/')) {
-        if (verification.isBot) {
-            return new NextResponse('Forbidden', { status: 403 });
-        }
-    }
+    Logger.info(`[middleware] BotId verification for ${pathname}:`, {
```
Woody4618, line 28: Is this just temporary for the bot implementation? Or should this always be logged?

middleware.ts
```diff
@@ -1,45 +1,55 @@
 import { isEnvEnabled } from '@utils/env';
+import Logger from '@utils/logger';
 import { checkBotId } from 'botid/server';
 import { NextResponse, type NextRequest } from 'next/server';
 
+const BOT_RESPONSE = { body: { error: 'Access denied: request identified as automated bot' }, status: 401 } as const;
+
 export async function middleware(request: NextRequest) {
```
Woody4618, line 8: Vercel docs say this should be in route instead of middle ware i think.

middleware.ts
```diff
@@ -1,45 +1,55 @@
...
-    }
+    Logger.info(`[middleware] BotId verification for ${pathname}:`, {
+        isBot: verification.isBot,
+        isHuman: verification.isHuman,
+        isVerifiedBot: verification.isVerifiedBot,
+        bypassed: verification.bypassed,
+    });
+
+    // Block bots only when challenge mode is enabled
+    if (verification.isBot) {
```
Woody4618, line 36: If the vercel feature is set to logging only. This would still block the bots or not?

middleware.ts
rogaldh: Everyone could send, yes. This is a header that BotId adds a fingerprint to. All subsequent requests made from the pages will include this header.
checkbotid on the server reads the header and processes its contents. It determines whether the user isBot or isHuman and provides us with a proper flag, `isBot`, `isVerifiedBot`, `isHuman`.

.env.example
rogaldh: I've added a separate env to allow responding with a different status code.
`NEXT_PUBLIC_BOTID_ENABLED` turns the feature on and provides logs.
`NEXT_PUBLIC_BOTID_CHALLENGE_MODE_ENABLED` allows us to turn the protection on and respond with 401 to block the access

app/utils/logger.ts
rogaldh: Fair enough. Let's use 0 as the default. Previous logic was: do not log errors with the logger if an env is not provided.
As these codes are enums, zero is the lowest one

middleware.ts
rogaldh: Middleware stands behind every route. Depends on the matcher config. So, to avoid adding protection to every route, we can handle this in the middleware.
I think that is a cleaner solution than to edit every route

middleware.ts
rogaldh: If it is set to "Log only", the protection won't work. It just registers bots.
Turning the Bot Protection On would challenge the user with an extra check. But from my tests, that protection does not affect requests to /api/* made from the Explorer page.

So to block the bots, protection should be on, and also `NEXT_PUBLIC_BOTID_CHALLENGE_MODE_ENABLED=true` is needed

middleware.ts
rogaldh: I put this to see the result for the preview builds. Production should be limited only to errors and this message won't be displayed

middleware.ts
rogaldh: With BotProtection turned on, Vercel would handle bot protection for pages itself. We do not need to call for checkBotId for pages, unless we want to leave "Log only" level and limit access to some pages.

But Vercel won't be able to challenge users for API routes. That is why I enabled middleware only for /api

PR #823, opened 2026-01-26
feat: Add lazy loading for token list requests

C0mberry:
## Description
- adding lazy loading for AddressCard component

## Type of change
-   [x] Bug fix

## Screenshots
Screenshot 2026-01-22 at 19 32 19

## Testing
1. open http://localhost:3000/tx/34qZEWsTW85uZSd4N53DptMaQ5HTx2cczgwWatW6CPhLbaetEX67jELmrCyz2M2ydkQin3NR9sMbgVxeqDwT8Sqp
2. check the console

## Related Issues
[HOO-214](https://linear.app/solana-fndn/issue/HOO-214/unable-to-load-the-token-list-on-the-transaction-page)

## Checklist

-   [x] CI/CD checks pass

### DIFF
--- a/app/components/common/Address.tsx
+++ b/app/components/common/Address.tsx
@@ -130,12 +130,12 @@ export function Address({
     );
 
     return (
-        <div ref={containerRef}>
+        <span ref={containerRef}>
             <div className={`d-none d-lg-flex align-items-center ${alignRight ? 'justify-content-end' : ''}`}>
                 {content}
             </div>
             <div className="d-flex d-lg-none align-items-center">{content}</div>
-        </div>
+        </span>
     );
 }
 
--- a/app/components/inspector/stories/AccountsCard.stories.tsx
+++ b/app/components/inspector/stories/AccountsCard.stories.tsx
@@ -111,7 +111,6 @@ export const Default: Story = {
                 </MockAccountsProvider>
             </ClusterProvider>
         ),
-        withTokenInfoBatch,
     ],
     play: async ({ canvasElement }) => {
         const canvas = within(canvasElement);
--- a/app/entities/token-info/model/token-info-batch-provider.tsx
+++ b/app/entities/token-info/model/token-info-batch-provider.tsx
@@ -2,7 +2,7 @@
 
 import { Cluster } from '@utils/cluster';
 import { getTokenInfoSwrKey } from '@utils/token-info';
-import React, { createContext, useCallback, useContext, useRef } from 'react';
+import React, { createContext, useCallback, useContext, useEffect, useRef } from 'react';
 import { mutate } from 'swr';
 
 import { getTokenInfos } from '../lib/fetch-token-mints';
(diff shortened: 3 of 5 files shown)


app/components/common/Address.tsx
```diff
@@ -163,27 +163,15 @@ const useTokenMetadata = (useMetadata: boolean | undefined, pubkey: string) => {
 };
 
 const useTokenInfo = (fetchTokenLabelInfo: boolean | undefined, pubkey: string) => {
-    const [info, setInfo] = useState<TokenLabelInfo>();
-    const { cluster, url } = useCluster();
+    let batchContext;
+    try {
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        batchContext = useTokenBatch();
```
rogaldh: I'd suggest handling the errors inside the hook. Breaking the rule of hooks is error-prone.

app/hooks/use-token-info.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { getTokenInfoWithoutOnChainFallback } from '@/app/utils/token-info';
+
+export function useTokenInfo(address: string | undefined, cluster: Cluster) {
+    const swrKey = address ? ['token-info', address, cluster] : null;
+
+    return useSWR<Token | undefined>(
+        swrKey,
+        async () => {
+            if (!address) return undefined;
```
rogaldh: There is no need for this condition to have the swrKey = null in that case. The fetcher won't be triggered.

app/hooks/use-token-info.ts
```diff
@@ -0,0 +1,24 @@
...
+            return await getTokenInfoWithoutOnChainFallback(new PublicKey(address), cluster);
+        },
+        {
+            dedupingInterval: 5 * 60 * 1000,
+            focusThrottleInterval: 5 * 60 * 1000,
+            revalidateOnFocus: false,
+            revalidateOnMount: false,
```
rogaldh: Could you, please, clarify this condition? If I am not mistaken, that won't trigger the request upon initialization, which is good. But setting `revalidateOnFocus: false` means revalidation won't be triggered on focus either.

I think we need swr and IntersectionObserver for the Address component together.

C0mberry: With `revalidateOnFocus: false` swr will not refetch when the window or tab regains focus

rogaldh: Yeah. I mean, we should set it to true or leave it unset. Same thing afaik

app/components/common/Address.tsx
```diff
@@ -47,6 +47,34 @@ export function Address({
     const addressPath = useClusterPath({ pathname: `/address/${address}` });
     const [showNicknameEditor, setShowNicknameEditor] = useState(false);
     const nickname = useNickname(address);
+    const containerRef = useRef<HTMLDivElement>(null);
+    const [isVisible, setIsVisible] = useState(false);
+
+    useEffect(() => {
```
askov: **issue**: The PR code is fine. However, the component is quite complex, difficult to understand, and lacks tests. The biggest issue is that the changes don’t solve the problem when I test the deployed version; there are still many failed requests when I scroll down, even slowly. I believe this problem requires more attention and possibly some refactoring.

Woody4618 (approved):

PR #831, opened 2026-01-28
feat: Add shareable receipts

rogaldh:
## Description
Adds shareable receipt views and images for Solana transactions. Transaction links can now generate OG images showing sender, receiver, amount, fee, and memo (for SOL and SPL/Token-2022 transfers).

- **OG image API**: `GET /api/og/receipt/[signature]` returns a PNG receipt image (edge, 30min cache, ETag).
- **Receipt feature**: `createReceipt(signature)` fetches tx, parses SystemProgram and token transfer instructions, resolves primary domains, validates with schemas, and returns formatted receipt data.
- **UI**: `BaseReceipt` / `BaseReceiptImage` components plus Storybook stories and fixtures.

## Type of change
-   [x] New feature

### OG Image ([example](http://localhost:3000/api/og/receipt/5CBBFXjdu32noWfSdzAdiEBad73ZmMChN17CEFeTLiQQJZBMzL4J9dj7YLcaag8z6xKiBws7GVPeQ7KmUGiozC4Z))
| Receipt       |  Placeholder        |
|--------------------|---------------------|
| Screenshot 2026-01-27 at 12 08 48 | image |

### Receipt view ([example](http://localhost:3000/tx/5CBBFXjdu32noWfSdzAdiEBad73ZmMChN17CEFeTLiQQJZBMzL4J9dj7YLcaag8z6xKiBws7GVPeQ7KmUGiozC4Z?view=receipt&cluster=devnet))
| Receipt       |  Placeholder        |
|--------------------|---------------------|
|   Screenshot 2026-01-27 at 12 09 11  |  Screenshot 2026-01-27 at 12 08 23 |

### Environment
- `NEXT_PUBLIC_RECEIPT_ENABLED` — enable receipt view and shareable OG images
- `RECEIPT_CLUSTER_PROBE_ENABLED` — when tx is not on mainnet, look it up on devnet/testnet
- `RECEIPT_BASE_URL` — base URL for previews (defaults to explorer base URL)
- `RECEIPT_OG_IMAGE_VERSION` — optional; bust third-party caches when the image changes
- `RECEIPT_CACHE_HEADERS` — optional Cache-Control override for OG images

**Vercel Firewall Configuration Required:** If Attack Challenge Mode is enabled, add a bypass rule for `/og/` path in Vercel Dashboard -> Firewall -> Add New... -> Rule, otherwise social media crawlers won't be able to fetch preview images.

Name: Allow bots for OG images
Description: Allow social media crawlers to fetch OG image previews
Rule:
```
If `Request Path` `Starts with` `/og/receipt/`
    Then `Bypass`
```

Name: Allow bots to visit receipt pages
Description: Allow social media crawlers to visit receipt pages
Rule:
```
If `Request Path` `Starts with` `/tx/`
    AND `Query` `view` `Equals` `receipt`
    Then `Bypass`
```

⚠️ Important note: `Log` mode does not allow the preview to work. `Bypass` is needed

### Storybook
- http://localhost:6006/?path=/story/features-receipt-receiptimage--default
- http://localhost:6006/?path=/story/features-receipt-basereceiptimage--default

### Example (SOL)
- Page http://localhost:3000/tx/pBApPVmSKwvLmGahzW7ZKHmqLuPNXpwJR3n37HQ4B279AG4JmfUWJRd9NnaaQAxJZ5npdoFj6r6iZ2cvoYdorFg?view=receipt&cluster=devnet

### Example (USDC)
- Page http://localhost:3000/tx/2cQ1xf4meFvQBcmhyAtnXQzWbqNWvy3eQrgWPov3EQbnj2WGpLzChjJrVopFURRCWeNyKe8grXk4hmcv3EFd7iA8?view=receipt&cluster=devnet

### Test signatures
| Transaction Signature | Description |
|----------------------|-------------|
| 4Y4sN6zmSQJPegVTFLZ3ctLfTVs5tA4asjurL6RY8jhLVxZiy5po38qGAbZBytLHpjWh6jAMshGcZV8QfGvYtznE | With memo |
| 5yKzCuw1e9d58HcnzSL31cczfXUux2H4Ga5TAR2RcQLE5W8BiTAC9x9MvhLtc4h99sC9XxLEAjhrXyfKezdMkZFV | s0 to s1, to s2, to s1 - shows placeholder |
| pBApPVmSKwvLmGahzW7ZKHmqLuPNXpwJR3n37HQ4B279AG4JmfUWJRd9NnaaQAxJZ5npdoFj6r6iZ2cvoYdorFg | single transfer |
| 2cQ1xf4meFvQBcmhyAtnXQzWbqNWvy3eQrgWPov3EQbnj2WGpLzChjJrVopFURRCWeNyKe8grXk4hmcv3EFd7iA8 | USDC simple (transferChecked) |
| 5CBBFXjdu32noWfSdzAdiEBad73ZmMChN17CEFeTLiQQJZBMzL4J9dj7YLcaag8z6xKiBws7GVPeQ7KmUGiozC4Z | USDC transfer not checked |
| 5ajzgLXt9GJNRmQCNT2hx9ruJRcg8XLweB7FRgQsCxo9JbsVKDKpnDir4wRJihZTVuTdCZuXURiQD1hEbjHoJL6x | receiver is the fee payer |
| 2MDTSo3dg7UxMnMhQVMVT2uVs5HzKrFjgoRmPwM8g951RZLhGeRcxZc9d1StzFR2FAmfKX8kbB5ZTjcu5vpXKAVN | send from s0 to s2 and fee payer s1 |
| 22xruGxWMALDoVksikb6kyv7wigrqQSEtmM8pbazAjmM3uSnvugvXhcmGhFhWd9uuUJDsegfddznFPC9f5Qpx8cq | zero transfer - placeholder |
| 4izwTCUeRGAMGReXeXDumiBzAgXPGz6KzccacCf1WU5YXCCpSDjAQT7J6D6dY45bL1NW9AiqwCuWEnz3hbGtZS2y | tx with .sol name |

## Related Issues
[HOO-245](https://linear.app/solana-fndn/issue/HOO-245/add-shareable-receipts-for-transactions-in-solana-explorer)
[HOO-260](https://linear.app/solana-fndn/issue/HOO-260/implement-receipt-view-for-the-transaction-page)

## Checklist

-   [x] CI/CD checks pass
-   [x] I have included screenshots for protocol screens

### DIFF
--- a/app/features/receipt/__e2e__/receipt.e2e.ts
+++ b/app/features/receipt/__e2e__/receipt.e2e.ts
@@ -2,57 +2,59 @@ import { expect, Page, test } from '@playwright/test';
 
 test.describe.configure({ retries: 2 });
 
+/** Delay between tests to avoid RPC rate limiting (429) */
+const RPC_COOLDOWN_MS = 500;
+const NAV_TIMEOUT = 60000;
+const CONTENT_TIMEOUT = 15000;
+
 const VALID_TX = '5VERv8NMvzbJMEkV8xnrLkEaWRtSz9CosKDYjCJjBRnbJLgp8uirBgmQpjKhoR4tjF3ZpRzrFmBV6UjKdiSZkQUW';
 const INVALID_TX = '34qZEWsTW85uZSd4N53DptMaQ5HTx2cczgwWatW6CPhLbaetEX67jELmrCyz2M2ydkQin3NR9sMbgVxeqDwT8Sqp';
 const FEATURE_ENABLED = process.env.NEXT_PUBLIC_RECEIPT_ENABLED === 'true';
 
-async function hasElement(page: Page, selector: string, timeout = 10000): Promise<boolean> {
-    try {
-        await page.locator(selector).waitFor({ state: 'visible', timeout });
-        return true;
-    } catch {
-        return false;
-    }
-}
-
-async function waitForPage(page: Page, tx: string, view?: 'receipt') {
-    const url = view ? `/tx/${tx}?view=${view}` : `/tx/${tx}`;
-
-    const responsePromise = page
-        .waitForResponse(response => response.url().includes('api.') || response.url().includes('rpc'), {
-            timeout: 60000,
-        })
-        .catch(() => null);
-
-    await page.goto(url, { timeout: 60000, waitUntil: 'domcontentloaded' });
-    await responsePromise;
-
-    await page
-        .locator('text=Loading')
-        .waitFor({ state: 'hidden', timeout: 30000 })
-        .catch(() => {});
-}
-
 test.describe('receipt feature validation', () => {
+    test.afterEach(async () => {
+        await delay(RPC_COOLDOWN_MS);
+    });
+
     test('respects NEXT_PUBLIC_RECEIPT_ENABLED flag', async ({ page }) => {
         await waitForPage(page, VALID_TX, 'receipt');
 
+        await page
+            .locator('h3:has-text("Solana Receipt")')
+            .or(page.locator('h2:has-text("Transaction")'))
+            .or(page.locator('text=There is no receipt'))
+            .first()
+            .waitFor({ state: 'visible', timeout: CONTENT_TIMEOUT });
+
         const hasReceipt = await hasElement(page, 'h3:has-text("Solana Receipt")');
+        const hasNoReceipt = await hasElement(page, 'text=There is no receipt');
+        const hasTransactionPage = await hasElement(page, 'h2:has-text("Transaction")');
 
         if (FEATURE_ENABLED) {
-            expect(hasReceipt).toBe(true);
+            expect(hasReceipt || hasNoReceipt).toBe(true);
         } else {
             expect(hasReceipt).toBe(false);
-            expect(await hasElement(page, 'h2:has-text("Transaction")')).toBe(true);
+            expect(hasTransactionPage).toBe(true);
         }
     });
 });
 
 test.describe('when feature enabled', () => {
+    test.describe.configure({ mode: 'serial' });
     test.skip(!FEATURE_ENABLED);
+    test.afterEach(async () => {
+        await delay(RPC_COOLDOWN_MS);
+    });
 
     test('renders receipt for valid transaction', async ({ page }) => {
         await waitForPage(page, VALID_TX, 'receipt');
+
+        await page
+            .locator('h3:has-text("Solana Receipt")')
+            .or(page.locator('text=Not Found'))
+            .or(page.locator('text=There is no receipt'))
+            .first()
+            .waitFor({ state: 'visible', timeout: CONTENT_TIMEOUT });
 
         const hasReceipt = await hasElement(page, 'h3:has-text("Solana Receipt")');
         const hasError = await hasElement(page, 'text=Not Found');
@@ -70,6 +72,14 @@ test.describe('when feature enabled', () => {
     test('handles invalid transaction', async ({ page }) => {
         await waitForPage(page, INVALID_TX, 'receipt');
 
+        await page.waitForFunction(
+            () => {
+                const text = document.body?.innerText || '';
+                return text.includes('There is no receipt') || text.includes('Fetch Failed') || text.includes('Error');
+            },
+            { timeout: CONTENT_TIMEOUT }
+        );
+
         const text = await page.textContent('body');
         const showsError =
             text?.includes('There is no receipt') || text?.includes('Fetch Failed') || text?.includes('Error');
@@ -80,12 +90,19 @@ test.describe('when feature enabled', () => {
     test('shows View Receipt button', async ({ page }) => {
         await waitForPage(page, VALID_TX);
 
+        await page
+            .locator('h3:has-text("Overview")')
+            .or(page.locator('text=Fetch Failed'))
+            .or(page.locator('text=Not Found'))
+            .first()
+            .waitFor({ state: 'visible', timeout: CONTENT_TIMEOUT });
+
         const hasOverview = await hasElement(page, 'h3:has-text("Overview")');
 
         if (hasOverview) {
-            expect(await hasElement(page, 'a:has-text("View Receipt")')).toBe(true);
+            await expect(page.locator('a:has-text("View Receipt")')).toBeVisible({ timeout: CONTENT_TIMEOUT });
         } else {
-            // just verify no error page
+            // just verify no error page (Overview didn't load - page may have errored)
             const bodyText = await page.textContent('body');
             expect(bodyText?.includes('Fetch Failed') || bodyText?.includes('Not Found')).toBe(true);
         }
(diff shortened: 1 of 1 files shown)


app/features/receipt/api/get-tx.ts
```diff
@@ -0,0 +1,90 @@
...
+    signature: string,
+    dependencies?: {
+        findCluster?: (signature: string) => Promise<Cluster | undefined>;
+        fetchDetails?: (signature: string, rpcUrl: string) => Promise<ParsedTransactionWithMeta>;
+    }
+): Promise<ApiData> {
+    const findClusterFn = dependencies?.findCluster ?? findTransactionCluster;
+    const fetchDetailsFn = dependencies?.fetchDetails ?? fetchTransactionDetails;
+
+    const cluster = await findClusterFn(signature);
```
rogaldh: Let's validate the signature. It may be invalid, and we will make requests to the chain.

app/features/receipt/api/get-tx.ts
```diff
@@ -0,0 +1,90 @@
...
+        throw new Error('Transaction not found');
+    }
+
+    return { cluster, transaction };
+}
+
+async function findTransactionCluster(signature: string): Promise<Cluster | undefined> {
+    const clustersToTry: Cluster[] = [Cluster.MainnetBeta, Cluster.Devnet, Cluster.Testnet];
+
+    const clusterChecks = clustersToTry.map(async (cluster): Promise<Cluster | undefined> => {
```
rogaldh: I think this is not optimal. To wait while all the promises are settled. I could propose:
- Pass a cluster flag together with the signature to fetch it from the specific cluster.
- Make a request to the Mainnet first and then ask for other clusters.

The first variant looks a bit better.

app/features/receipt/model/sol-transfer.ts
```diff
@@ -0,0 +1,69 @@
...
+        type: 'sol',
+    };
+}
+
+// We support only single sol transfer instruction per transaction by design.
+function getSingleSolTransferInstruction(transaction: ParsedTransactionWithMeta): SolTransferInstruction | undefined {
+    const instructions = transaction.transaction.message.instructions.filter(
+        (instruction): instruction is SolTransferInstruction => isSolTransfer(instruction)
+    );
+    return instructions.length === 1 ? instructions[0] : undefined;
```
rogaldh, line 35: A common scenario is to tip the transaction through the Jito program, and that might be a transfer as well. That means these transactions won't be recognized as receipts

rogaldh: Here is the list of Jito accounts https://www.quicknode.com/docs/solana/getTipAccounts

app/features/receipt/model/create-receipt.ts
```diff
@@ -0,0 +1,78 @@
...
+export async function createReceipt(signature: string): Promise<FormattedReceipt | undefined> {
+    const data = await getTx(signature);
+    return extractReceiptData(data.transaction, data.cluster);
+}
+
+export async function extractReceiptData(
+    tx: ParsedTransactionWithMeta,
+    cluster: Cluster
+): Promise<FormattedReceipt | undefined> {
+    let receipt: Receipt | undefined = createSolTransferReceipt(tx);
```
rogaldh: I think we should parse the Token transfer first. T22 transfer + jito tip would be displayed like sol transfer and not the token one

app/features/receipt/model/create-receipt.ts
```diff
@@ -0,0 +1,78 @@
...
+        receipt = await createTokenTransferReceipt(tx, (mint: string | undefined) => getParsedTokenInfo(mint, cluster));
+    }
+    if (!receipt) return undefined;
+
+    return formatReceiptData(receipt, cluster);
+}
+
+export function formatReceiptData(receipt: Receipt, cluster: Cluster): FormattedReceipt {
+    const timestamp = receipt.date * 1000;
+    const unit = receipt.type === 'sol' ? 'SOL' : receipt.symbol || 'TOKEN';
```
rogaldh: Maybe we need to move this check to some helper, as I've seen it in several places (type === 'sol')

askov: Hm, I'm not sure it has any benefits right now, but I've created guards for clarity and future use.

app/features/receipt/api/get-tx.ts
askov: For now, I have implemented the second option—sequential trying, as I don’t get the first idea: to pass the cluster flag, we need to get this cluster somewhere, and we rely only on a signature from the URL. Are you suggesting to get it from the URL?

app/features/receipt/api/get-tx.ts
askov: Makes sense, thanks, added validation

app/features/receipt/model/sol-transfer.ts
askov: Added jito filter

app/features/receipt/model/create-receipt.ts
askov: Updated

app/features/receipt/model/create-receipt.ts
2c9jj9rmgj-code: no you don't just had to tell me you were needing me to check it and maybe give me my cut so i get paid to?

2c9jj9rmgj-code: i'm late on the messages my bad but you didn't need to cut me out man.. i don't know how to check them i was just doing all of this to get my money moved from a trading app over seas i'm not a coder

app/features/receipt/api/get-tx.ts
rogaldh: Yes, I mean the url. Users have a link to the transaction/receipt. For devnet/testnet, we have a cluster parameter, so we already have the cluster URL.
No cluster param = this is mainnet 100%.

To preserve existing logic (Mainnet first, other clusters second), I think we could do it like this:
- request a tx from Mainnet, or if a query param with a specific cluster is provided at URL > request a tx from the specific cluster
- Check other clusters as they are implemented, and there were no queryparam with cluster (to not repeat the request to the same cluster).

What do you think?

rogaldh: I've added two commits with a small improvement to the cluster probing. Please take a look. Open for suggestions

app/features/receipt/model/create-receipt.ts
rogaldh: No clue what you're talking about.

app/og/receipt/[signature]/route.tsx
```diff
@@ -0,0 +1,80 @@
...
+        const status = statusFromError(message);
+        const body = status === 404 ? 'Receipt not found' : 'Failed to process request';
+        return new NextResponse(body, { status });
+    }
+}
+
+function statusFromError(message: string): number {
+    if (message === 'Transaction not found' || message === 'Cluster not found') return 404;
+    if (message === 'Failed to fetch transaction') return 502;
+    return 500;
```
Comment (rogaldh) [inline app/og/receipt/[signature]/route.tsx]: I'd suggest using Errors from the clients to check for specific errors instead of comparing strings

Comment (askov) [reply app/og/receipt/[signature]/route.tsx]: I agree that using message string matching is not the best practice, but the mix of possible client errors here is pretty overwhelming, and I don't think we should expose those to the route. I created an abstract error that allows us to focus on low-level errors in a single place, still not using them anywhere they shouldn't matter. Let me know if you have any other ideas.

Comment (rogaldh) [reply app/og/receipt/[signature]/route.tsx]: I agree that we may have different sources of errors. Also, we know some errors come from the @solana client. It exposes instances of errors. I'd suggest checking for specific ones and returning exact codes. All other cases might return just 500.
That allows not only predictable behaviour when handling errors, but also the ability to see other possible cases. As you'll see 500 and that measn that the error comes from different place

Comment (askov) [reply app/og/receipt/[signature]/route.tsx]: We decided to improve it as a separate [task](https://linear.app/solana-fndn/issue/HOO-274/improve-handling-for-errors-at-the-receipts-by-reusing-known-errors)

Woody4618 (approved):

app/og/receipt/[signature]/route.tsx
Comment (rogaldh) [reply app/og/receipt/[signature]/route.tsx]: Agreed. Closing convo

Woody4618 (approved):

PR #853, opened 2026-02-20
fix: resolve ipfs:// URIs to HTTP gateway for NFT images and metadata…

### LINKED ISSUE #807: [Bug] Solana Explorer fails to load NFT images stored on IPFS (ipfs:// URI not supported)
**Describe the bug**
Solana Explorer fails to display NFT images when the metadata image field points to an `ipfs://` URI. 
Modern browsers (Chrome, Firefox, Safari) do not support fetching `ipfs://` schemes directly, causing image and metadata fetch failures in Explorer.

The browser console logs:
"Fetch API cannot load ipfs://<CID>. URL scheme 'ipfs' is not supported."

**To Reproduce**

1. Go to https://explorer.solana.com
2. Open a compressed NFT that has its image stored using an `ipfs://` URI
3. Observe that the NFT image does not render
4. Open DevTools → Console
5. See error: `Fetch API cannot load ipfs://... URL scheme 'ipfs' is not supported`

**Example Links**

- https://explorer.solana.com/address/7rAhawfSxmqxjmSayjkLD44GoJA772R2QXBg174bobFW?cluster=devnet

**Expected behavior**
Solana Explorer should correctly display NFT images stored on IPFS by resolving `ipfs://` URIs through a public IPFS HTTP gateway (e.g. https://ipfs.io/ipfs/<CID>), similar to other NFT marketplaces and explorers.

**Screenshots**
Console error:
"Fetch API cannot load ipfs://<CID>. URL scheme 'ipfs' is not supported."

Image

**Additional context**
- This affects Metaplex NFTs (including compressed NFTs) whose metadata uses the standard `ipfs://` scheme.
- Most NFT platforms already normalize IPFS URIs to HTTP gateways before fetching.
- Without this fix, many valid NFTs appear broken on Solana Explorer.

holps-7:
## Description

This PR resolves an issue where the Explorer fails to load NFT images and metadata that use the `ipfs://` protocol scheme. Modern web browsers do not natively support the `ipfs://` scheme, causing fetch and rendering failures for compressed and standard NFTs relying on it.

This change intercepts `ipfs://` URIs in the [getProxiedUri](https://github.com/holps-7/explorer/blob/0249fd3c0b96b4b4741f56af3e485716aa6ebdfc/app/features/metadata/utils.ts) utility and converts them to use a public HTTP gateway (`https://ipfs.io/ipfs/`) before any proxy logic is evaluated.

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Protocol integration
-   [ ] Documentation update

## Screenshots

Before
before image

After
after image

## Testing

- Added unit tests in [utils.spec.ts](https://github.com/holps-7/explorer/blob/0249fd3c0b96b4b4741f56af3e485716aa6ebdfc/app/features/metadata/__tests__/utils.spec.ts) to strictly verify IPFS rewritten structures (`ipfs://{hash}` and `ipfs://ipfs/{hash}`) under both proxy-enabled and proxy-disabled environments.
- Ran `pnpm test` locally to ensure all suites pass.
- Verified manually by running the Explorer locally (`pnpm dev`) and checking an NFT with an `ipfs://` image URI (`7rAhawfSxmqxjmSayjkLD44GoJA772R2QXBg174bobFW` on devnet) to confirm it visually renders without console errors.

## Related Issues

Fixes #807

## Checklist

-   [x] CI/CD checks pass
-   [ ] I have included screenshots for protocol screens (if applicable)

## Additional Notes

We opted to fix this broadly at the [getProxiedUri](https://github.com/holps-7/explorer/blob/0249fd3c0b96b4b4741f56af3e485716aa6ebdfc/app/features/metadata/utils.ts) entry point rather than per-component, ensuring the HTTP gateway conversion cleanly applies to all metadata API fetchers and image rendering tools simultaneously.


app/features/metadata/utils.ts
```diff
@@ -13,6 +9,23 @@ export const getProxiedUri = (uri: string): string | '' => {
...
+    if (url.protocol === 'ipfs:') {
+        let path = url.host + url.pathname;
+        if (path.startsWith('ipfs/')) {
+            path = path.replace(/^ipfs\//, '');
+        }
+        uri = `https://ipfs.io/ipfs/${path}${url.search}`;
+        try {
+            url = new URL(uri);
+        } catch {
+            // Should not happen, but safe fallback
```
askov: **suggestion**: The flow would be clearer if we just returned in the catch block, since that’s what should actually happen

app/features/metadata/utils.ts
```diff
@@ -13,6 +9,23 @@ export const getProxiedUri = (uri: string): string | '' => {
         throw new Error(`Could not construct URL for "${uri}"`);
     }
 
+    if (url.protocol === 'ipfs:') {
+        let path = url.host + url.pathname;
+        if (path.startsWith('ipfs/')) {
+            path = path.replace(/^ipfs\//, '');
```
askov: **suggestion**: The guard does nothing in this case, 
```ts 
if (path.startsWith('ipfs/')) {
      path = path.replace(/^ipfs\//, '');
  }
```


  is equivalent to just:
```ts
  path = path.replace(/^ipfs\//, '');
```

app/features/metadata/utils.ts
```diff
@@ -13,6 +9,23 @@ export const getProxiedUri = (uri: string): string | '' => {
...
+        }
+        uri = `https://ipfs.io/ipfs/${path}${url.search}`;
```
askov: **thought**: The function mutates both uri and url through reassignment, making it hard to follow which value each variable holds at any given point.

app/features/metadata/utils.ts
```diff
@@ -13,6 +9,23 @@ export const getProxiedUri = (uri: string): string | '' => {
         throw new Error(`Could not construct URL for "${uri}"`);
     }
 
+    if (url.protocol === 'ipfs:') {
```
askov, line 21: **thought**: Thanks for the fix! This is really valuable. Do you think we could improve readability a bit? Just an example:

```ts
const IPFS_GATEWAY = 'https://ipfs.io/ipfs';

export const getProxiedUri = (uri: string): string | '' => {
    if (uri === '') return '';

    let url: URL;
    try {
        url = new URL(uri);
    } catch {
        throw new Error(`Could not construct URL for "${uri}"`);
    }

    const isProxyEnabled = process.env.NEXT_PUBLIC_METADATA_ENABLED === 'true';

    if (url.protocol === 'ipfs:') {
        const gatewayUri = resolveIpfsUri(url);
        return isProxyEnabled ? proxyUri(gatewayUri) : gatewayUri;
    }

    if (!isProxyEnabled) return uri;

    if (!['http:', 'https:'].includes(url.protocol)) return uri;

    return proxyUri(uri);
};

const resolveIpfsUri = (url: URL): string => {
    const path = (url.host + url.pathname).replace(/^ipfs\//, '');
    return `${IPFS_GATEWAY}/${path}${url.search}`;
};

const proxyUri = (uri: string): string => `/api/metadata/proxy?uri=${encodeURIComponent(uri)}`
```

markusknudtsen-droid (approved):

app/features/metadata/utils.ts
holps-7: Agreed, updated it to return the original URI

app/features/metadata/utils.ts
holps-7: Removed the redundant if statement

app/features/metadata/utils.ts
holps-7: Great suggestion, that's much clearer, will keep this noted for future contributions. Applied all the suggestions....

askov (dismissed):

app/features/metadata/utils.ts
```diff
@@ -13,7 +11,23 @@ export const getProxiedUri = (uri: string): string | '' => {
...
+    if (!isProxyEnabled) return uri;
+
     if (!['http:', 'https:'].includes(url.protocol)) return uri;
 
-    return `/api/metadata/proxy?uri=${encodeURIComponent(uri)}`;
+    return proxyUri(uri);
 };
+
+const resolveIpfsUri = (url: URL): string => {
+    const path = (url.host + url.pathname).replace(/^ipfs\//, '');
```
rogaldh: The latest eslint config requires a reason for using regular expressions.

app/features/metadata/utils.ts
```diff
@@ -13,7 +11,23 @@ export const getProxiedUri = (uri: string): string | '' => {
...
+    return `${IPFS_GATEWAY}/${path}${url.search}`;
```
rogaldh: This helper modifies the URL and produces the CID, but there is no validation that the CID is created correctly. That requires the incoming url to be a correct address. As this helper handles the strings from user input, they might be incorrect.
I'd suggest checking that the result is a proper CID resource

app/features/metadata/utils.ts
holps-7: added comment above regular expression to fix style issues

app/features/metadata/utils.ts
holps-7: The CID extracted here comes directly from on-chain metadata, not raw user input, so we trust it's already well-formed and checked while asset was minted by the developer.
The helper doesn't produce or modify the CID — it just strips the redundant `ipfs/` prefix that some clients include `(e.g. ipfs://ipfs/QmABC → QmABC)` before appending it to a standard HTTP gateway URL. A malformed CID would simply result in a 404 from the gateway. Adding CID validation adds complexity without meaningful benefit here.

rogaldh: Ok. My point was not to allow making a request to the resource with a malformed CID. Metadata files are created by users, and from that perspective, the data could be wrong.

holps-7: To verify whether a CID is not malformed, we can use the external package `multiformats`. The `CID.parse` method will throw an error if the CID is invalid, so successfully parsing it indicates the CID is well-formed.

```js
import { CID } from 'multiformats/cid';

const verifyCID = (cid: string): boolean => {
  try {
    CID.parse(cid);
    return true;
  } catch {
    return false;
  }
};
```
> @rogaldh I'm not sure if it’s ok to use an external package for this task.

scutuatua-crypto (approved):

app/features/metadata/utils.ts
rogaldh: lgtm. I think that having a validation with a specified package is a proper direction. Would love to see it here.

holps-7: ok moving ahead with developing function to validate CID. in case the CID is malformed it should skip the fetch and ```console.warn(`[metadata] Cannot fetch a malformed CID: ${<CID>}`)```

holps-7: hey @rogaldh the CID verification is implemented.

package.json
```diff
@@ -90,6 +90,7 @@
         "lighthouse-sdk": "2.0.1",
         "micromatch": "4.0.8",
         "moment": "2.29.4",
+        "multiformats": "^13.4.2",
```
rogaldh: Could you, please, remove ^ from the version? We will unlikely autoupdate the version by semver's pattern

app/features/metadata/utils.ts
```diff
@@ -13,7 +13,38 @@ export const getProxiedUri = (uri: string): string | '' => {
...
 
-    return `/api/metadata/proxy?uri=${encodeURIComponent(uri)}`;
+    return proxyUri(uri);
 };
+
+const resolveIpfsUri = (url: URL): string => {
+    // eslint-disable-next-line no-restricted-syntax -- Strips redundant "ipfs/" prefix from the path for a clean gateway URL.
+    const path = (url.host + url.pathname).replace(/^ipfs\//, '');
+    if (!verifyCID(path)) {
+        console.warn(`[metadata] Cannot fetch a malformed CID: ${path}`);
```
rogaldh: Could you, please, check the latest master? Using plain console is deprecated. Please use the project's Logger instead.

app/features/metadata/utils.ts
```diff
@@ -13,7 +13,38 @@ export const getProxiedUri = (uri: string): string | '' => {
...
+        return '';
+    }
+    return `${IPFS_GATEWAY}/${path}${url.search}`;
+};
+
+const proxyUri = (uri: string): string => `/api/metadata/proxy?uri=${encodeURIComponent(uri)}`;
+
+export const verifyCID = (cid: string): boolean => {
```
rogaldh: verifyCID is used only locally. There is no need to export, I suppose

package.json
holps-7: done

app/features/metadata/utils.ts
holps-7: done, I wasn't aware that plain console is deprecated, now using `Logger.warn()` instead of `console.warn()`

app/features/metadata/utils.ts
holps-7: makes sense, I've removed the export keyword since it's only used locally in this file.

rogaldh (dismissed):
Hi. To merge this feature, we have to improve the metadata proxy endpoint. After completing this, this PR could be merged.
Thank you for contribution.

fyi @Woody4618

askov:
Finally, we’re ready for this PR. @holps-7, please rebase onto the current master - there are some conflicts. If you don’t mind, an OpenSpec proposal doc would be nice (but it’s not required).

package.json
```diff
@@ -105,7 +105,10 @@
         "jspdf": "4.2.1",
         "lighthouse-sdk": "2.0.1",
         "micromatch": "4.0.8",
+        "moment": "2.29.4",
+        "multiformats": "13.4.2",
         "next": "16.2.6",
```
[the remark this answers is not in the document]
holps-7: solved here [8dc3f1a](https://github.com/solana-foundation/explorer/pull/853/commits/8dc3f1a0ed3bb50b84b8dc1e26c7099568935b1d)

app/features/metadata/utils.ts
```diff
@@ -1,23 +1,54 @@
...
+
+const resolveIpfsUri = (url: URL): string => {
+    // eslint-disable-next-line no-restricted-syntax -- Strips redundant "ipfs/" prefix from the path for a clean gateway URL.
+    const path = (url.host + url.pathname).replace(/^ipfs\//, '');
+    if (!verifyCID(path)) {
+        Logger.warn(`[metadata] Cannot fetch a malformed CID: ${path}`);
+        return '';
+    }
+    return `${IPFS_GATEWAY}/${path}${url.search}`;
+};
```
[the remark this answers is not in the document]
holps-7: solved here [2a5ec2d](https://github.com/solana-foundation/explorer/pull/853/commits/2a5ec2dd2f9bc23b28a144bcf92ec891d7c1beec)

holps-7:
> Finally, we’re ready for this PR. @holps-7, please rebase onto the current master - there are some conflicts. If you don’t mind, an OpenSpec proposal doc would be nice (but it’s not required).

Hey @askov I've rebased the branch onto the current master and resolved all the merge conflicts. I've also added an OpenSpec proposal doc for the IPFS URI resolution as you suggested.

package.json
```diff
@@ -138,6 +139,7 @@
         "@fission-ai/openspec": "1.3.1",
         "@mdx-js/react": "3.1.0",
         "@playwright/test": "1.57.0",
+        "@solana/eslint-config-solana": "1.0.1",
```
askov: **issue**: This seems to be unrelated to the PR. And the lock file has some extra changes too. Please remove them

app/features/metadata/utils.ts
```diff
@@ -1,10 +1,11 @@
-import { parseUrl, SAFE_EXTERNAL_PROTOCOLS } from '@/app/shared/lib/url';
```
askov, line 1: **todo**: Please, run prettier/eslint to fix CI errors

app/features/metadata/utils.ts
```diff
@@ -15,9 +16,44 @@ export const getProxiedUri = (uri: string): string | '' => {
...
+    const cid = firstSlash === -1 ? fullPath : fullPath.slice(0, firstSlash);
+    const subpath = firstSlash === -1 ? '' : fullPath.slice(firstSlash);
+    if (!verifyCID(cid)) {
+        Logger.warn(`[metadata] Cannot fetch a malformed CID: ${cid}`);
+        return '';
+    }
+    return `${IPFS_GATEWAY}/${cid}${subpath}${url.search}`;
+};
+
+const proxyUri = (uri: string): string => `/api/metadata/proxy?uri=${encodeURIComponent(uri)}`;
```
rogaldh, line 50: Explorer already has a helper to build a proxied url. I'd rather add another variant that supports an extended set of protocols. getProxiedUri

openspec/changes/resolve-ipfs-uris-to-http-gateway/proposal.md
```diff
@@ -0,0 +1,32 @@
+# Proposal: Resolve ipfs:// URIs to an HTTP gateway for NFT images and metadata
```
rogaldh, line 1: ❤️

package.json
holps-7: resolved

app/features/metadata/utils.ts
holps-7: resolved

Woody4618 (approved):

Woody4618:
LGTM: 
image

Thanks

PR #857, opened 2026-02-24
feat: save custom clusters

sporicle:
## Description

I add the feature to persist multiple custom RPC endpoints in browser storage.

## Type of change

-   [ ] Bug fix
-   [x] New feature
-   [ ] Protocol integration
-   [ ] Documentation update

## Screenshots

Screenshot 2026-02-24 at 3 25 27 PM
Screenshot 2026-02-24 at 3 24 57 PM
Screenshot 2026-02-24 at 3 25 03 PM
Screenshot 2026-02-24 at 3 25 09 PM

## Testing

- I created unit tests covering CRUD operations, localStorage unavailability, corrupted data recovery, duplicate name dedup, and findSavedClusterUrl lookup.
- I created component tests covering save flow, validation, display/delete
- tested the test suite - 0 failures in my file, but had some pre-existing fails in unrelated files

## Checklist
-   [ ] CI/CD checks pass

## Additional Notes
I go through my changes here: https://x.com/andyweng_/status/2026219580927201282


app/components/ClusterModal.tsx
```diff
@@ -85,8 +159,103 @@ function assertUnreachable(_x: never): never {
...
+                className="btn btn-sm position-absolute"
+                style={{ right: 4, top: '50%', transform: 'translateY(-50%)' }}
+                onClick={e => {
+                    e.stopPropagation();
+                    onDelete(cluster.name);
+                }}
+                data-testid={`delete-cluster-${cluster.name}`}
+                aria-label={`Delete ${cluster.name}`}
+            >
+                &times;
```
askov: **suggestion**: In my opinion, using a cross icon (the same size and color as the one for closing the modal) is a bit confusing. Don’t you think? Since the project already has react-feather icons, maybe something like a trash icon would work better?

app/components/ClusterModal.tsx
```diff
@@ -85,8 +159,103 @@ function assertUnreachable(_x: never): never {
...
+    const nextSearchParams = new URLSearchParams(searchParams?.toString());
+    nextSearchParams.set('cluster', 'custom');
+    nextSearchParams.set('customUrl', cluster.url);
+    const nextQueryString = nextSearchParams.toString();
+    const clusterUrl = `${pathname}?${nextQueryString}`;
+
+    return (
+        <div className="position-relative col-12 mb-3" data-testid={`saved-cluster-${cluster.name}`}>
+            <Link
+                className={`btn col-12 text-center ${btnClass}`}
```
askov: **suggestion**: Please use the cn utility for class manipulations

app/components/ClusterModal.tsx
```diff
@@ -85,8 +159,103 @@ function assertUnreachable(_x: never): never {
...
+}) {
+    const { customUrl, cluster } = useCluster();
+    const router = useRouter();
+    const pathname = usePathname();
+
+    const handleDelete = (name: string) => {
+        const isActive = cluster === Cluster.Custom && savedClusters.find(c => c.name === name)?.url === customUrl;
+        removeSavedCluster(name);
+        if (isActive) {
+            setPersistedCluster('mainnet-beta');
```
askov: **issue**: It's highly recommended to use DEFAULT_CLUSTER instead of hardcoding the value.

app/components/ClusterModal.tsx
```diff
@@ -85,8 +159,103 @@ function assertUnreachable(_x: never): never {
...
+            </Link>
+            <button
+                className="btn btn-sm position-absolute"
+                style={{ right: 4, top: '50%', transform: 'translateY(-50%)' }}
+                onClick={e => {
+                    e.stopPropagation();
+                    onDelete(cluster.name);
+                }}
+                data-testid={`delete-cluster-${cluster.name}`}
+                aria-label={`Delete ${cluster.name}`}
```
askov, line 235: **praise**: It's great to see that people care about accessibility!

app/utils/cluster.ts
```diff
@@ -101,4 +101,21 @@ export function serverClusterUrl(cluster: Cluster, customUrl: string): string {
     }
 }
 
+export function clusterFromSlug(slug: string): Cluster | null {
```
askov, line 104: **suggestion(non-blocking)**: clusterFromSlug is a hand-written inverse of clusterSlug. If a new cluster is added to clusterSlug but forgotten in clusterFromSlug, they silently desync. We can build it from clusterSlug:
```ts
return CLUSTERS.find(c => clusterSlug(c) === slug) ?? null;
```

app/utils/cluster-storage.ts
```diff
@@ -0,0 +1,57 @@
+import { localStorageIsAvailable } from './local-storage';
```
askov: **nitpick**: Hey! Thanks for your contribution, it's very nice to see this feature implemented. We try to put all the new features according to FSD so it's easier to maintain. In this case It would be great if you could move all the new code to the features/custom-cluster/{model/ui/lib}. This isn't mandatory, so it's up to you

markusknudtsen-droid (approved):

app/utils/cluster-storage.ts
sporicle: thanks for the review! went ahead and moved to fsd structure (features/custom-cluster/)

Woody4618:
Can you please resolve the merge conflicts, then we can merge this. 
@sporicle

app/features/custom-cluster/lib/cluster-storage.ts
```diff
@@ -0,0 +1,57 @@
+import { localStorageIsAvailable } from '@utils/local-storage';
```
rogaldh: I'd recommend using `jotai`, which is present in the Explorer. They have a layer to work with [Storages](https://jotai.org/docs/utilities/storage).

app/providers/cluster.tsx
```diff
@@ -99,7 +98,10 @@ export function ClusterProvider({ children }: ClusterProviderProps) {
         cluster === Cluster.Custom ||
         (localStorageIsAvailable() && localStorage.getItem('enableCustomUrl') !== null) ||
         isWhitelistedRpc(state.customUrl);
-    const customUrl = (enableCustomUrl && searchParams?.get('customUrl')) || state.customUrl;
+
+    const urlFromParams = enableCustomUrl ? searchParams?.get('customUrl') : null;
+    const urlFromSaved = cluster === Cluster.Custom && !urlFromParams ? findSavedClusterUrl(getPersistedCluster() ?? '') : undefined;
+    const customUrl = urlFromParams ?? urlFromSaved ?? state.customUrl;
```
rogaldh: I'm worried about this line. Imagine we have a clusterUrl parameter, but it is empty. If I am not mistaken, ?? would allow the first branch, and it will lead to an error because a url is gonna be created for an empty string.

I've also tried playing with the feature a bit, and it looks like it is possible to set the devnet to persisted, which would be stored in local storage and wouldn't be displayed at the URL. That is a breaking change.

Image

These changes definitely require some tests.

app/features/custom-cluster/lib/cluster-storage.ts
sporicle: fixed.

app/providers/cluster.tsx
sporicle: - added || null on line 104 to guard against empty strings before they reach the ??
- reverted to pre-pr behavior for persisted state + added a test for the hierachy

app/components/__tests__/ClusterModal.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+    });
+
+    it('shows name input after clicking Save this cluster', () => {
+        renderWithStore();
+        fireEvent.click(screen.getByTestId('save-custom-cluster-btn'));
+        expect(screen.getByTestId('cluster-name-input')).toBeInTheDocument();
+        expect(screen.getByTestId('name-required-hint')).toBeInTheDocument();
+    });
+
+    it('saves a custom cluster and persists the selection', () => {
```
rogaldh: Could you please add 'should' to each `it()` case? That is to align the current test suite with other tests present in the project.

app/components/ClusterModal.tsx
```diff
@@ -58,24 +76,81 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
         }
     }, 500);
 
-    const inputTextClass = editing ? '' : 'text-muted';
+    const handleSave = () => {
+        if (!savedName.trim()) return;
+        addSavedCluster({ name: savedName.trim(), url: customUrl });
+        setPersistedCluster(`${SAVED_CLUSTER_PREFIX}${savedName.trim()}`);
```
rogaldh: Theoretically, we could reach the limit for the storage, and this command would trigger an exception: https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#web_storage

I'd recommend handling this type of case.

app/components/ClusterModal.tsx
```diff
@@ -58,24 +76,81 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
...
+        setSavedName('');
+        setSaving(false);
+    };
+
     return (
         <>
             <Link
-                className={`btn col-12 mb-3 ${btnClass}`}
+                className={cn('btn col-12 mb-3', active ? `border-${activeSuffix} text-${activeSuffix}` : 'btn-white')}
```
rogaldh: Could you, please, use `class-variance-authority` to cover variants. `active` is a separate variant

app/components/ClusterModal.tsx
```diff
@@ -85,8 +160,99 @@ function assertUnreachable(_x: never): never {
...
+    const removeSavedCluster = useSetAtom(removeSavedClusterAtom);
+    const setPersistedCluster = useSetAtom(persistedClusterAtom);
+    const router = useRouter();
+    const pathname = usePathname();
+
+    const handleDelete = (name: string) => {
+        const isActive = cluster === Cluster.Custom && savedClusters.find(c => c.name === name)?.url === customUrl;
+        removeSavedCluster(name);
+        if (isActive) {
+            setPersistedCluster(clusterSlug(DEFAULT_CLUSTER));
```
rogaldh: I think this may cause problems. Deleting a cluster causes a rollback to the default one. I'd expect null here. WDYT?

app/providers/cluster.tsx
```diff
@@ -94,12 +95,20 @@ export function ClusterProvider({ children }: ClusterProviderProps) {
...
+    const [persistedCluster, setPersistedCluster] = useAtom(persistedClusterAtom);
+    const savedClusters = useAtomValue(savedClustersAtom);
+    const cluster = parseQuery(searchParams, persistedCluster);
     const enableCustomUrl =
         cluster === Cluster.Custom ||
         (localStorageIsAvailable() && localStorage.getItem('enableCustomUrl') !== null) ||
         isWhitelistedRpc(state.customUrl);
-    const customUrl = (enableCustomUrl && searchParams?.get('customUrl')) || state.customUrl;
+
+    const urlFromParams = enableCustomUrl ? searchParams?.get('customUrl') || null : null;
```
rogaldh: I'd suggest using undefined as the indicator for the absent value unless the library requires it.

app/components/ClusterModal.tsx
```diff
@@ -85,8 +160,99 @@ function assertUnreachable(_x: never): never {
...
+            </button>
+        </div>
+    );
+}
+
+function SavedClustersSection({
+    activeSuffix,
+    savedClusters,
+}: {
+    activeSuffix: string;
```
rogaldh: activeSuffix has a limited number of values so I'd suggest using literals instead of broad string

app/components/__tests__/ClusterModal.spec.tsx
sporicle: assuming you meant the wording - added

app/components/ClusterModal.tsx
sporicle: - wrapped the logic in a try/catch and added a test for storage full errors
- added a test for validating that this is caught correctly

app/components/ClusterModal.tsx
sporicle: done.

app/components/ClusterModal.tsx
sporicle: yeah you're right, null is cleaner. gets overwritten anyways. changed.

app/providers/cluster.tsx
sporicle: done.

app/components/ClusterModal.tsx
sporicle: solved as part of removing persisted cluster

app/components/ClusterModal.tsx
```diff
@@ -58,76 +92,220 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
...
-        case ClusterStatus.Connecting:
-            activeSuffix = 'warning';
-            break;
-        case ClusterStatus.Failure:
-            activeSuffix = 'danger';
-            break;
-        default:
-            assertUnreachable(status);
-    }
+    const savedClusters = useAtomValue(savedClustersAtom);
```
rogaldh, line 280: It looks like this line causes a hydration error. On the server, there is no localStorage, and after the mount, the component would have a cluster rendered from localStorage. Could you, please, check

sporicle: fixed

app/features/custom-cluster/lib/cluster-storage.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export interface SavedCluster {
+    name: string;
+    url: string;
+}
+
+export const savedClustersAtom = atomWithStorage<SavedCluster[]>('explorer:savedClusters', []);
+
+export const addSavedClusterAtom = atom(null, (get, set, cluster: SavedCluster) => {
+    set(savedClustersAtom, [...get(savedClustersAtom).filter(c => c.name !== cluster.name), cluster]);
```
rogaldh: It looks like we may have a problem here. If localstorage is not available and setItem throws an exception, Jotai's `set` would be called: https://github.com/pmndrs/jotai/blob/main/src/vanilla/utils/atomWithStorage.ts#L263-L264
Could you, please, check that the data won't be desynced?

app/providers/cluster.tsx
```diff
@@ -99,11 +90,11 @@ export function ClusterProvider({ children }: ClusterProviderProps) {
         cluster === Cluster.Custom ||
         (localStorageIsAvailable() && localStorage.getItem('enableCustomUrl') !== null) ||
         isWhitelistedRpc(state.customUrl);
-    const customUrl = (enableCustomUrl && searchParams?.get('customUrl')) || state.customUrl;
+
+    const customUrl = (enableCustomUrl ? searchParams?.get('customUrl') : undefined) ?? state.customUrl;
     const pathname = usePathname();
     const router = useRouter();
 
-    // Remove customUrl param if dev setting is disabled
```
rogaldh, line 106: Let's keep this comment unless it is irrelevant. Looks relevant to me.

app/components/ClusterModal.tsx
```diff
@@ -37,16 +69,23 @@ export function ClusterModal() {
...
-function CustomClusterInput({ activeSuffix, active }: InputProps) {
+type InputProps = { status: ClusterStatus; active: boolean; savedClusters: SavedCluster[] };
+function CustomClusterInput({ status, active, savedClusters }: InputProps) {
     const { customUrl } = useCluster();
     const updateCustomUrl = useUpdateCustomUrl();
+    const addSavedCluster = useSetAtom(addSavedClusterAtom);
     const [editing, setEditing] = React.useState(false);
+    const [saving, setSaving] = React.useState(false);
+    const [savedName, setSavedName] = React.useState('');
+    const [saveError, setSaveError] = React.useState<string | null>(null);
```
rogaldh: In the project, we are trying to use undefined instead of null. Could you, please, consider using undefined here?
If that is necessary, please leave a comment.

app/components/ClusterModal.tsx
```diff
@@ -58,76 +97,206 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
...
-    const inputTextClass = editing ? '' : 'text-muted';
+    const handleSave = () => {
+        if (!savedName.trim()) return;
+        setSaveError(null);
+        try {
+            addSavedCluster({ name: savedName.trim(), url: customUrl });
+            setSavedName('');
+            setSaving(false);
+        } catch {
+            setSaveError('Not enough storage space to save the cluster. Try removing unused clusters.');
```
rogaldh: I'd suggest preserving the original error. `Error`, according to the standard, has [`cause` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) to keep some original data.
Let's use Error here instead of a plain string

app/features/custom-cluster/lib/cluster-storage.ts
```diff
@@ -0,0 +1,20 @@
...
+});
+
+export const removeSavedClusterAtom = atom(null, (get, set, name: string) => {
+    set(
+        savedClustersAtom,
+        get(savedClustersAtom).filter(c => c.name !== name)
```
rogaldh: This filter is used two times here. I think it makes sense to move it into a helper to not break the filtering.

app/features/custom-cluster/lib/cluster-storage.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const savedClustersAtom = atomWithStorage<SavedCluster[]>('explorer:savedClusters', []);
+
+export const addSavedClusterAtom = atom(null, (get, set, cluster: SavedCluster) => {
+    set(savedClustersAtom, [...get(savedClustersAtom).filter(c => c.name !== cluster.name), cluster]);
+});
+
+export const removeSavedClusterAtom = atom(null, (get, set, name: string) => {
+    set(
```
rogaldh: The same problem might happen here. If localStorage allows reading but setItem is forbidden, we have a new value in-memory, but the value in storage is different.

app/components/ClusterModal.tsx
```diff
@@ -58,76 +97,206 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
...
+    const clusterUrl = `${pathname}?${nextQueryString}`;
+
+    return (
+        <div className="position-relative col-12 mb-3" data-testid={`saved-cluster-${cluster.name}`}>
+            <Link className={cn(clusterButtonVariants({ active: isActive, status }), 'text-center')} href={clusterUrl}>
+                {cluster.name}
+            </Link>
+            <button
+                className="btn btn-sm position-absolute"
+                style={{ right: 4, top: '50%', transform: 'translateY(-50%)' }}
```
rogaldh: I'd suggest using tailwind classes instead on inline styles

app/components/ClusterModal.tsx
```diff
@@ -58,76 +97,206 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
         }
     }, 500);
 
-    const inputTextClass = editing ? '' : 'text-muted';
+    const handleSave = () => {
+        if (!savedName.trim()) return;
```
rogaldh: Here, the savedValue is trimmed, but it seems reasonable to trim the value upon saving. That allows to only trim once instead of trying to trim the value when accessing it

rogaldh: Btw, is it possible to avoid using state to store the value we are trying to save in localStorage?

app/features/custom-cluster/lib/cluster-storage.ts
sporicle: Good catch - added pre-validation by writing to localStorage first. If the write fails, the atom doesn't get updated so they stay in sync.

app/providers/cluster.tsx
sporicle: Restored.

app/components/ClusterModal.tsx
sporicle: Fixed - switched to useState<Error | undefined>(undefined)

app/components/ClusterModal.tsx
sporicle: Now using new Error(message, { cause: error }) + rendering the error.

app/features/custom-cluster/lib/cluster-storage.ts
sporicle: Extracted into excludeByName helper, used by both atoms now.

app/features/custom-cluster/lib/cluster-storage.ts
sporicle: Same pre-validation approach as above by writing before updating the atom.

app/components/ClusterModal.tsx
sporicle: Switched to right-1 top-1/2 -translate-y-1.

app/components/ClusterModal.tsx
sporicle: Fixed - trimming once at the top of handleSave now. 

I think state is needed here since the value drives the disabled state on the save button and the "Name is required" hint. Without state the validation feedback wouldn't be live.

app/providers/cluster.tsx
```diff
@@ -99,11 +90,11 @@ export function ClusterProvider({ children }: ClusterProviderProps) {
         cluster === Cluster.Custom ||
         (localStorageIsAvailable() && localStorage.getItem('enableCustomUrl') !== null) ||
         isWhitelistedRpc(state.customUrl);
-    const customUrl = (enableCustomUrl && searchParams?.get('customUrl')) || state.customUrl;
+
+    const customUrl = (enableCustomUrl ? searchParams?.get('customUrl') : undefined) ?? state.customUrl;
```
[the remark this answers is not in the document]
rogaldh: I suggest adding a comment here to make it less fragile for future changes:
```
// `|| undefined` converts an empty string to undefined, so `??` falls back to state.customUrl
```

sporicle:
No, I wasn't familiar with the codebase so it's important to get it right - thanks for the review!

app/components/ClusterModal.tsx
```diff
@@ -58,76 +97,212 @@ function CustomClusterInput({ activeSuffix, active }: InputProps) {
...
+            addSavedCluster({ name, url: localUrl });
+            setSavedName('');
+            setSaving(false);
+        } catch (error) {
+            setSaveError(
+                new Error('Not enough storage space to save the cluster. Try removing unused clusters.', {
+                    cause: error,
+                })
+            );
+        }
```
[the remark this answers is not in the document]
rogaldh: Oh. That is fair. @sporicle could you check it please? 🙏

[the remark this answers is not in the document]
sporicle: added check

markusknudtsen-droid (commented):
Good work

PR #871, opened 2026-03-06
refactor: Centralize logging and Sentry error reporting

askov:
## Description

 - **Centralized Logger**: Replace `app/utils/logger` (static class, default export) with `app/shared/lib/logger` (singleton instance, named export)
  - **Sentry consolidation**: All direct `@sentry/nextjs` imports replaced by `Logger.panic()`/`Logger.warn({ sentry: true })` or re-exports from `app/shared/lib/sentry/`. ESLint `no-restricted-imports` rule enforces this boundary
  - **Global error boundaries**: New `app/error.tsx` and `app/global-error.tsx` catch route-level and root-level errors, replacing per-component ErrorBoundary wrappers
  - **ESLint enforcement**: `no-console: error` globally to prevent regression
  - **Sentry SDK upgrade**: Pin `@sentry/core` and `@sentry/nextjs` to `10.39.0`, migrate build config to `webpack.treeshake.removeDebugLogging`
  - **Test cleanup**: Global Logger mock in `test-setup.specs.ts` replaces ~8 per-file mock blocks

  ### Breaking changes

  - `NEXT_PUBLIC_ENABLE_CATCH_EXCEPTIONS` env var removed — Sentry capture is now always-on when DSN is configured
  - `SENTRY_SUPPRESS_GLOBAL_ERROR_HANDLER_FILE_WARNING` env var removed
  - Import paths changed: `@utils/logger` → `@/app/shared/lib/logger`, `SentryErrorBoundary` and `withTraceData` moved to `@/app/shared/lib/sentry`
## Type of change

-   [x] Other (please describe): Enable sentry for all pages, logger refactoring

## Related Issues

Closes [HOO-246](https://linear.app/solana-fndn/issue/HOO-246/enable-sentry-for-all-the-pages)
## Checklist

-   [x] CI/CD checks pass

### DIFF
--- a/app/api/programMetadataIdl/__tests__/route.spec.ts
+++ b/app/api/programMetadataIdl/__tests__/route.spec.ts
@@ -114,7 +114,9 @@ describe('GET api/programMetadataIdl', () => {
         const request = createRequest(mockAddress, Cluster.MainnetBeta, 'idl');
         const res = await GET(request);
         expect(res.status).toBe(200);
-        expect(Logger.error).toHaveBeenCalledWith(expectedError.cause);
+        expect(Logger.error).toHaveBeenCalledWith(
+            new Error('[api:programMetadataIdl] Request failed', { cause: expectedError.cause })
+        );
         expect(await res.json()).toEqual({ details: { cause: {} }, error: expectedError.message });
     });
 
--- a/app/api/verified-programs/list/[page]/__tests__/route.spec.ts
+++ b/app/api/verified-programs/list/[page]/__tests__/route.spec.ts
@@ -201,7 +201,7 @@ describe('GET /api/verified-programs/list/[page]', () => {
             await GET(mockRequest, params);
 
             expect(Logger.error).toHaveBeenCalledWith(
-                '[api:verified-programs] Failed to fetch verified programs page',
+                new Error('[api:verified-programs] Failed to fetch verified programs page'),
                 { page: 1, status: 503 }
             );
         });


app/shared/lib/logger.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @param expectedLevel - The log level of the message to be logged
+ * @returns True if the message should be logged, false otherwise. Returns false if NEXT_LOG_LEVEL is not set or invalid.
+ *
+ * @example
+ * // With NEXT_LOG_LEVEL=3 (INFO level)
+ * isLoggable(LOG_LEVEL.ERROR)  // returns true (1 <= 3)
+ * isLoggable(LOG_LEVEL.DEBUG)  // returns false (4 > 3)
+ */
+function isLoggable(expectedLevel: LOG_LEVEL) {
+    const currentLevel = process.env.NEXT_LOG_LEVEL ? parseInt(process.env.NEXT_LOG_LEVEL) : undefined;
```
rogaldh, line 97: nit: let's leave a comment somewhere that without `NEXT_LOG_LEVEL` being set, all the logging would be hidden

app/entities/idl/model/use-format-anchor-idl.ts
```diff
@@ -297,9 +299,10 @@ export function useFormatAnchorIdl(idl?: Idl): FormattedIdl | null {
                 })),
         };
 
-        // Log formatting errors with Sentry upon its delivery. Just show them at this point
         if (formattingErrors.length) {
-            console.error('Formatting Errors:', formattingErrors);
+            formattingErrors.forEach(([err, acc]) =>
+                Logger.panic('[idl] Formatting error', { error: err, hint: { extra: { acc } } })
```
rogaldh: I think here, specifically, we may need `warn` with the sentry flag instead of a panic. These errors kinda normal, and some of them may be invoked by incorrectly prepared IDL

app/shared/lib/logger.ts
```diff
@@ -0,0 +1,106 @@
...
+        // Libraries and boundary code (try/catch of unknown, promise rejections, fetch
+        // failures) can surface non-Error values — strings, numbers, null, or plain
+        // objects. Sentry groups and tracks real Error instances far better (stack traces,
+        // deduplication), so we wrap non-Error values to preserve that behavior.
+        const error = context.error instanceof Error ? context.error : new Error(String(context.error));
+        captureException(error, context.hint);
+        isLoggable(LOG_LEVEL.PANIC) && console.error(...formatArgs(message, context));
+    }
+    error(message: string, context?: Record<string, unknown>) {
+        isLoggable(LOG_LEVEL.ERROR) && console.error(...formatArgs(message, context));
```
rogaldh: Let's preserve the original logic about using the source error and fall back to a generic error, like here: https://github.com/solana-foundation/explorer/blob/master/app/utils/logger.ts#L44-L48

The intent was to save the original error without modifications. Having "Unrecognised error" means that something is wrong and requires attention.

rogaldh: That also means that we have to wrap strings passed to Logger.error with error instances

app/shared/lib/logger.ts
askov: Updated the fn docs and env.example with a hint

app/entities/idl/model/use-format-anchor-idl.ts
askov: Done

app/shared/lib/logger.ts
askov: I'm not sure if I fully got your idea. I updated it to log 'unexpected' error types and to only send error instances to sentry. But converting an unexpected unknown to an error and console.error it with a default message seems unnecessary, doesn't it? You already have a debug log associated with it.

rogaldh: I mean that `Logger.error` should allow passing an' Error' instance as its first argument. That seems reasonable.

askov: Oh, now I see what you mean.

Could you elaborate on this?

I don’t see any strong reasons for allowing the Error argument to come first.
Well, the only advantage I can see is that it makes catch blocks more ergonomic. 
```ts
Logger.error(err)
// is shorter than
Logger.error(‘[idl:form] failed to load deps’, { error: err })
```
However, the cons are:
- It breaks the uniform signature, making the Error the odd one out
- The quality of sentry messages degrades. If someone passes an Error directly and captureMessage is triggered, the sentry issue title becomes err.message, which is often generic (like "Network request failed" or “Bigint conversion failed”) instead of a stable, searchable string like [tx:fetch] failed to load transaction.
- Overloads are typically used for functions that can be called in fundamentally different ways for a reason. But why apply this to logs? I mean, it only increases developer friction and leads to unclear logs, such as Logger.error(error, { sentry: true }), where the error is often some third-party issue that the developer may not understand well.

The uniform (message, context) signature is a deliberate trade-off that prioritizes stable, searchable log messages and clean Sentry grouping over convenience at the call site. The "boilerplate" of `Logger.error('[module] descriptive message', { error: err })` is actually a feature—it compels callers to describe what failed, rather than simply forwarding an opaque error.

rogaldh: I see your point.

But I think not using `Errors` and allowing `strings` or `objects` to describe errors discourages their use in general, which is not good and prevents building structures that handle various errors separately.

> `captureMessage is triggered`

If we write code and specifically pass `e.message` into the capture, we did this intentionally; I do not mean we have to pass it as is to make it shorter, but we can definitely do that if needed, for ex.: `.catch(Logger.error)`.

What stops us from wrapping the original error with a fine-grained error?

```typescript
Logger.error(new NamespacedError('[idl:form] failed to load deps'), { cause: err })
// OR
Logger.error(new IdlFormError('Failed to load deps', { cause: err }))
```
I think the last line is almost identical to `Logger.error(‘[idl:form] failed to load deps’, { error: err })`, but the variant with an Error instance has more benefits in the long run than the variant with a string.

Speaking of the uniform signature. That is definitely a good thing, but we do not use captureMessage on par with `Logger.error`; it is inside the implementation, so we can extract the message from the error. Having a proper error-handling structure outweighs a uniform signature. I do not think we are aiming to allow using `Logger.error` and `Sentry.captureMessage` interchangeably.

Using overloads was just a proposal to describe the possibility of using `Errors` and all other types that fall into the `unknown` category. We can do unions or anything else.

To summarize:
- I do not mean we have to have the shortest form to log the error
- Error is designed to handle errors, and having fine-grained/namespaced errors is possible with it

askov: Okay, I agree with you that it's flexible and logical. I've updated the code.

However, the concern about this relaxed usage still remains. When you have an 'unknown' error parameter, it **always** (I'm intentionally exaggerating here, but you get what I mean - this is an open-source project 😄) leads to people simply using Logger.error(err) without actually creating any specialized error instances or causes.

app/api/metadata/proxy/route.ts
```diff
@@ -44,13 +44,13 @@ export async function GET(request: Request, { params: _params }: Params) {
...
         if (!isHTTPProtocol(parsedUrl)) {
-            Logger.error(new Error('Unsupported protocol'), parsedUrl.protocol);
+            Logger.error('[api:metadata-proxy] Unsupported protocol', { protocol: parsedUrl.protocol });
             return respondWithError(400);
         }
 
         const isPrivate = await checkURLForPrivateIP(parsedUrl);
         if (isPrivate) {
-            Logger.error(new Error('Private IP detected'), parsedUrl.hostname);
+            Logger.error('[api:metadata-proxy] Private IP detected', { hostname: parsedUrl.hostname });
```
C0mberry: the original error message will be lost here
resolveError() will convert it to new Error('Unrecognized error')

app/api/verification/coingecko/[coinId]/route.ts
```diff
@@ -45,9 +44,9 @@ export async function GET(_request: Request, { params: { coinId } }: Params) {
 
         if (!response.ok) {
             if (response.status === 429) {
-                Sentry.captureMessage('Coingecko API rate limit exceeded', { level: 'warning' });
+                Logger.error('[api:coingecko] Rate limit exceeded', { sentry: true });
```
Comment (C0mberry) [inline app/api/verification/coingecko/[coinId]/route.ts]: we could also lose the error message 
maybe Logger.warn would be fit for this place

app/error.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Logger } from '@/app/shared/lib/logger';
+
+export default function Error({ error, reset }: { error: Error & { digest?: string }; reset: () => void }) {
+    useEffect(() => {
+        Logger.panic(error);
+    }, [error]);
+
+    return (
+        <div className="container mt-4">
+            <ErrorCard text={error.message} retry={reset} />
```
C0mberry: raw error message could be displayed to the user

app/error.tsx
```diff
@@ -0,0 +1,18 @@
+'use client';
```
C0mberry, line 1: do we need tests for it or app/global-error.tsx?

askov: Yes, here're some [docs](https://nextjs.org/docs/app/getting-started/error-handling#global-errors)

app/error.tsx
askov: Updated to the generic message

app/api/verification/coingecko/[coinId]/route.ts
Comment (askov) [reply app/api/verification/coingecko/[coinId]/route.ts]: Replaced with warn (all the 429 in routes)

app/api/metadata/proxy/route.ts
askov: Yeh, I missed a few of these during the latest refactoring commit (changed the signature to less strict). Fixed all the cases like this

app/shared/lib/logger.ts
rogaldh: We could write an ESLint rule that would warn from using it that way. LGTM

app/shared/lib/logger.ts
rogaldh: I realised that on the client, `NEXT_LOG_LEVEL` would not be accessible, only on the server. So we have to support `NEXT_PUBLIC_LOG_LEVEL`.
Also, one should not overwrite another. I mean the case when NEXT_PUBLIC_ is available, but NEXT_ is not.
Smth like:
```typescript
const raw = typeof window === 'undefined'                                                                                                                                                                 
  ? process.env.NEXT_LOG_LEVEL                                                                                                                                                                          
  : process.env.NEXT_PUBLIC_LOG_LEVEL;
const currentLevel = raw ? parseInt(raw) : undefined;
```

app/api/verification/coingecko/[coinId]/route.ts
```diff
@@ -58,8 +57,7 @@ export async function GET(_request: Request, { params: { coinId } }: Params) {
         const data = await response.json();
         return NextResponse.json(data, { headers: CACHE_HEADERS });
     } catch (error) {
-        Logger.error(new Error('Coingecko API error', { cause: error }));
-        Sentry.captureException(error);
+        Logger.panic(error instanceof Error ? error : new Error('Failed to fetch coingecko data'));
```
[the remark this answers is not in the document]
Comment (askov) [reply app/api/verification/coingecko/[coinId]/route.ts:60]: We decided to go with panic for now

app/api/verification/coingecko/[coinId]/route.ts
```diff
@@ -45,9 +44,9 @@ export async function GET(_request: Request, { params: { coinId } }: Params) {
 
         if (!response.ok) {
             if (response.status === 429) {
-                Sentry.captureMessage('Coingecko API rate limit exceeded', { level: 'warning' });
+                Logger.warn('[api:coingecko] Rate limit exceeded', { sentry: true });
             } else {
-                Sentry.captureException(new Error(`Coingecko API error: ${response.status}`));
+                Logger.panic(new Error(`Coingecko API error: ${response.status}`));
```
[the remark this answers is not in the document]
Comment (askov) [reply app/api/verification/coingecko/[coinId]/route.ts:49]: We decided to go with panic for now

PR #874, opened 2026-03-10
feat: Share receipt to social + native sharing

C0mberry:
## Description
- adding share to x btn
- adding native sharing
- fixing share btn height

## Type of change
-   [x] Bug fix
-   [x] New feature

## Screenshots
| Desktop | Mobile |
|--|--|
| Screenshot 2026-03-05 at 14 46 01 | IMG_8760 |

## Testing
1. Open http://localhost:3000/tx/4izwTCUeRGAMGReXeXDumiBzAgXPGz6KzccacCf1WU5YXCCpSDjAQT7J6D6dY45bL1NW9AiqwCuWEnz3hbGtZS2y?view=receipt on desktop
2. See share popup

1. Open the same link on mobile
2. See native sharing menu

## Related Issues
[HOO-320](https://linear.app/solana-fndn/issue/HOO-320/investigate-share-options-that-native-to-the-current-os)
[HOO-319](https://linear.app/solana-fndn/issue/HOO-319/add-share-to-x-sharable-variant)

## Checklist
-   [x] CI/CD checks pass

### DIFF
--- a/app/features/receipt/ui/ReceiptView.tsx
+++ b/app/features/receipt/ui/ReceiptView.tsx
@@ -43,7 +43,10 @@ export function ReceiptView({ data, signature, transactionPath }: ReceiptViewPro
             await navigator.share(shareData);
             receiptAnalytics.trackShareNative(signature);
         } catch (e) {
-            if (e instanceof Error && e.name === 'AbortError') return;
+            if (e instanceof Error && e.name === 'AbortError') {
+                // Dismissing the native share sheet is an expected cancel path, not a share failure.
+                return;
+            }
             toast.custom({ title: 'Failed to share', type: 'error' });
         }
     }


app/features/receipt/ui/ReceiptView.tsx
```diff
@@ -19,10 +22,35 @@ interface ReceiptViewProps {
 }
 
 export function ReceiptView({ data, signature, transactionPath }: ReceiptViewProps) {
+    const [canNativeShare, setCanNativeShare] = useState(false);
+
+    useEffect(() => {
+        const hasCoarsePointer = window.matchMedia('(pointer: coarse)').matches;
+        const hasNoHover = window.matchMedia('(hover: none)').matches;
```
rogaldh: WDYT about moving this code into a separate hook? That should allow using it at other palces

askov: **suggestion**: I agree that this should at least be extracted to a local hook, like useCanNativeShare, because having raw implementation details in a UI component is inappropriate. Additionally, the heuristic itself is undocumented. It would be better to include either comments or meaningful helper function names, such as isMobileDevice

app/features/receipt/ui/ReceiptView.tsx
```diff
@@ -19,10 +22,35 @@ interface ReceiptViewProps {
...
+                title: 'Solana Transaction Receipt',
+                url: globalThis.location.href,
+            };
+
+            if (!navigator.canShare?.(shareData)) return;
+            await navigator.share(shareData);
+            receiptAnalytics.trackShareNative(signature);
+        } catch (e) {
+            if (e instanceof Error && e.name === 'AbortError') return;
+            Sentry.captureException(e);
```
askov: **thought**: I don't think this kind of error deserves sentry logging, but the decision is yours. Wdyt @rogaldh?

app/og/receipt/[signature]/__tests__/route.spec.ts
```diff
@@ -1,6 +1,11 @@
 import { NextRequest } from 'next/server';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+vi.mock('@sentry/nextjs', () => ({
```
Comment (askov) [inline app/og/receipt/[signature]/__tests__/route.spec.ts]: **thougth**: This isn't really an issue, but it might not be clear for everyone. According to FSD, everything is correct; however, barrel imports can leak unwanted elements in the development environment. Perhaps, in addition to the index, we could have something like `server.ts` to support separate feature API surfaces. Then, the import for server routes could look like this: `import { X } from "@features/receipt/server"`

app/features/receipt/ui/icons/XIcon.tsx
```diff
@@ -0,0 +1,10 @@
+export function XIcon() {
```
askov: **nit**: Typically, such icons should have aria-hidden attributes since they're purely decorative. However, I'm not sure if the other icons in our project follow this recommendation.

app/og/receipt/[signature]/__tests__/route.spec.ts
Comment (askov) [reply app/og/receipt/[signature]/__tests__/route.spec.ts]: Wdyt @rogaldh ?

app/features/receipt/ui/ReceiptView.tsx
rogaldh: Agreed. Too harsh for the inability to share. To inform the user would be enough. We have a `toast` for that

app/features/receipt/ui/icons/XIcon.tsx
rogaldh: That is fair, but AFAIK, event react-feathers icons do not have this attribute. Pure SVGs, same as this. So this should be applied to each icon specifically.
I do like the suggestions. Let's maybe implement a wrapper or similar solution to apply this attribute to make icons a11y-friendly. @C0mberry @askov thoughts?

app/og/receipt/[signature]/__tests__/route.spec.ts
Comment (rogaldh) [reply app/og/receipt/[signature]/__tests__/route.spec.ts]: I also agree. Having everything exported to a single place mixes client and server code. That's why we have to mock something that is not even used.
Let's split the backend endpoint code into a separate file, `server.ts`. 

I do not like barrel files very much, but according to the FSD, it seems reasonable and also allows us to expose a "public" API for the feature.

TY @askov for noticing this.

app/features/receipt/ui/ReceiptView.tsx
C0mberry: done

app/og/receipt/[signature]/__tests__/route.spec.ts
Comment (C0mberry) [reply app/og/receipt/[signature]/__tests__/route.spec.ts]: done

app/features/receipt/ui/ReceiptView.tsx
C0mberry: done

app/features/receipt/ui/icons/XIcon.tsx
C0mberry: done

app/og/receipt/[signature]/__tests__/route.spec.ts
Comment (askov) [reply app/og/receipt/[signature]/__tests__/route.spec.ts]: Now we can remove this sentry mock, it does nothing

Comment (C0mberry) [reply app/og/receipt/[signature]/__tests__/route.spec.ts]: done

app/features/receipt/ui/ReceiptView.tsx
```diff
@@ -19,10 +21,28 @@ interface ReceiptViewProps {
...
+                title: 'Solana Transaction Receipt',
+                url: globalThis.location.href,
+            };
+
+            if (!navigator.canShare?.(shareData)) return;
+            await navigator.share(shareData);
+            receiptAnalytics.trackShareNative(signature);
+        } catch (e) {
+            if (e instanceof Error && e.name === 'AbortError') return;
+            throw e;
```
rogaldh: If I'm not mistaken, this leads to an unhandled exception. Let's maybe show a toast here

app/features/receipt/ui/ShareOnXShareItem.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface ShareOnXShareItemProps {
+    onShare?: () => void;
+}
+
+export function ShareOnXShareItem({ onShare }: ShareOnXShareItemProps) {
+    function handleClick() {
+        const url = encodeURIComponent(globalThis.location.href);
+        const win = globalThis.open(`https://x.com/intent/tweet?url=${url}`, '_blank', 'noreferrer');
+        if (win) onShare?.();
```
rogaldh: In case the win was not opened due to some restrictions, not handling the `!win` negative case leads to the limb state. User clicks the share button, but nothing happens. Let's cover that as well

app/shared/lib/use-can-native-share.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { useEffect, useState } from 'react';
+
+export function useCanNativeShare() {
+    const [canNativeShare, setCanNativeShare] = useState(false);
+
+    useEffect(() => {
+        // Heuristic for detecting a mobile device: coarse pointer (touch) and no hover support
+        const isMobileDevice =
+            window.matchMedia('(pointer: coarse)').matches && window.matchMedia('(hover: none)').matches;
```
rogaldh: We have `window` here, but `globalThis` elsewhere. Let's unify that if possible

app/features/receipt/ui/ShareOnXShareItem.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { PopoverMenuItem } from './PopoverMenuItem';
+
+interface ShareOnXShareItemProps {
+    onShare?: () => void;
+}
+
+export function ShareOnXShareItem({ onShare }: ShareOnXShareItemProps) {
+    function handleClick() {
+        const url = encodeURIComponent(globalThis.location.href);
+        const win = globalThis.open(`https://x.com/intent/tweet?url=${url}`, '_blank', 'noreferrer');
```
rogaldh: Definitely need here a comment on why we use only a `noreferrer`

app/features/receipt/ui/icons/XIcon.tsx
```diff
@@ -0,0 +1,17 @@
...
+    return (
+        <svg
+            xmlns="http://www.w3.org/2000/svg"
+            width="11"
+            height="11"
+            fill="none"
+            viewBox="0 0 11 11"
+            aria-hidden="true"
```
rogaldh: Why add `aria-hidden` here? Let's adjust with other icons.

app/features/receipt/ui/ShareOnXShareItem.tsx
```diff
@@ -0,0 +1,18 @@
+'use client';
```
rogaldh, line 1: I'd recommend adding tests for the component

app/shared/lib/use-can-native-share.ts
```diff
@@ -0,0 +1,19 @@
+'use client';
```
rogaldh, line 1: To have tests for the hook would be nice

app/features/receipt/server.ts
```diff
@@ -0,0 +1,5 @@
+export { ReceiptError } from './api/errors';
+export { isReceiptEnabled } from './env';
+export { createReceipt } from './model/create-receipt';
+export { buildCompositeSignature, parseCompositeSignature } from './model/composite-signature';
+export { IMAGE_SIZE as OG_IMAGE_SIZE, BaseReceiptImage } from './ui/BaseReceiptImage';
```
rogaldh, line 5: Looks like we may remove some from `index.ts`, right?

C0mberry: done

app/shared/lib/use-can-native-share.ts
C0mberry: done

app/features/receipt/ui/ShareOnXShareItem.tsx
C0mberry: done

app/features/receipt/ui/icons/XIcon.tsx
C0mberry: done

app/features/receipt/ui/ShareOnXShareItem.tsx
C0mberry: done

app/shared/lib/use-can-native-share.ts
C0mberry: done

app/features/receipt/ui/ShareOnXShareItem.tsx
C0mberry: done

app/features/receipt/ui/ReceiptView.tsx
C0mberry: done

app/features/receipt/ui/ReceiptView.tsx
```diff
@@ -19,10 +22,29 @@ interface ReceiptViewProps {
...
     }
 
+    async function handleNativeShare() {
+        try {
+            const shareData = {
+                title: 'Solana Transaction Receipt',
+                url: globalThis.location.href,
+            };
+
+            if (!navigator.canShare?.(shareData)) return;
```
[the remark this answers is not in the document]
C0mberry: we are returning toast as feedback

app/features/receipt/ui/__tests__/ShareOnXShareItem.spec.tsx
```diff
@@ -0,0 +1,64 @@
...
+        await userEvent.click(screen.getByRole('button', { name: /share on x/i }));
+
+        expect(openSpy).toHaveBeenCalledWith(
+            expect.stringContaining('https://x.com/intent/tweet?url='),
+            '_blank',
+            'noreferrer'
+        );
+
+        openSpy.mockRestore();
+    });
```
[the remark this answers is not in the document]
C0mberry: ShareOnXShareItem no longer uses useToast since the !win toast was dropped. There is nothing to clear in beforeEach

app/shared/lib/use-can-native-share.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    useEffect(() => {
+        // Heuristic for detecting a mobile device: coarse pointer (touch) and no hover support
+        const isMobileDevice =
+            globalThis.matchMedia('(pointer: coarse)').matches && globalThis.matchMedia('(hover: none)').matches;
+
+        setCanNativeShare(
+            typeof navigator.share === 'function' && typeof navigator.canShare === 'function' && isMobileDevice
+        );
+    }, []);
```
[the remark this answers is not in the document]
C0mberry: this is a very low-probability event

app/features/receipt/server.ts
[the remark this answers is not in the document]
C0mberry: BaseReceiptImage depends on next/og, which would already cause a Next.js build error if accidentally imported into a client bundle

app/features/receipt/ui/ReceiptView.tsx
```diff
@@ -19,10 +22,32 @@ interface ReceiptViewProps {
...
+            };
+
+            if (!navigator.canShare?.(shareData)) {
+                toast.custom({ title: 'Sharing not supported for this content', type: 'error' });
+                return;
+            }
+            await navigator.share(shareData);
+            receiptAnalytics.trackShareNative(signature);
+        } catch (e) {
+            if (e instanceof Error && e.name === 'AbortError') return;
```
rogaldh: Let's add a comment here about the purpose of returning. I assume we do not do a toast to not show the banner if the user rejected the native share, right?

C0mberry: done

PR #879, opened 2026-03-16
Fix out of data feature gate schedule 

Woody4618:
## Description
Feature gate schedule was out of date 6 months. 
In the action i triggered i saw ratelimits. So i added some delays and ran it locally. 
Also in the PR docs it said the github token parameter should be token not branch-token so i changed that. 
These should run every night. 
Lets monitor if it succeeds tomorrow. But we will have the up to data data right now

### DIFF
--- a/scripts/parse_feature_gates.py
+++ b/scripts/parse_feature_gates.py
@@ -211,15 +211,18 @@ async def fetch_cluster_activations(cluster_url: str, features_to_check: list[tu
 
     async with AsyncClient(cluster_url) as connection:
         epoch_schedule = (await connection.get_epoch_schedule()).value
+        if epoch_schedule is None:
+            print(f"[{cluster_name}] Failed to fetch epoch schedule, skipping cluster.")
+            return
 
         for existing, new_feature in features_to_check:
             if 'devnet' in cluster_url:
                 existing.devnet_activation_epoch = await fetch_activation_epoch(
-                    connection, epoch_schedule, existing.key, new_feature.devnet_activation_epoch
+                    connection, epoch_schedule, existing.key, existing.devnet_activation_epoch
                 )
             elif 'testnet' in cluster_url:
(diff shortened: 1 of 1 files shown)


scripts/parse_feature_gates.py
```diff
@@ -160,12 +161,24 @@ def safe_model_validate(model, data):
...
+        except Exception as e:
+            if '429' in str(e) and attempt < MAX_RETRIES - 1:
+                wait = 2 ** (attempt + 1)
+                print(f"Rate limited on {key}, retrying in {wait}s...")
+                await asyncio.sleep(wait)
+            else:
+                print(f"Failed to fetch {key}: {e}")
+                return backup_epoch
 
     if account.value and account.value.data:
```
rogaldh, line 189: Maybe it makes sense to add a guard `if account is None` here as well

scripts/fetch_mainnet_activations.py
rogaldh: I think it makes sense to move the RPC address to the const to make it easier to change it for https://github.com/solana-foundation/explorer/pull/844

scripts/parse_feature_gates.py
```diff
@@ -176,15 +189,34 @@ async def fetch_feature_activations(cluster_url: str, key: str, backup_epoch: in
...
-            return activation_epoch
+            return get_epoch_for_slot(epoch_schedule, activation_slot) + 1
         else:
             return backup_epoch
     else:
         return backup_epoch
 
 
+async def fetch_cluster_activations(cluster_url: str, features_to_check: list[tuple[StoredFeature, Feature]]) -> None:
+    """Fetch activation epochs for all features from a single cluster, reusing one connection."""
```
rogaldh, line 206: I'm not an expert in Python, but would the connection be closed after the fn finishes its work? I mean we create a connection with AsyncClient to the endpoint, and at the fn's end we have to close it, right?

PR #937, opened 2026-04-09
Try multiple signer seeds when resolving the Otter Verify build-params PDA

meumar-osec:
## Description

We were only trying one pubkey to derive the Otter Verify PDA. That works only when it matches whoever actually uploaded the build on-chain. In practice, Signer can be a placeholder (111…), or the PDA might have been created under a trusted uploader key that isn’t the program’s current upgrade authority, so we’d miss the account even when it exists.

This change passes the program upgrade authority into enrichment and tries a small list: authority first, then the registry signer if it’s valid, then the known trusted keys. We walk that list, derive the PDA for each, and use the first buildParams fetch that succeeds, closer to how solana-verify scans for a program’s PDAs.

Should fix Verified Build showing “not yet uploaded” when on-chain data is actually there.

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Protocol integration
-   [ ] Documentation update

### DIFF
--- a/app/utils/verified-builds.tsx
+++ b/app/utils/verified-builds.tsx
@@ -42,11 +42,20 @@ export type OsecInfo = {
     is_frozen: boolean;
 };
 
+function parsePublicKey(value: string | undefined): PublicKey | null {
+    if (!value) return null;
+    try {
+        return new PublicKey(value);
+    } catch {
+        return null;
+    }
+}
+
 const TRUSTED_SIGNERS: Record<string, string> = {
     '11111111111111111111111111111111': 'Explorer',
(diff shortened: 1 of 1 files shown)


Woody4618:
Hey, thank you for this! 
Greptile has some comments. Please have a look. 
I think the osec api does even one more check where it looks into the last transactions on the program data account to find the authority for frozen programs. You think this would also make sense to add?

meumar-osec:
Good call, agree this would help with frozen programs where the uploader authority differs.
If no PDA is found from authority/registry/trusted signers, we could scan recent programData tx history for past authorities and retry PDA derivation with those keys.
Let's keep the current fix for now and add this as a follow-up, unless you prefer to include it in this PR.

Woody4618 (approved):

PR #950, opened 2026-04-16
feat: Changing pre-filled value for interact tab

C0mberry:
## Description
- updating wallet id value for fields in interact tab 

## Type of change
-   [x] New feature

## Screenshots
https://github.com/user-attachments/assets/878bc5a9-5689-4f5c-8a62-ac8959d07e7e

## Testing
1. open [program](https://explorer-git-fork-hoodieshq-feat-prefi-0580d7-solana-foundation.vercel.app/address/whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc/idl) 
2. go to interact tab
3. connect 2 wallet account
4. change them and see how values are changing

## Related Issues
[HOO-263](https://linear.app/solana-fndn/issue/HOO-263/update-interact-idl-prefilled-wallet-when-connection-changes)

## Checklist
-   [x] CI/CD checks pass

### DIFF
--- a/app/features/idl/interactive-idl/model/form-prefill/providers/__tests__/wallet-prefill-provider.spec.ts
+++ b/app/features/idl/interactive-idl/model/form-prefill/providers/__tests__/wallet-prefill-provider.spec.ts
@@ -65,8 +65,6 @@ const INSTRUCTION_WITH_TWO_SIGNERS: InstructionData = {
     name: 'testInstruction',
 };
 
-const makeRef = (val?: string): { current: string | undefined } => ({ current: val });
-
 describe('createWalletPrefillDependency', () => {
     it('should fill signer accounts with wallet address', () => {
         const { result } = renderHook(() =>
@@ -78,13 +76,9 @@ describe('createWalletPrefillDependency', () => {
         const { form, fieldNames } = result.current;
 
         const walletPublicKey = PublicKey.default;
-        const dependency = createWalletPrefillDependency(
-            INSTRUCTION_WITH_SIGNER_AND_NON_SIGNER,
-            {
-                account: fieldNames.account,
-            },
-            makeRef(),
-        );
+        const dependency = createWalletPrefillDependency(INSTRUCTION_WITH_SIGNER_AND_NON_SIGNER, null, {
+            account: fieldNames.account,
+        });
 
         const walletAddress = walletPublicKey.toBase58();
         dependency.onValueChange(walletPublicKey, form);
@@ -102,13 +96,9 @@ describe('createWalletPrefillDependency', () => {
         );
         const { form, fieldNames } = result.current;
 
-        const dependency = createWalletPrefillDependency(
-            INSTRUCTION_WITH_SIGNER,
-            {
-                account: fieldNames.account,
-            },
-            makeRef(),
-        );
+        const dependency = createWalletPrefillDependency(INSTRUCTION_WITH_SIGNER, null, {
+            account: fieldNames.account,
+        });
 
         const setValueSpy = vi.spyOn(form, 'setValue');
         dependency.onValueChange(null, form);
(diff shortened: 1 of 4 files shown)


app/features/idl/interactive-idl/ui/InteractInstruction.tsx
```diff
@@ -50,12 +51,18 @@ export function InteractInstruction({
...
+        createWalletPrefillDependency(instruction, fieldNames, lastPrefillAddressRef).onValueChange(publicKey, form);
+        // instruction and fieldNames are stable for the lifetime of this component instance
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [publicKey, form]);
+
     const knownAccountsPrefillDependency = createKnownAccountsPrefillDependency(instruction, fieldNames);
     const pdaPrefillDependency = createPdaPrefillDependency(idl, instruction, fieldNames);
     useFormPrefill({
         config: {
-            externalDependencies: [walletPrefillDependency, knownAccountsPrefillDependency, pdaPrefillDependency],
```
askov, line 58: **issue**: There are some concerns with architecture and a simpler fix available.

The current approach lifts wallet prefill out of the `useFormPrefill` framework into a an ad-hoc `useEffect` + `useRef` in `InteractInstruction.tsx`, while `pda` and `knownAccounts` providers still live inside the framework. A few problems with this:

1. **Asymmetric pattern.** Three sibling providers, two follow the `ExternalDependency` contract, one doesn't. A future maintainer adding a fourth provider has to guess which pattern to copy. The "stale prefill" problem isn't unique to wallet - it could apply to any provider, so the right fix is either fully inside the framework or fully outside, not split.

2. **Implementation detail leaks into the call site.** `lastPrefillAddressRef` is created in the UI component and threaded into the provider as a parameter. The provider's internal bookkeeping should not appear in the component that renders accounts.

3. **Misleading eslint-disable.** The comment claims `fieldNames` is stable, but `useInstructionForm` returns a fresh `fieldNames` object on every render (see `use-instruction-form.ts:48`). The closure captures the latest value, so it's harmless in practice - but the comment will mislead the next reader.

4. **Reinvents what react-hook-form already tracks.** We're maintaining external state to answer "did the user type this value, or did we prefill it?" RHF already answers that via `isDirty`. Every prefill in this provider is called with `shouldDirty: false`, so:
   - prefilled fields → `isDirty: false`
   - user-typed fields → `isDirty: true` (RHF sets this automatically on `onChange`)

   The "match previous wallet address" check is a workaround for not using the signal RHF already gives us.


**Suggested fix:**

1. **Revert `InteractInstruction.tsx`** to the symmetric form — drop the `useRef`, `useEffect`, and the eslint-disable. Wallet rejoins the `externalDependencies` array next to the other two:
   ```ts
   const walletPrefillDependency = createWalletPrefillDependency(instruction, publicKey, fieldNames);
   const knownAccountsPrefillDependency = createKnownAccountsPrefillDependency(instruction, fieldNames);
   const pdaPrefillDependency = createPdaPrefillDependency(idl, instruction, fieldNames);
   useFormPrefill({
       config: { externalDependencies: [walletPrefillDependency, knownAccountsPrefillDependency, pdaPrefillDependency] },
       form,
   });
   ```

2. **Restore `id` and `getValue`** on `createWalletPrefillDependency` so it conforms to `ExternalDependency<PublicKey>`. Drop the `lastPrefillAddressRef` parameter — no external state needed.

3. **Replace the `isEmpty || hasPreviousWallet` check with a single dirty check** inside `onValueChange`:
   ```ts
   for (const path of signerPaths) {
       if (form.getFieldState(path).isDirty) continue;
       form.setValue(path, walletAddress as unknown as FormValue, {
           shouldDirty: false,
           shouldValidate: false,
       });
   }
   ```
   Note: `getFieldState` reads internal RHF state directly, no `formState` subscription required.

4. **Keep `shouldDirty: false`** on every `setValue` (already the case) — that's the load-bearing invariant that makes step 3 work.

C0mberry: done

app/features/idl/interactive-idl/model/form-prefill/providers/__tests__/wallet-prefill-provider.spec.ts
```diff
@@ -133,16 +126,6 @@ describe('createWalletPrefillDependency', () => {
         expect(form.getValues('accounts.testInstruction.group.nestedSigner')).toBe(walletAddress);
     });
 
-    it('should return correct dependency id and getValue', () => {
```
askov, line 136: **suggestion**: I would suggest keeping this test. And consider one integration-style test that drives the wallet-change transition through getValue.

askov (approved):

app/features/idl/interactive-idl/model/form-prefill/providers/__tests__/wallet-prefill-provider.spec.ts
C0mberry: done

Woody4618 (approved):

PR #963, opened 2026-04-22
fix: return 404 for coingecko tokens with no market data

askov:
## Description

Some tokens are listed on CoinGecko but have no trade data yet — upstream returns `200` with empty currency maps (`current_price: {}`, `market_cap: {}`, `total_volume: {}`) and `last_updated: null`. Today that falls through as a schema-validation failure and the route returns `502 Invalid response from coingecko API` with a Sentry warn, even though the upstream response is structurally fine and simply lacks USD data.

This PR adds a tiny `HasUsdMarketDataSchema` pre-check in the route. If the USD price is missing, the route returns `404 No market data` with a plain warn (no Sentry). The strict `CoinGeckoInfoSchema` check still runs afterwards, so genuine upstream schema drift still surfaces as `502` + Sentry.

The client is untouched: `useCoinGeckoVerification` already treats `404` as a permanent `FetchFailed` (`app/features/token-verification-badge/model/use-coingecko.ts`), so affected tokens just don't render the badge instead of erroring.

Example — `7wBBPnj2TA5XhB3ADm8D1odSKh9fDqAWMPaUqaMPH17e` (glowSOL, listed on CoinGecko but no trade data):

- Before: `502 {"error":"Invalid response from coingecko API"}` + Sentry warn on every request
- After:  `404 {"error":"No market data"}`, no Sentry noise

## Type of change

-   [x] Bug fix

## Testing

- Verified live against dev server:
  - `GET /api/verification/coingecko/7wBBPnj2TA5XhB3ADm8D1odSKh9fDqAWMPaUqaMPH17e` → `404`
  - `GET /api/verification/coingecko/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v` (USDC) → `200` with full market data

## Related Issues

N/A

## Checklist

## Additional Notes

The pre-check is intentionally narrow (only asserts `market_data.current_price.usd`); the full strict schema still validates the remaining fields afterwards. If upstream later changes shape in a way that removes USD *and* breaks other fields, we'll route to the `404` path rather than `502`, which is a slight loss of Sentry signal for that specific drift — the trade-off is eliminating the ongoing false-positive warns for every listed-but-empty token.


bench/BUILD.md
```diff
@@ -57,6 +57,6 @@
 | Static | `/supply` | 10 kB | 1.04 MB |
 | Static | `/tos` | 330 B | 170 kB |
 | Dynamic | `/tx/[signature]` | 60 kB | 1.53 MB |
-| Dynamic | `/tx/[signature]/inspect` | 620 B | 1.30 MB |
+| Dynamic | `/tx/[signature]/inspect` | 630 B | 1.30 MB |
 | Static | `/tx/inspector` | 630 B | 1.30 MB |
-| Static | `/verified-programs` | 10 kB | 180 kB |
+| Static | `/verified-programs` | 10 kB | 180 kB |
```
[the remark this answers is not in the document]
askov: You are not a linter, let's keep

app/api/verification/coingecko/[address]/route.ts
```diff
@@ -17,6 +17,14 @@ const COINGECKO_QUERY = [
     'tickers=false',
 ].join('&');
 
+// Some tokens are listed on CoinGecko but have no trade data yet — upstream
+// returns 200 with empty currency maps and last_updated: null. This pre-check
+// catches that case so we can return 404 (a cacheable miss) instead of letting
```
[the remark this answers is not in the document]
Comment (askov) [reply app/api/verification/coingecko/[address]/route.ts:22]: It is cacheable, but for now we don't cache. Not related to the PR

app/api/verification/coingecko/__tests__/route.test.ts
```diff
@@ -97,8 +97,45 @@ describe('CoinGecko API Route', () => {
         expect(await response.json()).toEqual(expect.objectContaining({ market_cap_rank: null }));
     });
 
+    it('should return 404 when upstream lists the token but has no USD market data', async () => {
```
rogaldh, line 100: Are you sure that the absence of market data is a valid criterion for a token that has not been verified? I am not sure about it

askov: The answer is: we don't know - that's why it's a 404. What do you suggest here?

rogaldh: I mean that Congecko does not add a token to the platform automatically. A new token should pass its verification. If we require market_data as well, this means some of the tokens will loose the verification badge as they might not have market_data yet.
I would treat market_data as optional, not required

PatSayJak (approved):

app/api/verification/coingecko/__tests__/route.test.ts
askov: Fair point, but I think it's actually a separate thing from what this PR is doing.                                                                                                                          
                                                                                                                                                                                                                                    This PR doesn't change the badge outcome at all.
Before: `empty market data -> CoinGeckoInfoSchema fails -> 502 → hook treats it as FetchFailed -> no badge`
After: `same token -> 404 -> still FetchFailed -> still no badge`.
   
The only real difference is we stop spamming Sentry and 502'ing on a response that's structurally fine, just empty.

What you're describing - "listed on CoinGecko = verified, even without market data" is a real product question, but it was already true before this PR. Honoring it means making the USD fields optional in the schema, returning 200 with a partial payload, probably a new ListedNoMarketData status, and tweaking TokenMarketData to render without a price block.

Happy to do that as a follow-up task, but it's a bigger semantic change than this bug fix. Or do you prefer a fix in this PR?

rogaldh: I see. Looks good to me. Let's clarify whether we need market_data to display a verification within a separate scope.

rogaldh (approved):
LGTM

app/api/verification/coingecko/__tests__/route.test.ts
askov: Created a task for this [HOO-485](https://linear.app/solana-fndn/issue/HOO-485/coingecko-tokens-with-no-market-data)

Woody4618 (approved):

Woody4618 (approved):

PR #1006, opened 2026-05-20
fix(feature-gates): correct off-by-one in on-chain activation epoch derivation

askov:
## Description

`/feature-gates` was showing every script-populated activation epoch one ahead of reality. Top of the table currently links SIMD-0266 to `/epoch/972`, but the feature actually went live in epoch **971** on mainnet. Same story for most rows below it.

Root cause: both `scripts/fetch_mainnet_activations.py` and `scripts/parse_feature_gates.py` did `activation_epoch = get_epoch_for_slot(activated_at_slot) + 1`, with the comment *"Technically, feature gates only become active in the following epoch"*. The premise is wrong — `apply_feature_activations` runs during the new-epoch bank transition with `bank.slot` already pointing at the first slot of the new epoch, and `compute_active_feature_set` activates a feature for any `slot >= activated_at`. The feature is therefore live in the same epoch the stored slot belongs to.

The `+ 1` has been in the script since #453, but only became visible at scale after #879 got the daily automation running. Entries hand-set or sourced from the Anza wiki happy path were unaffected, which is why the JSON ended up internally inconsistent.

This PR:
- Drops `+ 1` (and the misleading comment) from both Python scripts.
- Re-derives every populated `mainnet_/devnet_/testnet_activation_epoch` in `app/utils/feature-gate/featureGates.json` from on-chain `Feature.activated_at` across all three clusters. Most diffs are `-1`; a handful are larger where the originally-planned wiki epoch had slipped before the activation actually landed (e.g. chained-Merkle-shreds testnet `674 → 705`, priority-fee reward testnet `716 → 744`), plus a few previously-missing mainnet/devnet values filled in.

## Type of change

-   [x] Bug fix

## Screenshots

UI change is the activation-epoch column on `/feature-gates`. After the fix, the topmost rows on mainnet read:

| Feature | Before | After |
|---|---:|---:|
| SIMD-0266: Efficient Token program | 972 | **971** |
| SIMD-0444: Relax program data account check in migration | 957 | **956** |
| SIMD-0083: Relax intrabatch account locks | 954 | **953** |
| SIMD-0321: Instruction Data Pointer in VM Register 2 | 951 | **950** |
| SIMD-0185: Vote Account V4 | 950 | **949** |
| SIMD-0332: Reduce ChaCha rounds for Turbine from 20 to 8 | 947 | **946** |
| SIMD-0194: Deprecate rent exemption threshold | 944 | **943** |
| SIMD-0334: Fix alt_bn128_pairing Syscall Length Check | 943 | **942** |
| Increase Account Limit to 40% of Block CU | _(missing)_ | **878** |
| SIMD-0180: Use Vote Account Address To Key Leader Schedule | 842 | **841** |

Each row's "Activation" link now points to the actual epoch where the feature took effect on chain.

Screenshot 2026-05-20 at 18 47 19

## Testing

Live: https://explorer-git-fork-hoodieshq-fix-featur-ec02ec-solana-foundation.vercel.app/feature-gates

- Cross-checked every non-`−1` delta against the on-chain `Feature` account for that key and confirmed `floor(activated_at / slots_per_epoch)` (with testnet warm-up applied via `getEpochSchedule`) matches the new JSON value.
- Re-ran the full chain backfill against the rewritten JSON and got `changeCount: 0`, i.e. idempotent against current chain state across all three clusters.
- Loaded `/feature-gates` in the dev server after the change and verified the displayed epochs match the values in the table above; clicking an "Activation" link routes to the correct `/epoch/N` page.
- Spot-checked the two scripts: `fetch_mainnet_activations.py` and `parse_feature_gates.py` no longer offset chain-derived epochs by 1.

## Related Issues

Close [HOO-553](https://linear.app/solana-fndn/issue/HOO-553/feature-gates-page-shows-activation-epoch-one-ahead-of-actual)

## Checklist
-   [x] CI/CD checks pass

## Additional Notes
The scripts still only fill in *missing* values on subsequent runs — neither `fetch_mainnet_activations.py` nor `parse_feature_gates.py` will retroactively reconcile an entry once all three cluster epochs are populated. That's why some testnet/devnet entries had stale "planned" epochs from the Anza wiki even though the activation actually landed later (e.g. chained-Merkle-shreds testnet `674 → 705`). Out of scope for this PR; happy to follow up with a "reconcile populated entries" pass if desired.

### DIFF
--- a/app/utils/feature-gate/featureGates.json
+++ b/app/utils/feature-gate/featureGates.json
@@ -59,10 +59,10 @@
     "min_fd_versions": [],
     "min_jito_versions": [],
     "planned_testnet_order": null,
-    "testnet_activation_epoch": 716,
+    "testnet_activation_epoch": 744,
     "devnet_activation_epoch": 805,
     "comms_required": null,
-    "mainnet_activation_epoch": 742,
+    "mainnet_activation_epoch": 741,
     "description": "Passes on 100% of priority fee to validators"
   },
   {
@@ -98,7 +98,7 @@
     "testnet_activation_epoch": 715,
     "devnet_activation_epoch": 802,
     "comms_required": null,
-    "mainnet_activation_epoch": 735,
+    "mainnet_activation_epoch": 734,
     "description": "Removes legacy floating number arithmetic operation in fee calculation, making it integer-based"
   },
   {
@@ -120,7 +120,7 @@
     "testnet_activation_epoch": 746,
     "devnet_activation_epoch": 806,
     "comms_required": null,
-    "mainnet_activation_epoch": 746,
+    "mainnet_activation_epoch": 745,
     "description": "Unified syscall interface for all sysvars, without needing to include sysvar address in transactions"
   },
   {
@@ -142,7 +142,7 @@
     "testnet_activation_epoch": 746,
     "devnet_activation_epoch": 813,
     "comms_required": null,
-    "mainnet_activation_epoch": 750,
+    "mainnet_activation_epoch": 749,
     "description": "Enable new voting instruction"
   },
   {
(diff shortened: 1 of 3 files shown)


rogaldh (approved):
LGTM

Woody4618 (approved):

PR #1013, opened 2026-05-25
fix(proxy): harden metadata proxy against SSRF

askov:
## Description

Tightens the `/api/metadata/proxy` route against SSRF (initial-hop, redirect-hop, and DNS-rebinding) and a URI-decoding bypass, plus fixes a dev-only `500` for gzip-encoded upstreams.

- **Double-decode fix.** Removed the redundant `decodeURIComponent` on top of `searchParams.get()`, which already decodes. A double-encoded URI (e.g. `%2520` → `%20`) was being collapsed to a different resource, opening a URI-bypass / cache-poisoning surface and corrupting legitimately percent-encoded paths.
- **Redirect following with per-hop validation.** `redirect: 'manual'` disables automatic 3xx handling, then each hop is followed manually after re-running the private-IP / protocol guards on the `Location` target (max 3 hops, loop detection via a `visited` set). Closes an SSRF bypass where the initial hostname resolved to a public IP but the upstream returned a 302 pointing at an internal address (e.g. `169.254.169.254`). Legitimate 302s from Arweave / CDNs continue to work.
- **DNS rebinding (TOCTOU) fix.** The legacy code resolved DNS once during validation and again — independently — when `fetch()` opened the socket. A malicious authoritative DNS server could answer "public" then "private" between those two lookups. Now `lookupHostnameSafely(hostname)` resolves DNS once, validates every returned address, and returns a `net.LookupFunction` that replays the validated addresses. That function is plugged into an undici `Agent` via `connect.lookup` and passed as the fetch `dispatcher`, so the kernel never re-resolves the hostname — `connect()` only sees the IP we already approved. Re-applied per redirect hop.
- **`502 Bad Gateway` for upstream failures.** Non-2xx upstreams, redirects without a `Location` header, and exceeding the redirect ceiling now surface as 502 instead of collapsing to 500.
- **`403 Access Denied` for non-HTTP redirect targets.** A redirect to `file://`, `gopher://`, `data:` etc. is rejected.
- **Tighter error-status handling in the route.** `e instanceof StatusError && isKnownStatus(e.status)` replaces the previous `(e as StatusError)?.status` cast — synthetic objects with a `.status` field can no longer dictate response codes.
- **Per-site `StatusError` construction.** Dropped the module-load-time error singletons (`badGatewayError`, `accessDeniedError`, …) in favour of a `statusError(status, message, options?)` factory called at each throw site. Fresh stack trace + distinct site-specific message per call (Sentry can now tell "too many redirects" from "redirect loop" from "non-2xx upstream"); response body still uses canonical text from a `STATUS_MESSAGES` catalog. `cause` preserved for chaining.
- **`isRedirect` narrowed to spec-defined redirects** (`301`/`302`/`303`/`307`/`308`). `304`/`305`/`300`/`306` fall through to the upstream-error branch instead of being routed through the "Location header missing" path and producing a misleading 502.
- **Node `Readable` body in Next.js dev.** Next.js's patched `fetch` exposes gzip-encoded upstreams (e.g. Pinata via Cloudflare) as the raw Node `zlib.Gunzip` stream rather than a Web `ReadableStream`. `readBodyWithLimit` now branches on `instanceof ReadableStream` vs `instanceof Readable` and uses each stream's native API. Before this fix the route returned `500 General Error` for ~every IPFS gateway SVG in `pnpm dev`. Includes a defensive `instanceof Uint8Array` chunk guard so a string-emitting Readable surfaces as a clear error instead of silently bypassing the size limit (`chunk.byteLength` would be `undefined` → `NaN > maxSize` is always false).
- Route handler cleanup: `parseUrl` and `buildResponse` extracted, unused `params` argument and dead "missing headers" branch removed, stepdown ordering. `validateRedirectTarget` collapsed into `executeHop` so the initial hop and redirect hops run the same lookup-then-fetch path.

### Dependency

Adds `undici` as a direct dependency. Node ships it internally but doesn't expose the `Agent` constructor; the npm package's `Dispatcher` is interoperable with native `fetch` via the `dispatcher` option.

## Type of change

- [x] Other: SSRF hardening (initial-hop, redirect-hop, DNS-rebinding TOCTOU) and decode-bypass fix on the metadata proxy route.

## Testing

Unit coverage (5 spec files, 71 tests):

- `endpoint.test.ts` — feature toggle, URI validation, SSRF redirect blocking across 301/302/307/308, upstream-status passthrough, encoding round-trip across `%20` / `%23` / `%2F` / `%2520` / `%2523`.
- `fetch-resource.spec.ts` — public-IP redirect follow, private-IP redirect block, missing-Location 502, too-many-redirects 502, redirect-loop 502, 304/305 classified as upstream errors, Sentry warn on general fetch failure, **DNS-rebinding regression** (`lookupHostnameSafely` called exactly once per hop).
- `ip.spec.ts` — public/private classification, mixed-result blocking (any private address blocks), localhost short-circuit (no DNS), DNS-failure handling, pinned-lookup ignores its hostname argument, pinned-lookup respects family filter.
- `read-body-with-limit.spec.ts` — Web `ReadableStream` and Node `Readable` branches: concat, limit exceeded, stream cancel/destroy on overflow, exact-limit acceptance, non-binary-chunk guard.
- `processors.test.ts` — unchanged behaviour pinned.

### Pages checked

Each page renders an image that flows through the proxy.

- [x] [USDC (GitHub PNG)](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/address/4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU?cluster=devnet)
- [x] [Test program 1 (Pinata + redirect)](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/address/AXcxp15oz1L4YYtqZo6Qt6EkUj1jtLR6wXYqaJvn4oye?cluster=devnet)
- [x] [Test NFT 1 (`https://famousfoxes.com/hd/3104.png`)](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/address/5YqBjkPxxihGCB5zXbVQtH8yX1svcSxrGUJoKeYR56dV)

### Manual — double-decode bypass

What was wrong: the route called `decodeURIComponent` on top of `searchParams.get()`, which already decodes. An attacker could double-encode a URI to smuggle characters past validation.

**Test — double-encoded URL:** [open](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/api/metadata/proxy?uri=https%253A%252F%252Fexample.com)
Expected: **400** (invalid URL after the single decode — `https%3A%2F%2Fexample.com` is not a valid URL).

**Positive control — normal request still works:** [open](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/api/metadata/proxy?uri=https%3A%2F%2Fraw.githubusercontent.com%2Fsolana-labs%2Ftoken-list%2Fmain%2Fassets%2Fmainnet%2FSo11111111111111111111111111111111111111112%2Flogo.png)
Expected: **200** + PNG body.

### Manual — SSRF via redirect

What was wrong: the proxy followed 3xx redirects automatically. An attacker could host a URL that passes the public-IP check but redirects to an internal address (e.g. the AWS metadata endpoint at `169.254.169.254`). Now redirects are followed manually with per-hop private-IP re-validation.

**Test — initial public host, redirect target is private:**
The redirect target must resolve to a private/link-local IP. The cleanest harness is a one-off public endpoint that returns `302 Location: http://169.254.169.254/latest/meta-data/` (e.g. via a temporary `httpbin`-like instance or a webhook.site echo). Then hit `https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/api/metadata/proxy?uri=<redirect-source-url>`.

Expected: **403 Access Denied** (logged as `Hostname resolution blocked (SSRF protection)`).

Same behaviour is exercised in `fetch-resource.spec.ts` against 301/302/307/308 with `169.254.169.254` as the redirect target.

### Manual — DNS rebinding (TOCTOU)

What was wrong: DNS was resolved twice — once during `checkURLForPrivateIP` validation and again inside `fetch()`. An attacker controlling an authoritative DNS server could return a public IP for the first lookup and a private IP for the second.

Why this is hard to test manually: requires a malicious authoritative DNS server that alternates responses (TTL=0 records, dnschef / dnsmasq scripted answers, …). Can't be reproduced with curl alone.

The fix is structural — see `fetch-resource.ts:doFetch` and `ip.ts:lookupHostnameSafely`. The resolved addresses from the validation step are pinned into an undici `Agent`'s `connect.lookup` and passed via the fetch `dispatcher`, so the kernel never performs a second resolution. `fetch-resource.spec.ts` has a regression test asserting `lookupHostnameSafely` is called exactly once per hop.

### Manual — gzip-encoded SVG (Pinata)

Direct proxy hit against an IPFS gateway that returns `image/svg+xml; charset=UTF-8` with `content-encoding: gzip`: [open](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/api/metadata/proxy?uri=https%3A%2F%2Fchocolate-necessary-mandrill-958.mypinata.cloud%2Fipfs%2Fbafkreicaq2d53b6m5qr727krs5xhmfnwr7ivfox64zq4xvj744c3cfopzi)

Expected: **200** + SVG body. Before this PR this returned `500 General Error` in dev due to Next.js exposing the raw `zlib.Gunzip` stream as `response.body`.

### Manual — protocol guard

[open](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/api/metadata/proxy?uri=ftp%3A%2F%2Fexample.com%2Ffile.json) — expected: **400**.

### Manual — private-IP guard

[open](https://explorer-git-fork-hoodieshq-fix-proxy-7954c8-solana-foundation.vercel.app/api/metadata/proxy?uri=http%3A%2F%2F127.0.0.1%2F) — expected: **403**.

## Out of scope (follow-ups)

- **Rate limiting.** No request quota on this route yet. Recommend a Vercel Firewall custom rule on `/api/metadata/proxy` (deny > N/min/IP) before this is widely advertised.

## References

- [OWASP — Double Encoding](https://owasp.org/www-community/Double_Encoding)
- [OWASP — Server-Side Request Forgery (SSRF)](https://owasp.org/Top10/A10_2021-Server-Side_Request_Forgery_%28SSRF%29/)
- [OWASP — SSRF Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html)
- [HackTricks — DNS Rebinding / TOCTOU](https://book.hacktricks.wiki/en/pentesting-web/ssrf-server-side-request-forgery/url-format-bypass.html#dns-rebinding-toctou)
- [CWE-918: Server-Side Request Forgery](https://cwe.mitre.org/data/definitions/918.html)
- [undici — Agent options](https://undici.nodejs.org/#/docs/api/Agent)

## Related Issues

Closes [HOO-336](https://linear.app/solana-fndn/issue/HOO-336/enable-metadata-service)
Closes [HOO-340](https://linear.app/solana-fndn/issue/HOO-340/check-potential-double-decoding-issue)
Closes [HOO-338](https://linear.app/solana-fndn/issue/HOO-338/ssrf-issue-with-potential-redirect)
Closes [HOO-337](https://linear.app/solana-fndn/issue/HOO-337/potential-dns-rebinding-issue)

## Checklist

- [x] All tests pass locally and in CI
- [x] CI/CD checks pass
- [x] For security-related features, I have included links to related information


app/api/metadata/proxy/feature/fetch-resource.ts
```diff
@@ -0,0 +1,168 @@
...
+        });
+        throw accessDeniedError;
+    }
+
+    return nextUrl;
+}
+
+function isRedirect(response: Response): boolean {
+    return response.status >= 300 && response.status < 400;
+}
```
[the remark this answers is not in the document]
askov: narrowed `isRedirect` to 301/302/303/307/308 in 53e512fb, with a test for 304/305 falling through to the upstream-error branch.

app/api/metadata/proxy/feature/read-body-with-limit.ts
```diff
@@ -17,7 +35,25 @@ export async function readBodyWithLimit(response: Response, maxSize: number): Pr
...
     }
-    const buffer = new Uint8Array(received);
+    return concat(chunks, received);
+}
+
+async function collectFromNodeStream(body: Readable, maxSize: number): Promise<ArrayBuffer> {
+    const chunks: Uint8Array[] = [];
+    let received = 0;
+    for await (const chunk of body) {
+        received += chunk.byteLength;
```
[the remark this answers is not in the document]
askov: added an `instanceof Uint8Array` guard in 23be9c46 — throws fast and destroys the stream on a string chunk instead of silently bypassing the size limit. skipped the `Buffer.from(chunk as string)` path since the encoding would be ambiguous and we ban `as` casts here.

app/api/metadata/proxy/feature/errors.ts
```diff
@@ -1,24 +1,28 @@
...
-export const unsupportedMediaError = new StatusError('Unsupported Media Type', { cause: 415 });
+export const unsupportedMediaError = new StatusError('Unsupported Media Type', { status: 415 });
 
-export const generalError = new StatusError('General Error', { cause: 500 });
+export const generalError = new StatusError('General Error', { status: 500 });
 
-export const gatewayTimeoutError = new StatusError('Gateway Timeout', { cause: 504 });
+export const badGatewayError = new StatusError('Bad Gateway', { status: 502 });
+
+export const gatewayTimeoutError = new StatusError('Gateway Timeout', { status: 504 });
```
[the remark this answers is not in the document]
askov: fair, refactored in 7480c89b — `STATUS_MESSAGES` map for the response body, `statusError(status, message, options?)` factory at each throw site. fresh stack per call, distinct site-specific messages in logs, `cause` chained where there is an underlying error. dropped the `errors[]` map and the named singletons.

app/api/metadata/proxy/feature/processors.ts
```diff
@@ -62,12 +60,11 @@ export async function processTextAsJson(data: Response) {
...
-            throw errors[413];
+            throw statusError(413, 'Text body exceeds max size', { cause: error });
         } else if (error instanceof SyntaxError) {
-            throw errors[415];
-        } else {
-            Logger.debug('[api:metadata-proxy] Failed to process text-as-JSON data', { error });
-            throw errors[500];
+            throw statusError(415, 'Malformed JSON in text upstream response', { cause: error });
         }
+        Logger.debug('[api:metadata-proxy] Failed to process text-as-JSON data', { error });
```
rogaldh: Let's increase the level and classify as a warning at least to monitor that in the logs

app/api/metadata/proxy/__tests__/ip.spec.ts
```diff
@@ -21,91 +21,166 @@ vi.mock('dns', async () => {
...
+// The pinned lookup is the core of the DNS-rebinding fix: it must replay the
+// pre-validated addresses no matter what the kernel would have resolved.
+describe('lookupHostnameSafely — pinned lookup behaviour', () => {
     beforeEach(() => {
         vi.clearAllMocks();
     });
 
-    test('should block localhost', async () => {
-        await expect(checkURLForPrivateIP('http://localhost')).resolves.toBe(true);
+    test('pinned lookup should return the validated address regardless of the hostname argument', async () => {
```
rogaldh: According to the latest rules, tests start with "should"

app/api/metadata/proxy/__tests__/fetch-resource.integration.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+afterAll(async () => {
+    await new Promise<void>((resolve, reject) => server.close(err => (err ? reject(err) : resolve())));
+});
+
+beforeEach(() => {
+    // Bypass the private-IP guard for 127.0.0.1 by stubbing
+    // lookupHostnameSafely. The pinned lookup we return resolves the loopback
+    // hostname to 127.0.0.1 so undici can actually connect.
+    vi.mocked(lookupHostnameSafely).mockReset();
+    // Use the real `makePinnedLookup` path by stubbing only the validation
```
rogaldh: Some lines are duplicated here. i'd propose rewriting the comment

rogaldh (approved):
Except these ^, LGTM

app/api/metadata/proxy/feature/processors.ts
askov: done

app/api/metadata/proxy/__tests__/ip.spec.ts
askov: fixed

app/api/metadata/proxy/__tests__/fetch-resource.integration.spec.ts
askov: updated

Woody4618 (approved):

PR #1016, opened 2026-05-25
feat: Verified programs search

Woody4618:
## Description

Uses the osec api to get all verified programs, then fetches the IDLs and writes it all into a json if there is changes. 
Once per day. 
The search then uses that json to show verified programs in the search.


package.json
```diff
@@ -63,6 +63,7 @@
         "@solana-program/token": "0.13.0",
         "@solana-program/token-2022": "0.9.0",
         "@solana/buffer-layout": "3.0.0",
+        "@solana/idl": "0.1.0",
```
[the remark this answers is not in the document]
Woody4618: Yeah i need to update the package

app/features/search/model/verified-programs-search-provider.ts
```diff
@@ -0,0 +1,55 @@
+import { Cluster } from '@utils/cluster';
+
+import verifiedPrograms from '@/public/verified-programs.json';
```
rogaldh, line 4: I think that it is better to serve the json via api endpoint. To not include it into build, but fetch it. That allows it to be gzipped and reuse browser caching

app/features/search/model/verified-programs-search-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { SearchGroup } from '../lib/filter-tabs';
+import type { SearchContext, SearchOptions, SearchProvider } from '../lib/types';
+
+type VerifiedProgramEntry = {
```
rogaldh, line 9: This type dos not satisfy the interface that is used here: https://github.com/solana-foundation/explorer/pull/1016/changes#diff-2bc02ccb6b744e96e51fd3368e858f72bc98c323f548e22c5c99bedaa7046d5fR41-R46

scripts/update-verified-programs.ts
```diff
@@ -0,0 +1,230 @@
...
+    // Only include programs confirmed as verified
+    const verifiedAddresses = addresses.filter(addr => statuses.get(addr)?.is_verified === true);
+    const skipped = addresses.length - verifiedAddresses.length;
+    if (skipped > 0) {
+        console.log(`  Excluded ${skipped} programs (unverified or status unavailable)`);
+    }
+
+    const programs: VerifiedProgram[] = verifiedAddresses
+        .map(addr => {
+            const status = statuses.get(addr);
```
rogaldh, line 109: I think it is worth checking that the provided address is a valid PublicKey to not store invalid results.
It is good to check that incoming data satisfies the expected structure anyways

scripts/update-verified-programs.ts
```diff
@@ -0,0 +1,230 @@
...
+            const idlName = idlNames.get(addr);
+            const repoUrl = status?.repo_url || undefined;
+            const name = idlName || deriveNameFromRepo(repoUrl) || `${addr.slice(0, 12)}...`;
+
+            return {
```
rogaldh, line 114: Here we prepare every record, but we rely on the fact that every helper will receive proper data, which is provided by the remote API.
I'd suggest validating it before we return it to make sure that satisfies expected shape.
See suggestions

scripts/update-verified-programs.ts
```diff
@@ -0,0 +1,230 @@
...
+ * IDL-derived names and repo URLs, and writes a compact JSON file.
+ *
+ * Usage:
+ *   MAINNET_RPC_URL=https://... pnpm exec tsx scripts/update-verified-programs.ts
+ *
+ * Output:
+ *   - public/verified-programs.json
+ */
+
+import { address, createSolanaRpc } from '@solana/kit';
```
rogaldh, line 15:
```suggestion
import { address, createSolanaRpc } from '@solana/kit';
import { writeFile } from 'fs/promises';
import { dirname, join } from 'path';
import { fileURLToPath } from 'url';
import { array, create, Infer, optional, refine, string, type } from 'superstruct';

const Base58Address = refine(string(), 'Base58Address', value =>
    /^[1-9A-HJ-NP-Za-km-z]{32,44}$/.test(value) || `not a base58 address: ${value}`,
);
const NonEmptyString = refine(string(), 'NonEmptyString', value => value.length > 0 || 'must be non-empty');
const HttpsUrl = refine(string(), 'HttpsUrl', value => {
    try {
        return new URL(value).protocol === 'https:' || `not https: ${value}`;
    } catch {
        return `not a URL: ${value}`;
    }
});
const IsoDateTime = refine(string(), 'IsoDateTime', value =>
    /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}$/.test(value) || `not ISO datetime: ${value}`,
);

const VerifiedProgramStruct = type({
    address: Base58Address,
    name: NonEmptyString,
    repoUrl: optional(HttpsUrl),
    verifiedAt: optional(IsoDateTime),
});
const VerifiedPrograms = array(VerifiedProgramStruct);
type VerifiedProgram = Infer<typeof VerifiedProgramStruct>;
```

scripts/update-verified-programs.ts
```diff
@@ -0,0 +1,230 @@
...
+            return {
+                address: addr,
+                name,
+                repoUrl,
+                verifiedAt: status?.last_verified_at?.split('.')[0] ?? undefined,
+            };
+        })
+        .sort((a, b) => a.address.localeCompare(b.address));
+
+    if (programs.length === 0) {
```
rogaldh, line 123:
```suggestion
    if (programs.length === 0) {
        console.error('\nNo verified programs found -- aborting to avoid overwriting existing data');
        process.exit(1);
    }

    const validated = create(programs, VerifiedPrograms);
    await writeFile(OUTPUT_PATH, `${JSON.stringify(validated, undefined, 2)}\n`, 'utf8');
    console.log(`\nWritten ${validated.length} programs to ${OUTPUT_PATH}`);
```

app/features/search/model/verified-programs-search-provider.ts
Woody4618: The search provider's search() is synchronous, so switching to an async fetch would require refactoring the search infrastructure. The JSON is ~30KB. Good follow-up for later.

.github/workflows/update-verified-programs.yaml
```diff
@@ -0,0 +1,60 @@
...
+          git diff --exit-code public/verified-programs.json && echo "changed=false" >> $GITHUB_OUTPUT || echo "changed=true" >> $GITHUB_OUTPUT
+
+      - name: Commit and push
+        if: steps.diff.outputs.changed == 'true'
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add public/verified-programs.json
+          git commit -m "chore: update verified programs list"
+          git push
```
rogaldh: Are you sure the script is allowed to make a push?

rogaldh: At least we need a way to make a revert

Woody4618: I switched it to do PRs

rogaldh (approved):
LGTM

app/features/search/model/registry.ts
```diff
@@ -14,11 +14,13 @@ import { specialSearchProvider } from './special-search-provider';
 import { sysvarSearchProvider } from './sysvar-search-provider';
 import { tokenSearchProvider } from './token-search-provider';
 import { transactionSearchProvider } from './transaction-search-provider';
+import { verifiedProgramsSearchProvider } from './verified-programs-search-provider';
 
 const allProviders: SearchProvider[] = [
     ...(process.env.NEXT_PUBLIC_DISABLE_TOKEN_SEARCH ? [] : [tokenSearchProvider]),
     explorerUrlSearchProvider,
     programSearchProvider,
+    verifiedProgramsSearchProvider,
```
rogaldh, line 23: Oh. It looks like we may have duplicated results if one program is within two lists at the same time

Woody4618: I added dedup and marked verified programs as such 
image

Woody4618:
@copilot resolve the merge conflicts in this pull request

Woody4618:
@greptile-apps please update your review

PR #1024, opened 2026-05-26
Add slot filters for addresses' transactions

anselsol:
## Description

The goal of this PR is to provide a way for users of [explorer.solana.com](https://explorer.solana.com/) to filter an address's transactions — by slot range, block time, and status — without relying on any kind of indexing for the explorer.

To do this, the PR switches the transaction history from the `getSignaturesForAddress` + per-signature `getTransaction` pattern over to Triton's  [`getTransactionsForAddress`](https://docs.triton.one/chains/solana/gettransactionsforaddress) method, which combines address-history lookup, filtering, and pagination into a single RPC call.

  The filter API exposed in the UI maps **one-to-one** onto the method's `filters` object, so the explorer's URL params share the same names and shape as the RPC:

  | Filter | URL param(s) | gTFA filter path |
  |---|---|---|
  | Slot range | `slot.gte`, `slot.lte` | `filters.slot.{gte,lte}` |
  | Block time range | `blockTime.gte`, `blockTime.lte` | `filters.blockTime.{gte,lte}` |
  | Status | `status` (`succeeded` / `failed`) | `filters.status` |

Pagination uses the opaque `paginationToken` cursor returned by the method (replacing the old trailing-signature `before` cursor); loading stops once the RPC returns a null token.

### Graceful degradation on non-Triton/non-Helius endpoints

`getTransactionsForAddress` is a Triton extension. If the configured RPC endpoint doesn't implement it (the call returns a JSON-RPC "method not found"), the explorer:
- falls back to the standard `getSignaturesForAddress` so transaction history still loads, and
- **disables the filter UI** (the "Filters" button is shown disabled, and any active filter params are cleared) — since the fallback can't honour block-time/status filters, this avoids showing unfiltered results next to misleading "active filter" pills.

Detection is behavioral (based on the method-not-found response), so it works for Triton, Helius, or any other endpoint that implements the method — no provider allowlist is hardcoded.

  ### How it's implemented in the UI:

  *Funnel 1: user filters from the UI*
  1. User lands on [https://explorer.solana.com/address/LGDSXVcDx4Ynw7UXavGEe5nwzyUZZ5d3sLkwYk26LUf](https://explorer.solana.com/address/LGDSXVcDx4Ynw7UXavGEe5nwzyUZZ5d3sLkwYk26LUf)
  2. User clicks the "Filters" button, next to the "Refresh" button, and sets any of the slot bounds, block-time bounds, or status
  3. The selection is written to the URL (`slot.gte`, `slot.lte`, `blockTime.gte`, `blockTime.lte`, `status`) and picked up by `app/components/account/history/HistoryFilterBar.tsx`
  4. That address's history cache is reset (superseding any in-flight request) and transactions are re-fetched directly from the RPC with the corresponding `filters` object
  6. User experience is then similar to an unfiltered view (load next transactions via the "Load More" button; transactions stop loading once the filtered range is exhausted)

  *Funnel 2: the URL already contains filter params*
  1. User lands on
  [https://explorer.solana.com/address/LGDSXVcDx4Ynw7UXavGEe5nwzyUZZ5d3sLkwYk26LUf?slot.gte=375952400](https://explorer.solana.com/address/LGDSXVcDx4Ynw7UXavGEe5nwzyUZZ5d3sLkwYk26LUf?slot.gte=375952400)
  2. The params are parsed by `useHistoryFilters` in `app/components/account/history/HistoryFilterBar.tsx`
  3. Transactions are fetched directly from the RPC using the parsed `filters` object, and the active filters render as removable pills above the table
  5. User experience is then similar to an unfiltered view (load next transactions via the "Load More" button; transactions stop loading once the filtered range is exhausted)

  ## Type of change

  -   [ ] Bug fix
  -   [x] New feature
  -   [ ] Protocol integration
  -   [ ] Documentation update

## Screenshots

Screenshot 2026-04-28 at 15 15 54
 -
Screenshot 2026-04-28 at 15 15 59

Screenshot 2026-05-26 at 13 34 30

## Testing

`app/components/account/history/__tests__/HistoryFilterBar.spec.tsx`: tests the filters button, URL param round-tripping, and pills behavior across slot, block-time, and status filters

`app/providers/accounts/__tests__/history.spec.tsx`: tests that the provider maps filters onto the `getTransactionsForAddress` request, threads the `paginationToken` correctly when loading more (including short pages that still carry a token), discards superseded in-flight responses on filter change, clears only the target address on reset, and falls back to `getSignaturesForAddress` (disabling filtering) when the method isn't available

  ## Related Issues

## Checklist

-   [ ] I have run `build:info` script to update build information
-   [x] CI/CD checks pass


app/components/account/history/TransactionHistoryCard.tsx
```diff
@@ -38,6 +43,18 @@ export function TransactionHistoryCard({ address }: { address: string }) {
...
+    // only, so we drop the cached entries before refetching to avoid mixing pre- and
+    // post-filter results in combineFetched.
+    const previousFiltersKey = React.useRef(filtersKey);
+    React.useEffect(() => {
+        if (previousFiltersKey.current !== filtersKey) {
+            previousFiltersKey.current = filtersKey;
+            clearHistories();
+            refresh();
+        }
+    }, [filtersKey, clearHistories, refresh]);
```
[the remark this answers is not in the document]
anselsol: Added a per-address generation counter in `HistoryProvider` (GenerationContext). The new useResetAccountHistory(address) hook:
  1. bumps the address's generation (so any in-flight response is recognized as superseded),
  2. evicts the address from the inFlight set (so the immediate refetch isn't deduplicated and silently dropped),
  3. clears only that address's cache entry.

`fetchAccountHistory` captures the generation at dispatch time and discards its result if superseded. The card now calls `resetHistory(address)` instead of the old global clear. 

New test: an in-flight unfiltered response that resolves after a filter change no longer overwrites the filtered data.

app/providers/accounts/history.tsx
```diff
@@ -178,6 +263,17 @@ async function fetchAccountHistory(
...
+export function useClearAccountHistories() {
+    const { url } = useCluster();
+    const dispatch = React.useContext(DispatchContext);
+    if (!dispatch) {
+        throw new Error(`useClearAccountHistories must be used within a HistoryProvider`);
+    }
+    return React.useCallback(() => {
+        dispatch({ type: ActionType.Clear, url });
+    }, [dispatch, url]);
+}
```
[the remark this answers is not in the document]
anselsol: Clear now takes an optional key the reducer deletes just that entry when present (with a URL guard), falling back to the full clear when omitted so the URL-change clear and all other providers are unchanged.
   
New test confirms resetting address A leaves address B's history intact.

app/providers/accounts/history.tsx
```diff
@@ -116,14 +133,77 @@ async function fetchParsedTransactions(url: string, transactionSignatures: strin
...
+        }),
+        headers: { 'content-type': 'application/json' },
+        method: 'POST',
+    });
+    const json = await response.json();
+    if (json.error) {
+        throw new Error(json.error.message ?? 'getTransactionsForAddress failed');
+    }
+    return json.result as GetTransactionsForAddressResult;
+}
```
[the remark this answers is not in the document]
anselsol: `getTransactionsForAddress` errors now carry the JSON-RPC code. On a `method-not-found` (-32601 or matching message), `fetchAccountHistory` falls back to `connection.getSignaturesForAddress`, whose cursor is the trailing before signature (threaded from the load-more branch). Slot bounds pass through as the Hydrant untilSlot/beforeSlot extension; block-time/status filters don't apply on the legacy path (as flagged in the option). Two new tests: fallback fires + maps slot bounds on -32601, and a generic error does not fall back.

app/providers/accounts/history.tsx
```diff
@@ -116,17 +141,124 @@ async function fetchParsedTransactions(url: string, transactionSignatures: strin
...
+    if (options.filters.slot?.gte !== undefined) rpcOptions.untilSlot = options.filters.slot.gte;
+    if (options.filters.slot?.lte !== undefined) rpcOptions.beforeSlot = options.filters.slot.lte;
+    const fetched = await connection.getSignaturesForAddress(
+        pubkey,
+        rpcOptions as Parameters<Connection['getSignaturesForAddress']>[1],
+    );
+    return {
+        fetched,
+        foundOldest: fetched.length < options.limit,
+        paginationToken: null,
```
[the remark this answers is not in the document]
anselsol: Removed all filters when method non found is triggered

app/providers/accounts/history.tsx
```diff
@@ -138,18 +288,42 @@ async function fetchAccountHistory(
...
+            filters: options.filters,
+            limit: options.limit,
+            paginationToken: options.paginationToken,
+        });
         history = {
-            fetched,
-            foundOldest: fetched.length < options.limit,
+            fetched: result.data,
+            // A null/absent cursor or a short page means there is nothing more to load.
+            foundOldest: !result.paginationToken || result.data.length < options.limit,
```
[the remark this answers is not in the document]
anselsol: Fixed

rogaldh:
Hi. Thank you for contribution!
Could you please make a rebase from the latest master? `eslint.config` was improved. To ensure that this change does not break those

michellebrant33-boop (approved):

anselsol:
Gm gm @rogaldh ! Rebased and fixed code for ESLint new rules, I've also ran the full CI locally and it goes through 🤝

app/providers/accounts/history.tsx
```diff
@@ -165,17 +199,127 @@ async function fetchParsedTransactions(url: string, cluster: Cluster, transactio
...
+    const response = await fetch(url, {
+        body: JSON.stringify({
+            id: 1,
+            jsonrpc: '2.0',
+            method: 'getTransactionsForAddress',
+            params: [address, params],
+        }),
+        headers: { 'content-type': 'application/json' },
+        method: 'POST',
+    });
```
askov, line 296: **issue (blocking):** `response.ok` isn't checked before `response.json()`.

If the endpoint answers with a 5xx HTML page, a 429 with no body, or anything that isn't a JSON-RPC envelope, one of two things happens:

- `response.json()` throws a parse error — caught upstream as a generic failure (fine, but the message is unhelpful), or
- the body parses but has neither `error` nor `result` — `json.error` is falsy, so we fall through and `return json.result as ...` returns `undefined`, which then explodes on `result.data` in the caller.

**suggestion:** Guard right before `response.json()`:

```ts
if (!response.ok) {
    throw new Error(`getTransactionsForAddress HTTP ${response.status}`);
}
```

The `isMethodNotFound` check below will correctly treat this as a non-method-not-found error and avoid spuriously falling back.

app/providers/accounts/history.tsx
```diff
@@ -165,17 +199,127 @@ async function fetchParsedTransactions(url: string, cluster: Cluster, transactio
...
+    return json.result as GetTransactionsForAddressResult;
+}
+
+// getTransactionsForAddress is a Triton extension; standard RPC nodes answer with a
+// JSON-RPC "method not found" (-32601). We use that to fall back to getSignaturesForAddress.
+function isMethodNotFound(error: unknown): boolean {
+    const e = error as { code?: number; message?: string };
+    if (e?.code === -32601) return true;
+    const message = typeof e?.message === 'string' ? e.message.toLowerCase() : '';
+    return message.includes('method not found') || message.includes('unsupported method');
```
askov, line 327: **issue (non-blocking):** The `-32601` code check is the right contract; the substring fallback on `message` is loose enough to false-positive.

Any transient server error whose message happens to contain "method not found" (a proxy error page, a generic upstream message, etc.) will silently downgrade the endpoint to the legacy path and disable filtering for the rest of the session.

**suggestion:** Drop the string check entirely, or only consult it when `code` is unset:

```ts
if (e?.code === -32601) return true;
if (typeof e?.code === 'number') return false; // trust the code
const message = ...
```

app/providers/accounts/history.tsx
```diff
@@ -229,6 +402,39 @@ async function fetchAccountHistory(
     });
 }
 
+export function useClearAccountHistories() {
```
askov: **nitpick (non-blocking):** `useClearAccountHistories` is exported but I don't see a caller — the existing URL-change handler in `HistoryProvider` still dispatches `{ type: ActionType.Clear, url }` inline.

**suggestion:** Delete this hook, or use it in place of the inline dispatch so the public surface of the provider matches what's actually consumed.

app/features/transaction-history/ui/TransactionHistoryCard.tsx
```diff
@@ -40,6 +57,27 @@ export function TransactionHistoryCard({ address }: { address: string }) {
...
+            previousFiltersKey.current = filtersKey;
+            resetHistory(address);
+            refresh();
+        }
+    }, [filtersKey, address, resetHistory, refresh]);
+
+    // If the endpoint turns out not to support filtering, drop any active filters so the
+    // (unfiltered) results aren't shown alongside misleading filter chips.
+    React.useEffect(() => {
+        if (!filtersSupported && hasActiveFilters) {
```
askov, line 75: **issue (non-blocking):** Cold-loading a non-Triton endpoint with filter params in the URL causes a redundant refetch + brief content flash.

Chain of events:

1. Initial render → `if (!history) refresh()` sends gTFA with the URL filters.
2. gTFA returns method-not-found → `markUnsupported()` flips `filtersSupported` to `false`.
3. Fallback `fetchViaSignatures` runs with `untilSlot`/`beforeSlot`, results are dispatched and rendered.
4. This effect now sees `!filtersSupported && hasActiveFilters` → `clearFilters()`.
5. URL change → `filtersKey` change → the effect above (`resetHistory` + `refresh`) fires → second method-not-found round-trip → unfiltered results render.

That's two full request cycles and a visible flip from slot-filtered to unfiltered on every cold load.

**suggestion:** A couple of ways out:

- Mark unsupported *before* dispatching the first fallback's results, and accept that the first render shows slot-bounded results (the user *asked* for them in the URL, after all) — then this clear effect can be removed, and the chips for `status`/`blockTime` can just be hidden when `!filtersSupported`.
- Or, skip the slot translation in `fetchViaSignatures` so the first response matches the post-clear state and the second fetch returns identical data (still wasteful but at least no flash).

The in-popover trigger path is unaffected — this only matters for users who land on a deep-linked filter URL.

askov (commented):
Nice migration — the URL ↔ filters ↔ RPC mapping is clean, and the generation-counter pattern for invalidating in-flight requests on filter change is the right primitive (the dropped-stale-write test is exactly the right shape). A few inline notes on things I'm fairly sure are real, mostly around the fallback path and HTTP-level error handling.

app/components/account/history/HistoryFilterBar.tsx
```diff
@@ -0,0 +1,380 @@
...
+                            variant="dark"
+                            value={blockTimeGteDraft}
+                            aria-invalid={timeRangeInvalid}
+                            onChange={e => setBlockTimeGteDraft(e.target.value)}
+                        />
+                    </div>
+                    <div className="e-flex e-flex-col e-gap-1">
+                        <label className="e-text-xs e-text-neutral-300">Block time ≤</label>
+                        <Input
+                            type="datetime-local"
```
askov, line 344: **issue**: I understand this wasn't introduced by the PR, but it surfaces the bug clearly: the icon is the wrong color and mostly invisible. I think it should be fixed here or in a separate PR before merging.

Image

app/providers/accounts/history.tsx
anselsol: Great finding! Fixed in next commit

app/providers/accounts/history.tsx
anselsol: Yes absolutely, fixed in next commit 🤝

app/providers/accounts/history.tsx
anselsol: Dead code, nuking it

app/features/transaction-history/ui/TransactionHistoryCard.tsx
anselsol: Ok so now that I'm reading this again, including the filters at all was a mistake (standard solana rpc api doesn't support those). I cleaned it up in the next commit

app/components/account/history/HistoryFilterBar.tsx
anselsol: Fixed

anselsol:
@askov all your findings are fixed 🤝

anselsol:
What are the next steps here guys?

jeduardo2092-design:
> @askov all your findings are fixed 🤝

anselsol:
Gm gm guys, how can I help push this through the finish line?

Justfunnyboy1 (approved):

askov:
> Gm gm guys, how can I help push this through the finish line?

Sorry for the long response. Please do a rebase/merge and make sure nothing breaks regarding the UI (there was a huge ui migration, no more e- prefixes for tailwind classes). We will try to merge it as soon as possible

askov:
@anselsol Could you please run the prettier/eslint, CI is failing

askov:
The current state doesn't work on preview with triton, investigation needed (works for other preview branches):
Screenshot 2026-07-01 at 15 48 01

Woody4618:
Hey thanks for this, our explorer API key does not Support the getTransactionsForAddress call as of now. We need to put this one on ice for a bit. Let's continue once we have that supported. We might get that support soon, so you could already fix the linting and tests and set up singed commits.

askov:
@anselsol Hey! Now we have support for getTransactionsForAddress, so we could merge the feature. Please rebase and make sure that CI checks pass (you could check tests/prettier/eslint locally). Thanks in advance.

anselsol:
@askov will take care of it this week

anselsol:
@askov ready for tests on your end

askov:
> @askov ready for tests on your end
LGTM, but please fix the CI errors (prettier/storybook)

askov (dismissed):

anselsol:
@askov CI fixed!

Woody4618 (approved):

Woody4618 (approved):

PR #1033, opened 2026-05-28
bugfix: Add check to only show verified for verified programs

Woody4618:
## Description
There was a bug that the text for verified programs was also shown for tokens. 
image

image


rogaldh (approved):
LGTM

PR #1050, opened 2026-06-02
feat(metadata-proxy): prep for enabling — client image fallback, 4 MB cap, browser caching

askov:
## Description

Preparation for **enabling** the metadata proxy (HOO-533). Turning the proxy on reroutes every metadata/image request through one function and changes the failure surface, so this PR makes that surface safe to ship: it surfaces *why* an image failed and handles the oversize/failure tail in the UI, sets a cap chosen for real traffic rather than one sized to fit every image, and hardens the input edges so a single bad on-chain value can't crash a card or silently disable a guard. The proxy stays behind `NEXT_PUBLIC_METADATA_ENABLED`.

**Graceful client fallback with a failure reason.** New `ProxiedImage` (`app/features/metadata/ui`) built over two proxy-agnostic primitives — `ImageWithFallback` and `ExternalResourceLink` (`app/components/shared/ui/image`). On any load failure it degrades to a Solana-logo placeholder that **names why** the image couldn't be shown. Because a browser `` can't read the HTTP status, on failure the component re-`fetch`es the same-origin proxy `src` to read `response.status` and maps it to a per-status reason — `413` "exceeds maximum size", `415` "unsupported type", `502`/unreachable upstream "source unavailable", `504` "timed out". Error responses carry a short browser-only `Cache-Control: private, max-age=30`, so the failed `` request primes the cache and the probe is *usually* a cache hit rather than a fresh round-trip — and it hits our same-origin proxy, never the upstream, so it adds no privacy leak. A `504` whose error response wasn't cached can re-incur the upstream timeout, so the loading skeleton is **held** until the reason resolves (`useImageFailureReason` exposes an `idle`/`probing`/`resolved` status) rather than flashing a reasonless fallback — the failure state appears once, fully formed. The reason rides as the placeholder's accessible name and hover tooltip; a `src` with no readable status (proxy disabled, or a non-HTTP passthrough) shows a generic reason without probing.

- NFT displays (`NFTArt` → `MetaplexNFTHeader`/`CompressedNftCard`, `NFTokenAccountSection`) opt into an explicit **"View original"** link (`showOriginalLink`; `target="_blank"`, `rel="noopener noreferrer"`) shown beside the image in every state, with the failure reason in its info tooltip.
- Small logo slots (`AccountHeader`, `OwnedTokensCard`, `ProgramHeader`) use the **default** fallback — no `fallback` prop, no link: a slot-shaped Solana logo with the reason on hover, so a failed/oversize load no longer renders a broken-image icon. They also stop calling `getProxiedUri` directly (the component does it internally).
- Removes the dead, env-gated "view original" path and the `NEXT_PUBLIC_VIEW_ORIGINAL_DISPLAY_ENABLED` flag.

**External-link scheme guard.** Ships the reusable safety primitives the "View original" link rides on: `getSafeExternalHref` / `parseUrl` (`app/shared/lib/url.ts`, `new URL`-parse + absolute-`http(s)`-only) and a safe-by-default `ExternalLink` component (`app/components/shared/ui/external-link.tsx`) that owns `target`/`rel` and renders nothing for an unsafe href — closing the `javascript:`/`data:` stored-XSS vector for on-chain URLs. This PR guards its own sink and lands the `harden-external-link-schemes` proposal/spec/tasks for the follow-up that migrates the remaining raw-`<a>` sinks (deferred to keep that cross-cutting sweep on its own review).

**Size cap.** Set the default `MAX_SIZE` to **4 MB** (overridable via `NEXT_PUBLIC_METADATA_MAX_CONTENT_SIZE`). Net of this PR the cap rises from 1 MB on `master`; an earlier commit here had bumped it to 10 MB alongside the Sentry oversize reporting, and the cap work settles it at 4 MB — just under Vercel's ~4.5 MB buffered-response limit (a thin ~0.5 MB margin), so our own cap stays the binding constraint while fitting more images. The rejected tail degrades to the fallback above instead of a platform-layer rejection or a broken image; the success-path size log (below) is the signal to trim the cap back if real payloads crowd the 4.5 MB line.

**Browser caching + invocation cap.** Successful (2xx) responses carry a browser-only `Cache-Control: public, max-age=86400` (1 day) — no `Vercel-CDN-Cache-Control`, so they are **not** edge-cached. The proxied content is effectively immutable (content-addressed Arweave/IPFS, static CDN assets), so a day-long browser cache is safe; `immutable` is omitted so a rare in-place overwrite is still picked up once the window lapses. Repeat views from the same browser skip the round-trip, while cold and cross-client views still invoke the function. The upstream's own `Cache-Control` is not forwarded (often `no-store`/`no-cache`/`private`), and the upstream `ETag` is not forwarded either — the route does no conditional revalidation (never reads `If-None-Match` nor emits `304`), so a forwarded validator would be inert. Errors stay uncached at the edge. A platform backstop `export const maxDuration = 15` bounds the whole invocation, since the per-hop fetch timeout doesn't cover DNS resolution or redirect-chain totals. Route config (`MAX_SIZE`/`TIMEOUT`/headers) moved into a colocated `config.ts`.

**Cap-tuning signal.** The success path now logs the fetched size (`Logger.info('[api:metadata-proxy] Resource fetched', { byteLength, contentType, host, maxSize })`, **not** sent to Sentry), so the full size distribution — not just the over-cap tail from the existing Sentry warnings — can be queried to settle the long-term cap.

**Input hardening.** `getProxiedUri` now returns a malformed (unparseable, non-empty) URI unchanged instead of throwing, so a single bad on-chain `logoURI` can't crash a `ProxiedImage` caller rendered outside an error boundary. `MAX_SIZE`/`TIMEOUT` env overrides are validated as positive integers and fall back to their defaults (with a logged warning) on a malformed value — a value like `"abc"` previously parsed to `NaN` and silently disabled the size cap and per-hop timeout, failing open on the very guards they configure.

_(Builds on the earlier commits in this PR that report both `413` paths and the network-failure path as Sentry `warning`s.)_

## Type of change

-   [x] New feature

## Screenshots
Screenshot 2026-06-04 at 16 00 30

Screenshot 2026-06-04 at 16 01 07

Screenshot 2026-06-04 at 16 02 42

Screenshot 2026-06-09 at 13 58 16

Screenshot 2026-06-09 at 13 58 07

## Testing

- `vitest --project specs run app/api/metadata/proxy/__tests__` → **83 passed** (SSRF/private-IP matrix, redirects, size-cap paths incl. Sentry assertions, env-override validation, content-type routing, the endpoint incl. the `502`-on-unreachable-upstream path, and a real-undici streaming integration spec).
- Failure-reason layer unit-tested — `lib/__tests__/imageFailure.spec.ts` (status→reason map, probe caching, unreadable-status fallback) and `model/__tests__/useImageFailureReason.spec.ts` (on-error probe, `idle`/`probing`/`resolved` status, reset on src change, no probe off-origin). `getProxiedUri` malformed-input passthrough covered in `metadata/__tests__/utils.spec.ts`; the scheme guard in `shared/lib/__tests__/url.spec.ts` and `shared/ui/__tests__/external-link.spec.tsx`.
- Client states covered by Storybook stories with `play` assertions — `ProxiedImage.stories.tsx` (loaded, with-link, generic fallback, oversize-413 reason, reason-probe-pending, fallback-with-link, loading skeleton, no-image), over the shared `ImageWithFallback` story.
- Changed files pass type-check, ESLint, and Prettier (enforced by the pre-commit hook); `AccountHeader.spec.tsx` and `MetaplexNFTHeader.spec.tsx` green.
- `openspec validate --all --strict` → valid (`add-metadata-proxy` + `harden-external-link-schemes`).

### Pages checked

Each page renders an image that flows through the proxy. (Boxes unchecked pending the preview deployment — it needs Solana Foundation team authorization before the preview host is live.)

- [x] [USDC (GitHub PNG)](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU?cluster=devnet)
- [x] [Test program 1 (Pinata + redirect)](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/AXcxp15oz1L4YYtqZo6Qt6EkUj1jtLR6wXYqaJvn4oye?cluster=devnet)
- [x] [Test NFT 1 (`https://famousfoxes.com/hd/3104.png`)](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/5YqBjkPxxihGCB5zXbVQtH8yX1svcSxrGUJoKeYR56dV)
- [x] [Large img 1 (~12 MB)](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/JEKNVWWk5oPzAawWu1BKDS4s11PMvnQ9scDCvRa7eSuf) — exceeds the 4 MB cap, so it should render the "View original" fallback.
- [x] [Large img 2 (~4.2 MB)](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/7DG5A54uJqhD9pqUuUk9DJhhYPm8SXyyqhUEpvwBkLLb) — just over the 4 MB cap, so it should render the "View original" fallback.
- [x] [NFT collection](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/6LGkUr3Fyy7weyJVhpdHnEXN7kUn5awVpE5zGrJVTDNE)
- [x] [Random token 1](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/5QmbJw7mM6tcCdXVy8ftc2bu8izded7Etc57TMA2pump)
- [x] [Original 522](https://explorer-git-fork-hoodieshq-metadata-p-6eca89-solana-foundation.vercel.app/address/9ECGDCU4SZFJwz6hrmUta1vTXri1Bk9dHaDqCtgXejVU) 

## Related Issues

Closes https://linear.app/solana-fndn/issue/HOO-533/enable-the-metadata-service-and-test-it-out

## Checklist

## Additional Notes

The 4 MB default sits just below Vercel's ~4.5 MB buffered-response limit ([`FUNCTION_RESPONSE_PAYLOAD_TOO_LARGE`](https://vercel.com/docs/errors/FUNCTION_RESPONSE_PAYLOAD_TOO_LARGE)) — a thin ~0.5 MB margin — so our own cap, not the platform, is the binding constraint, and oversize content degrades to the client fallback rather than a platform-layer rejection. The margin is deliberately small: the success-path size log feeds the decision on the long-term cap (and a possible split between a small JSON cap and a larger image cap), and is the signal to trim back if payloads crowd the limit.


app/components/account/nftoken/NFTokenAccountSection.tsx
```diff
@@ -90,14 +90,18 @@ export const NftokenImage = ({ url, size }: { url: string | undefined; size: num
...
+                    onError={() => {
+                        setIsLoading(false);
+                    }}
                     onLoad={() => {
                         setIsLoading(false);
                     }}
-                    src={url ? getProxiedUri(url) : url}
+                    uri={url}
                     width={size}
                 />
```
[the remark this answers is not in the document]
askov: Good catch — fixed in eef03529: added `showOriginalLink` to `NftokenImage`'s `ProxiedImage`, matching the Metaplex path. The skeleton-clearing `onLoad`/`onError` still fire (spread onto the inner image), so a failed/oversize NFToken NFT now degrades to the Solana-logo placeholder **plus** the "View original" link.

app/components/shared/ui/image/ExternalResourceLink.tsx
```diff
@@ -0,0 +1,49 @@
...
+                'e-inline-flex e-items-center e-gap-1.5 e-text-dk-xs e-uppercase e-tracking-wider e-text-dk-gray-600',
+                className,
+            )}
+        >
+            <a
+                className="e-text-dk-gray-600 hover:e-text-dk-white"
+                href={href}
+                rel="noopener noreferrer"
+                target="_blank"
+            >
```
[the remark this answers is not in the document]
askov: That is **not** introduced in this PR. But I agree that this is important. Added minimal updates, a todo.

app/features/metadata/ui/ProxiedImage.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * (a `413`) or the upstream is unreachable. A browser `<img>` can't read the HTTP
+ * status, so one fallback covers every cause.
+ *
+ * This is the only metadata-aware layer; it composes the proxy-agnostic
+ * primitives from `shared/ui/image`.
+ */
+export function ProxiedImage({
+    uri,
+    fallback,
+    showOriginalLink = false,
```
rogaldh, line 65: According to the `ProxiedImageProps`, `uri` is optional, but nothing prevents passing `showOriginalLink`.
Could we use discriminated union here, for example?

app/components/shared/ui/image/__stories__/ImageWithFallback.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import solanaLogo from '@img/logos-solana/low-contrast-solana-logo.svg';
```
rogaldh, line 1: Through the work ofd replacing the Dashkit I've started to locate stories by `__stories__` path. Could you, please, update these stories accordingly?

openspec/changes/harden-external-link-schemes/tasks.md
```diff
@@ -0,0 +1,29 @@
...
+- [ ] `app/features/feature-gate/ui/*` and `app/components/account/FeatureAccountSection.tsx` — `simd_link`
+- [ ] `app/components/DeveloperResources.tsx` and `app/entities/cluster/ui/ExplorerLink.tsx` — `link`
+
+## 3. De-duplicate the server-side copy
+
+- [ ] `app/api/metadata/proxy/route.ts` — replace the local `parseUrl` + `isHTTPProtocol` with the shared `parseUrl` / `SAFE_EXTERNAL_PROTOCOLS`, keeping a thin wrapper for the invalid-URL `Logger.error`. No behaviour change (the route already rejects non-http(s)).
+
+## 4. Verify
+
+- [ ] `pnpm typecheck`, targeted tests asserting unsafe schemes render no anchor, and `pnpm openspec:validate` pass.
```
rogaldh, line 37: Let's mark the ones that were done.

askov: Nothing from here is done yet, but I added a section with what's done (so it's clear)

app/components/shared/ui/image/__stories__/ImageWithFallback.stories.tsx
askov: Sure, done

app/features/metadata/ui/ProxiedImage.tsx
askov: I agree - this isn't elegant. Introducing a union would complicate the call site, storybook, or require constructing object props, so this ugly choice is deliberate: it's simple and nothing breaks if you pass nonsense values independently. I'm happy to change it if you have suggestions that keep the interface clean and standard without complicating the call site.

rogaldh (approved):
LGTM

Woody4618 (approved):

PR #1101, opened 2026-06-19
chore(vitest): add vite deps into catalogue

rogaldh:
## Description

Moves the lockstep-released vitest core packages into a dedicated pnpm `vitest` catalog and aligns them on a single version, so they can no longer drift apart.

-   New `vitest` catalog in `pnpm-workspace.yaml` pinning `vitest`, `@vitest/browser`, `@vitest/coverage-v8` to `3.2.6`
-   `package.json` references all three via `catalog:vitest`
-   Lands on `3.2.6` because `3.2.5` is a corrupted release (`vite-node@3.2.5` was never published, so the family can't resolve); `3.2.6` reverts `vite-node` to `3.2.4` internally
-   Independently-versioned `@vitejs/plugin-react`, `@vitest/eslint-plugin`, and `vite-plugin-node-polyfills` are intentionally left out — they don't share the release line

## Type of change

-   [ ] Bug fix
-   [ ] New feature
-   [ ] Protocol integration
-   [ ] Documentation update
-   [x] Other (please describe):

Dependency/tooling maintenance (no runtime code change).

## Screenshots

N/A — no UI changes.

## Testing

-   `pnpm test` (specs): 234 files, 2763 passed
-   `pnpm test:sb` (storybook browser, chromium): 279 files, 989 passed
-   `pnpm coverage --run` (v8 provider): 513 files, 3752 passed
-   Full `Build-And-Test` CI job run locally via `act` on both node `22.x` and `24.x` matrix legs — `pnpm install --frozen-lockfile` resolves the catalog cleanly and all steps pass

## Related Issues

[HOO-706](https://linear.app/solana-fndn/issue/HOO-706)

Supersedes the dependabot bump that raised only `@vitest/browser` to the corrupted `3.2.5`.

## Checklist

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -8,6 +8,10 @@ catalogs:
     '@storybook/react': 10.2.6
     eslint-plugin-storybook: 10.2.6
     storybook: 10.2.6
+  vitest:
+    '@vitest/browser': 3.2.6
+    '@vitest/coverage-v8': 3.2.6
+    vitest: 3.2.6
 autoInstallPeers: true
 minimumReleaseAge: 20160
 minimumReleaseAgeExclude:
--- a/package.json
+++ b/package.json
@@ -159,8 +159,8 @@
         "@types/react-dom": "19.2.3",
         "@types/testing-library__jest-dom": "5.14.5",
         "@vitejs/plugin-react": "4.3.4",
-        "@vitest/browser": "3.2.4",
-        "@vitest/coverage-v8": "3.2.4",
+        "@vitest/browser": "catalog:vitest",
+        "@vitest/coverage-v8": "catalog:vitest",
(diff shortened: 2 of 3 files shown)


askov (approved):

pnpm-workspace.yaml
```diff
@@ -8,6 +8,10 @@ catalogs:
     '@storybook/react': 10.2.6
     eslint-plugin-storybook: 10.2.6
     storybook: 10.2.6
+  vitest:
```
Woody4618, line 11: This is a pnpm only feature right? Would not not break npm install? is that worth doing over just updating the dependencies in the package json?

rogaldh: This project uses pnpm. It is forbidden to use npm, as we configure pnpm specifically to prevent supply-chain attacks

Woody4618 (approved):

PR #1109, opened 2026-06-24
Fix: removed css file for tx page

C0mberry:
## Description
- removing css file and moving styles to tailwind
- adding complete state for download btn
- fixing styles for address wrapping in accounts card

## Type of change
-   [x] Bug fix

## Screenshots
Screenshot 2026-06-24 at 17 22 50

## Testing
1. open tx page
2. Scroll to the accounts block

## Related Issues
[HOO-713](https://linear.app/solana-fndn/issue/HOO-713/consider-adding-a-template-for-issues)

## Checklist
-   [x] Screenshots included — required for protocol screens, recommended for other UI changes

### DIFF
--- a/app/components/shared/RawDataField.tsx
+++ b/app/components/shared/RawDataField.tsx
@@ -7,7 +7,7 @@ import { Tabs, TabsContent, TabsList, TabsTrigger } from '@components/shared/ui/
 import React, { useEffect, useMemo, useState } from 'react';
 import { Check, ChevronDown, Copy, Download } from 'react-feather';
 
-import { DownloadDropdown, type DownloadState } from '@/app/shared/components/DownloadDropdown';
+import { DownloadDropdown, DownloadState } from '@/app/shared/components/DownloadDropdown';
 import { type ByteArray, toBase64, toHex } from '@/app/shared/lib/bytes';
 import { useCopyToClipboard } from '@/app/shared/lib/useCopyToClipboard';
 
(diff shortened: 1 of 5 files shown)


rogaldh (approved):
LGTM

Woody4618 (dismissed):

app/components/shared/RawDataField.tsx
```diff
@@ -33,6 +33,14 @@ export function RawDataField({ data, loading, filename }: RawDataFieldProps) {
     const [tab, setTab] = useState<'hex' | 'base64'>('hex');
     const [expanded, setExpanded] = useState(false);
     const [copyState, copy] = useCopyToClipboard();
+    const [downloadState, setDownloadState] = useState<DownloadState>('idle');
+
+    useEffect(() => {
+        if (downloadState === 'downloaded') {
```
Woody4618: Why are these string and not the DownloadState type?

app/tx/[signature]/page-client.tsx
```diff
@@ -85,7 +83,7 @@ export function TransactionDetailsPageClient({ params: { signature: raw } }: Pro
     }
 
     return (
-        <div className="transaction-page mx-auto flex max-w-5xl flex-col space-y-9 px-4 pt-3 lg:space-y-12 lg:px-6 lg:pt-5">
+        <div className="mx-auto flex max-w-5xl flex-col space-y-9 px-4 pt-3 selection:bg-[#13d89b40] selection:text-inherit lg:space-y-12 lg:px-6 lg:pt-5">
```
Comment (Woody4618) [inline app/tx/[signature]/page-client.tsx:86]: Is this a custom color on purpose?

app/components/shared/RawDataField.tsx
C0mberry: let me replace it with enum. it was a type from the beginning

app/tx/[signature]/page-client.tsx
Comment (C0mberry) [reply app/tx/[signature]/page-client.tsx:86]: i could use bg-accent/25 instead of a hex value, but the opacity modifier isn't working for the selection. Without that '/25' modifier, the color will be strong 
Screenshot 2026-06-25 at 14 52 34

Comment (Woody4618) [reply app/tx/[signature]/page-client.tsx:86]: Yeah maybe i misunderstand, but we should not hotfix this in a sepcific place and copy paste it everywhere but instead find a good solution we can apply and reuse everywhere where we have selections, no?

Comment (rogaldh) [reply app/tx/[signature]/page-client.tsx:86]: That really depends. Sometimes it makes sense to wrap up a solution, sometimes it does not. There might be different reasons: complexity, attempt to have a self-sustained solution. That really depends.
Here, it looks appropriate as the context is localised. But it is really worth unifying the styles, and it should be done, but within a separate scope.
I'd suggest accepting this and reworking the whole thing later.

Woody4618 (approved):

PR #1116, opened 2026-06-26
feat: Program Metadata Program — decode instructions, accounts & events; use PMP IDLs everywhere

ChewingGlass:
## Summary

Makes the explorer understand the **Program Metadata Program** (PMP, `ProgM6JCCvbYkfKqJYHePx4xxSUSqJp7rh8Lyv7nk7S`) and actually use PMP‑published IDLs everywhere it parses program data — instructions, accounts, and events — not just on the transaction page.

### 1. Decode the PMP's own instructions
Instructions to the Program Metadata Program (`Initialize` / `Allocate` / `SetData` / `SetAuthority` / `SetImmutable` / `Trim` / `Close` / `Extend` / `Write`) rendered as raw hex / "Unknown Instruction". A new `ProgramMetadataDetailsCard` decodes them with the `@solana-program/program-metadata` Codama decoders (deterministic, no IDL fetch), wired into both the transaction page and the inspector.

### 2. Use PMP‑published IDLs in the inspector
The transaction page already preferred a program's PMP IDL when decoding instructions; the inspector was Anchor‑only. The inspector now resolves the PMP IDL and renders the `ProgramMetadataIdlInstructionDetailsCard`, so a program that publishes its interface via the PMP decodes there too.

### 3. Decode account data via PMP‑published IDLs
`AnchorAccountCard` only used the legacy on‑chain Anchor IDL, so accounts owned by programs that publish only a PMP IDL showed nothing and the "Anchor Data" tab was hidden. It now falls back to an Anchor‑format IDL fetched from the PMP (reusing the existing Borsh decode + row rendering), and the tab is shown for such programs.

### 4. Decode Anchor‑format PMP IDLs that Codama can't convert
The PMP IDL instruction card decoded only through Codama (`parseInstruction` / `rootNodeFromAnchor`). Real Anchor IDLs can fail that conversion — e.g. an instruction with an unnamed arg makes `rootNodeFromAnchor` throw `Argument name [id] is missing…` — leaving the instruction "Unknown" even with the IDL present. It now falls back to building an Anchor `Program` and decoding with Anchor's (more lenient) `BorshInstructionCoder` before giving up.

### 5. Decode Anchor events emitted via `Program log:` (and fix index matching)
Two pre‑existing bugs in `extractEventsFromLogs`:
- It only read `Program data:` (`sol_log_data`) lines, so programs that emit events as base64 via `Program log:` (`msg!`) had their events ignored. It now also collects base64‑shaped `Program log:` payloads (plain‑text logs fail the base64 check; the event decoder rejects any straggler).
- It matched events to instructions by counting `invoke` log lines, but instructions that emit none — the ed25519/secp256k1 precompiles — shifted the index, mis‑attributing or dropping events. Given the ordered top‑level program ids, invocations are now matched to their instruction by program id. `AnchorDetailsCard` passes those ids, and the PMP Anchor fallback forwards the signature so events can be read from the tx logs.

### 6. Make IDL resolution resilient to transient fetch errors
Resolving a large IDL through the server runtime occasionally aborts the response body with `ERR_STREAM_PREMATURE_CLOSE`. The `/api/idl-latest` route treated that as fatal (paged Sentry, non‑retryable 502). It now classifies these connection‑level errors as retryable and retries with a fresh client.

## Motivation

Programs increasingly publish their IDL through the Program Metadata Program rather than a legacy Anchor IDL account. Without these changes the explorer shows raw hex for PMP instructions, can't decode the instructions/accounts/events of PMP‑only programs, and (for IDLs Codama can't convert or events logged via `Program log:`) shows "Unknown" even when the IDL is present.

## Testing

`pnpm test:ci`, `pnpm typecheck`, and `pnpm lint` all green. New tests cover: PMP instruction decoding (`ProgramMetadataDetailsCard`), the Codama→Anchor instruction‑decode fallback and event‑signature forwarding (`ProgramMetadataIdlInstructionDetailsCard`), and the IDL route's retry/transient‑error handling.

> PMP IDL resolution is gated behind the existing `NEXT_PUBLIC_PMP_IDL_ENABLED` flag (same as the transaction page's current PMP path), so items 2–5 require it enabled.

## Screenshots

> _To be added from a local instance before requesting review_: a decoded Program Metadata Program instruction; a PMP‑only program's instruction decoded in the inspector; a PMP‑only program account's fields under the Anchor Data tab; and a decoded Anchor event.


KSH9NT (approved):

Woody4618:
Thanks for this. Was included in https://github.com/solana-foundation/explorer/pull/1119/commits

PR #1125, opened 2026-07-01
fix(token-2022): decode core token instructions in inspector

ChewingGlass:
## Problem

In the transaction inspector, Token-2022 transfers (and other core token instructions) render as raw hex with no decoded fields — no amount, no source/destination, etc.

Example: https://explorer.solana.com/tx/inspector?squadsTx=949B6HVx1T87TAwtia8guXguQhUCoSAKnzA7gopfqydL

The instruction data `0c ae 5e 27 23 00 00 00 00 06` is a valid `TransferChecked` (discriminator `0x0c`=12, amount `589782702`, decimals `6`), yet the inspector shows only the raw bytes.

## Root cause

The Token-2022 byte-path parser (`parseToken2022Instruction`) only handled the Token-2022–specific extension instructions (metadata / group / pointers). The core instructions that Token-2022 inherits from SPL Token — `Transfer`, `TransferChecked`, `CloseAccount`, `InitializeAccount`, `SyncNative` — hit the `default` case and returned `undefined`, so the dispatcher wrapped them as unparsed and the inspector fell back to `UnknownDetailsCard` (raw hex).

This only affected the **byte path** (inspector). The tx page's RPC pre-parsed path already decoded these via the shared validator map, which is why it only showed up in the inspector.

## Fix

Token-2022 shares these instruction layouts byte-for-byte with SPL Token, so decode them the same way the SPL Token slice does — mirroring `parseTokenInstruction` in the `decode-instruction-token` slice, using `@solana-program/token-2022`'s parse functions. The output `{ type, info }` shapes match the existing `IX_STRUCTS`/`IX_TITLES` validators in `app/components/instruction/token/types.ts`, which `TokenDetailsCard` already renders (the `spl-token-2022` case was already wired up in the inspector's `InstructionsSection`).

The duplication with the SPL Token slice is intentional and follows the existing pattern in this file (see the `parseToken2022RpcInstruction` comment) — slices stay independent with no feature-to-feature imports.

## Test

Adds a byte-vs-RPC parity contract test for Token-2022 `TransferChecked` alongside the existing SPL Token one. `pnpm test` passes.

New:
Screenshot 2026-07-01 at 9 38 38 AM

Old:
Screenshot 2026-07-01 at 9 38 58 AM

### DIFF
--- a/app/features/decode-instruction-token-2022/lib/token-2022-parser.ts
+++ b/app/features/decode-instruction-token-2022/lib/token-2022-parser.ts
@@ -4,7 +4,9 @@ import { unwrapOption } from '@solana/kit';
 import { type ParsedInstruction, PublicKey } from '@solana/web3.js';
 import {
     identifyToken2022Instruction,
+    parseCloseAccountInstruction,
     parseEmitTokenMetadataInstruction,
+    parseInitializeAccountInstruction,
     parseInitializeGroupMemberPointerInstruction,
     parseInitializeGroupPointerInstruction,
     parseInitializeMetadataPointerInstruction,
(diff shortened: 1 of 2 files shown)


rogaldh (approved):
LGTM

Woody4618 (approved):

PR #1126, opened 2026-07-01
feat: Render permissioned burn token extension

gitteri:
Adds parsing and display for the token-2022 permissioned burn mint extension. The RPC now returns it as `permissionedBurnConfig` (via the agave account-decoder change in anza-xyz/agave#13504), so the extensions page can surface the burn authority.

Adds the `permissionedBurnConfig` extension type + validator, a "Permissioned Burn" label with description and docs link, and a row rendering the burn authority.

Example mint that carries it: `6rS8E5CQHLHmLenNNevUdJ9APzKNairr1kfWAM3EhDX3`

### DIFF
--- a/app/components/account/TokenAccountSection.tsx
+++ b/app/components/account/TokenAccountSection.tsx
@@ -30,6 +30,7 @@ import {
     MintCloseAuthority,
     PausableConfig,
     PermanentDelegate,
+    PermissionedBurnConfig,
     ScaledUiAmountConfig,
     TokenExtension,
     TokenExtensionType,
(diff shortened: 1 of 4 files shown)


gitteri:
Thanks @greptile-apps, both addressed:

- **P1 (empty header):** Permissioned burn has a single field (`authority`), so I switched it to the single-row, no-header pattern used by `permanentDelegate` (renders one "Permissioned Burn Authority" row, or nothing when `authority` is null) and moved it out of the header-section group in the sort order. No more floating header.
- **P2 (null case untested):** Added a test asserting nothing renders when `authority` is null.

Woody4618 (approved):

Woody4618:
thanks for this!

PR #1132, opened 2026-07-03
fix: custom url removal from requests

C0mberry:
## Description

Keeps the custom-cluster RPC URL out of requests to the Explorer's own server. Custom-cluster RPC traffic is resolved directly by the browser, so our server never needs the value.

- **Stop sending `customUrl` on rpc requests** — `pickClusterParams` no longer preserves `customUrl` when the resulting cluster isn't `custom`, so navigation on mainnet/devnet/testnet stops carrying it (`adc1fd30f`). The `additionalParams` path decides from the *merged* cluster, so switching custom → devnet drops the endpoint rather than inheriting it. **This is the one user-visible change in the PR.**
- **`useDasImage` stops appending `customUrl` to `/api/token-image`.** This fired on *every cluster for every user*, not only on Custom: `rememberedCustomUrlAtom` defaults to `http://localhost:8899` and `useCluster().customUrl` returns that remembered value whatever the active cluster, so the `if (customUrl)` guard was always true. Anyone with a saved keyed endpoint shipped it to us on every token-image load while browsing mainnet.
- **The server can no longer read a client-supplied URL at all.** `serverClusterUrl` drops its `customUrl` parameter and takes `ServerCluster = Exclude<Cluster, Cluster.Custom>`, so resolving one is a compile error rather than a per-route runtime guard. The narrowing propagates to the helpers that carry a cluster toward an outbound RPC call: `clusterFromGenesisHash`, `findTransactionCluster`, `getSignatureStatus`, `getTx`, `resolveSearchTokens`. The token-image route keeps its `Custom → 400` guard as defense-in-depth.

Previously this invariant held only by convention — every `serverClusterUrl` caller remembered to pass `''` — plus that single runtime 400.

The server-side and transport changes are behaviour-neutral: the token-image route reads only `cluster` and rejects Custom with 400 either way, so their only observable effect is a shorter request URL — expect a one-time `/api/token-image` CDN cache warm after deploy, then a better hit rate (each distinct saved `customUrl` previously fragmented the cache key). The `pickClusterParams` change above *is* a behaviour change, and is the fix the issue asks for.

Design rationale, rejected alternatives, and the remaining scope are in `openspec/changes/keep-custom-url-off-our-server/`.

## Type of change

-   [x] Bug fix

## Testing

- [Token address page (mainnet) — Network tab shows `/api/token-image/…?cluster=mainnet-beta` with no `customUrl`](https://explorer-git-fork-hoodieshq-fix-hoo-79-bb4038-solana-foundation.vercel.app/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v)
- [Same page, navigate between tabs — `customUrl` is not added to the URL on a non-custom cluster](https://explorer-git-fork-hoodieshq-fix-hoo-79-bb4038-solana-foundation.vercel.app/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v/transfers)
- [Custom cluster — endpoint still resolves client-side and the page loads; token images still don't render (unchanged)](https://explorer-git-fork-hoodieshq-fix-hoo-79-bb4038-solana-foundation.vercel.app/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v?cluster=custom)

## Related Issues

Closes [HOO-792](https://linear.app/solana-fndn/issue/HOO-792/requests-are-sending-customurl-even-there-is-no-customurl)

## Checklist

## Additional Notes

**Still out of scope:** when the Custom cluster is active, the endpoint remains in the page URL (`?cluster=custom&customUrl=…`). A keyed endpoint therefore still reaches our access logs, Google Analytics (`page_location` is the full `href`, and there is no `page_location` override or query stripping) and Sentry (`request.url`, with `sampleRate: 1` and no `beforeSend`) on those page loads. Two follow-ups are recorded under "Known residual" in the OpenSpec proposal: scrub query strings at the analytics/Sentry sinks first, then move the endpoint to `localStorage`.

`build:info` reports one first-load change: `/block/[slot]/programs` 210 kB → 220 kB. This PR doesn't touch that route and the client diff only removes code, so this reads as bucket-rounding jitter rather than a regression.


app/utils/__tests__/url.spec.ts
```diff
@@ -26,10 +26,10 @@ describe('pickClusterParams', () => {
             expect(result).toBe('/address/abc123?cluster=devnet');
         });
 
-        it('should preserve customUrl param from current search', () => {
+        it('should not preserve customUrl param when cluster is not custom', () => {
             const currentParams = new URLSearchParams('customUrl=http://localhost:8899');
```
rogaldh, line 30: I think it is worth adding a case when we pass cluster=custom + customUrl

app/utils/url.ts
```diff
@@ -68,6 +73,10 @@ export function pickClusterParams(
             }
             params.set(key, value); // Override current with additional
         });
+        // Remove customUrl when not on custom cluster
+        if (params.get('cluster') !== CUSTOM_MONIKER) {
+            params.delete('customUrl');
```
rogaldh, line 84: Checked quickly, it looks like this branch was not tested. Could you please validate that?

app/utils/url.ts
```diff
@@ -68,6 +73,10 @@ export function pickClusterParams(
             }
             params.set(key, value); // Override current with additional
         });
+        // Remove customUrl when not on custom cluster
```
rogaldh: Could you, please, improve the comment to explain WHY, not WHAT

openspec/changes/keep-custom-url-off-our-server/design.md
```diff
@@ -0,0 +1,66 @@
...
+- Make it impossible for server code to resolve a client-supplied RPC URL — at the type level, not runtime-only.
+- Stop the one client fetch that forwards `customUrl` to our own API for nothing.
+- Remove the dead `customUrl` parameters and types that imply the server consumes the value.
+
+**Non-Goals:**
+
+- Changing where the custom endpoint lives. It stays in the page URL. See `proposal.md`, follow-up 2.
+- Scrubbing query strings at the analytics / Sentry / logging sinks. See `proposal.md`, follow-up 1.
+- Touching link generation, `pickClusterParams`, the cluster switcher, `isCustomUrlAllowed`, or the cross-cluster discovery probes.
+- Making custom-cluster token images render. They do not today; that is a separate feature.
```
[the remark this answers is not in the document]
askov: Fixed

app/utils/url.ts
askov: Replaced both WHAT comments

app/utils/url.ts
askov: Added four cases

app/utils/__tests__/url.spec.ts
askov: Covered

openspec/changes/keep-custom-url-off-our-server/specs/custom-cluster-url/spec.md
```diff
@@ -0,0 +1,59 @@
...
+- **THEN** the request to `/api/token-image` SHALL NOT include a `customUrl` parameter
+- **AND** the image SHALL simply not render (the Explorer server does not serve custom-cluster images)
+
+#### Scenario: A new client fetch needs the active cluster
+
+- **WHEN** client code adds a fetch to an Explorer `/api/*` route that depends on the active cluster
+- **THEN** it SHALL identify the cluster by its slug alone
+- **AND** SHALL NOT forward the custom endpoint, because no server route can consume it
+
+### Requirement: Navigation params SHALL carry customUrl only when the target cluster is Custom
```
rogaldh: That requirement is a bit incorrect.
Explorer has [a helper](https://github.com/solana-foundation/explorer/blob/310ee3f949e35c57144d325d92531cdbb675b9d8/app/entities/cluster/lib/resolve-cluster.ts#L15) which contains logic to enable customUrl for the address string or not.

So the navigation should rely on this criterion.

Also important that this should survive direct visits to the specific url, but if Marco sends a link with `cluster=custom&cusomUrl=<%malformed_url%>` to Polo, it should survive only if `devFlagEnabled=true`

askov: I fixed the spec/navigation. But the second problem seems to be a different story not related to the current task. Please, let me know if the task description is incomplete and we need to handle everything in this PR

rogaldh (approved):

PR #1166, opened 2026-07-27
feat: add verified builds for devnet

askov:
## Description

OSEC now hosts a verified-builds registry for devnet (`https://verify-devnet.osec.io`). This wires the explorer's verified-build UI to query the correct registry per cluster, so verified builds now work on **Devnet** as well as **Mainnet**.

- Added `getOsecRegistryUrl(cluster)` — returns the mainnet registry for Mainnet, the devnet registry for Devnet, and `undefined` for clusters with no registry (Testnet/Custom/SIMD-296) — plus a `supportsVerifiedBuilds(cluster)` predicate.
- `useVerifiedProgramRegistry` and `useIsProgramVerified` now select the registry URL by cluster. The full URL is the SWR cache key, so Mainnet and Devnet never share a cached entry; on unsupported clusters no request is made.
- The generated `solana-verify` command uses the correct cluster moniker: `-ud` on Devnet, `-um` on Mainnet.
- `VerifiedProgramBadge` now renders on Mainnet and Devnet; the fallback notice reads "only available on Mainnet and Devnet".

The verified-programs **search autocomplete** stays Mainnet-only: it reads a pre-built snapshot (`public/verified-programs.json`) generated by a daily GitHub Action against Mainnet. Extending it to Devnet needs a separate snapshot + CI job + a `DEVNET_RPC_URL` secret — tracked as follow-up, not in this PR.

## Type of change

-   [x] New feature

## Testing

- [Devnet program page — badge queries `verify-devnet.osec.io` (previously the "only available on Mainnet" block)](https://explorer-git-fork-hoodieshq-feat-add-v-9f882a-solana-foundation.vercel.app/address/HnbLDGALwWtMHYFYrcCwz5vHAtWXcnZKT6iKBibZvyiP?cluster=devnet)
- [Devnet verified-build tab — resolved against the devnet registry](https://explorer-git-fork-hoodieshq-feat-add-v-9f882a-solana-foundation.vercel.app/address/HnbLDGALwWtMHYFYrcCwz5vHAtWXcnZKT6iKBibZvyiP/verified-build?cluster=devnet)
- [Testnet — "only available on Mainnet and Devnet" guard](https://explorer-git-fork-hoodieshq-feat-add-v-9f882a-solana-foundation.vercel.app/address/HnbLDGALwWtMHYFYrcCwz5vHAtWXcnZKT6iKBibZvyiP?cluster=testnet)
- [Mainnet verified card (unchanged) — full card with the `-um` verify command](https://explorer-git-fork-hoodieshq-feat-add-v-9f882a-solana-foundation.vercel.app/address/BUYuxRfhCMWavaUWxhGtPP3ksKEDZxCD5gzknk3JfAya/verified-build)

## Related Issues

Closes [HOO-871](https://linear.app/solana-fndn/issue/HOO-871/add-verified-builds-for-devnet)

## Checklist

## Additional Notes

The OSEC devnet registry is **live but currently reports zero verified programs**: 218 programs have uploaded Otter Verify PDAs on-chain (`solana-verify … -ud`), but the devnet backend returns `is_verified: false` (empty `on_chain_hash`) for all of them — its build-and-compare pipeline has not populated devnet results yet. This client change is correct and will surface verified cards the moment OSEC populates the registry, with no further code change. The preview links above therefore exercise the wiring and the not-verified path; a green verified card on devnet is not yet reproducible.

Per the template note: Verified Builds bugs should be reported to disclosures@solana.org.

### DIFF
--- a/app/components/common/__stories__/VerifiedProgramBadge.stories.tsx
+++ b/app/components/common/__stories__/VerifiedProgramBadge.stories.tsx
@@ -23,12 +23,17 @@ const expectedHash = hashProgramData(programData);
 
 type OsecOutcome = 'verified' | 'unverified' | 'loading' | 'error';
 
+// Matches any OSEC registry request (mainnet verify.osec.io + devnet verify-devnet.osec.io).
+function isOsecRequest(input: RequestInfo | URL): boolean {
+    const url = typeof input === 'string' ? input : input instanceof URL ? input.toString() : input.url;
+    return url.includes('osec.io');
+}
+
 const withMockedOsec = (outcome: OsecOutcome, state?: ClusterState): Decorator =>
     function WithMockedOsec(Story) {
         const originalFetch = window.fetch;
         window.fetch = (async (input: RequestInfo | URL, init?: RequestInit) => {
-            const url = typeof input === 'string' ? input : input instanceof URL ? input.toString() : input.url;
-            if (!url.includes('osec.io')) return originalFetch(input as any, init);
+            if (!isOsecRequest(input)) return originalFetch(input as any, init);
             if (outcome === 'loading') return new Promise<Response>(() => {});
             if (outcome === 'error') throw new Error('mocked OSEC failure');
             return new Response(JSON.stringify({ is_verified: outcome === 'verified', on_chain_hash: expectedHash }), {
--- a/app/components/common/__stories__/VerifiedProgramBadge.stories.tsx
+++ b/app/components/common/__stories__/VerifiedProgramBadge.stories.tsx
@@ -23,12 +23,12 @@ const expectedHash = hashProgramData(programData);
 
 type OsecOutcome = 'verified' | 'unverified' | 'loading' | 'error';
 
-const withMockedOsec = (outcome: OsecOutcome): Decorator =>
+const withMockedOsec = (outcome: OsecOutcome, state?: ClusterState): Decorator =>
     function WithMockedOsec(Story) {
         const originalFetch = window.fetch;
         window.fetch = (async (input: RequestInfo | URL, init?: RequestInit) => {
             const url = typeof input === 'string' ? input : input instanceof URL ? input.toString() : input.url;
-            if (!url.includes('verify.osec.io')) return originalFetch(input as any, init);
+            if (!url.includes('osec.io')) return originalFetch(input as any, init);
             if (outcome === 'loading') return new Promise<Response>(() => {});
             if (outcome === 'error') throw new Error('mocked OSEC failure');
             return new Response(JSON.stringify({ is_verified: outcome === 'verified', on_chain_hash: expectedHash }), {
@@ -38,7 +38,7 @@ const withMockedOsec = (outcome: OsecOutcome): Decorator =>
         }) as typeof window.fetch;
         return (
             <SWRConfig value={{ provider: () => new Map() }}>
-                <MockClusterProvider>
+                <MockClusterProvider state={state}>
                     <Story />
                 </MockClusterProvider>
             </SWRConfig>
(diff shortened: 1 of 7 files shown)


askov:
@greptile-apps Good catch on the P1 (`app/utils/verified-builds.tsx` PDA cache crosses clusters).

Fixed in 2d81650f: the Otter Verify PDA `useSWRImmutable` key now includes the cluster RPC endpoint (`pda-${clusterUrl}-${programId}-${signers}`). Since the PDA is a per-cluster on-chain account and the cache is immutable (never revalidates), the previous cluster-agnostic key could reuse the other network's PDA and pair its repo/commit/args with the current network's `-um`/`-ud` flag. Scoping by endpoint keeps Mainnet, Devnet, and custom RPCs from colliding — consistent with how the `status-all`/`status` registry keys are already scoped by URL.

app/components/common/__stories__/VerifiedProgramBadge.stories.tsx
```diff
@@ -23,12 +23,12 @@ const expectedHash = hashProgramData(programData);
...
 type OsecOutcome = 'verified' | 'unverified' | 'loading' | 'error';
 
-const withMockedOsec = (outcome: OsecOutcome): Decorator =>
+const withMockedOsec = (outcome: OsecOutcome, state?: ClusterState): Decorator =>
     function WithMockedOsec(Story) {
         const originalFetch = window.fetch;
         window.fetch = (async (input: RequestInfo | URL, init?: RequestInit) => {
             const url = typeof input === 'string' ? input : input instanceof URL ? input.toString() : input.url;
-            if (!url.includes('verify.osec.io')) return originalFetch(input as any, init);
+            if (!url.includes('osec.io')) return originalFetch(input as any, init);
```
rogaldh: nit: I think it would be better to have a helper that contains matching logic.

app/utils/verified-builds.tsx
```diff
@@ -79,14 +96,17 @@ export function useVerifiedProgramRegistry({
...
     const {
         data: registryData,
         error: registryError,
         isLoading: isRegistryLoading,
     } = useSWRImmutable(
-        `${programId.toBase58()}`,
-        async (programId: string) => {
-            const response = await fetch(`${OSEC_REGISTRY_URL}/status-all/${programId}`);
+        // The full URL is the cache key so Mainnet and Devnet never share an entry.
+        registryUrl ? `${registryUrl}/status-all/${programId.toBase58()}` : null,
```
rogaldh: I'd align the key structure with other keys, like this, for example: https://github.com/solana-foundation/explorer/pull/1166/changes#diff-29b82d6a236af0b72ed221ed1a5c393ec83ce5d9a32256d450918807f10d2e2eR276

app/utils/verified-builds.tsx
```diff
@@ -337,15 +370,17 @@ export function buildEnrichedOsecInfo({
...
-              : 'Verify command only available on mainnet.',
+              : 'Verify command not available on this cluster.',
     };
 }
 
-function coalesceCommandFromPda(programId: PublicKey, pdaData: OtterVerifyBuildParams) {
-    let verify_command = `solana-verify verify-from-repo -um --program-id ${programId.toBase58()} ${pdaData.gitUrl}`;
+function coalesceCommandFromPda(programId: PublicKey, pdaData: OtterVerifyBuildParams, cluster: Cluster) {
+    // solana-verify targets a cluster with the URL moniker: -ud for Devnet, -um for Mainnet.
+    const clusterFlag = cluster === Cluster.Devnet ? '-ud' : '-um';
```
rogaldh, line 382: Let's maybe move that into the entity? Looks to me like domain-level knowledge, but exists within the utility

rogaldh (dismissed):

app/components/common/__stories__/VerifiedProgramBadge.stories.tsx
askov: Done in 4a0bedeb — extracted the matching logic into an `isOsecRequest(input)` helper that the decorator calls.

app/utils/verified-builds.tsx
askov: Done in 4a0bedeb — the `status-all` key is now the array form `['status-all', registryUrl, programId.toBase58()]`, matching the `is-program-verified` key structure, and the fetcher destructures `[_prefix, base, id]`.

app/utils/verified-builds.tsx
askov: I totally agree, but it adds too much noise and greptile won't be happy with moved code, so I'll have to rework many things. Let me create a follow-up refactoring task for it ([HOO-910](https://linear.app/solana-fndn/issue/HOO-910/refactor-verified-builds-to-use-fsd))

gmw37:
Thanks

Woody4618 (approved):

PR #1170, opened 2026-07-29
fix: use legacy account history RPC for custom clusters

sporicle:
## Description

Custom clusters now use the standard `getSignaturesForAddress` account-history path directly instead of first calling the RPC 2.0 method, `getTransactionsForAddress`.

The filter UI is marked unsupported immediately for custom clusters because the standard RPC method cannot apply the slot, block-time, or status filters. Known clusters keep the existing RPC 2.0 request and runtime method-not-found fallback unchanged.

History state is scoped to both the cluster identity and RPC URL. This prevents cached pagination state or an in-flight response from crossing between a known and custom cluster that happen to use the same endpoint.

## Type of change

-   [x] Bug fix
-   [ ] New feature
-   [ ] Protocol integration
-   [ ] Documentation update

## Screenshots

Not applicable: this changes RPC selection and cache invalidation without introducing a visual UI change.

## Testing

- Added provider regression coverage confirming that custom clusters:
  - call `getSignaturesForAddress` directly
  - never call `getTransactionsForAddress`
  - report transaction filters as unsupported
  - do not forward unsupported filters to the legacy RPC
- Added same-URL cluster-transition coverage confirming that:
  - cached history is cleared when the cluster identity changes
  - an in-flight response from the previous cluster cannot populate the new state
- `pnpm format:ci`
- `pnpm lint`
- `pnpm openspec:validate`
- `pnpm build`
- `pnpm typecheck`
- All package tests passed
- App suite: 304 files passed; 3,313 tests passed and 28 skipped
- `pnpm build:info`

## Related Issues

None.

## Checklist

-   [ ] Screenshots included — required for protocol screens, recommended for other UI changes

## Additional Notes

This is not a protocol integration or security-related change.

### DIFF
--- a/app/providers/accounts/__tests__/history.spec.tsx
+++ b/app/providers/accounts/__tests__/history.spec.tsx
@@ -322,6 +322,73 @@ describe('useResetAccountHistory', () => {
     });
 });
 
+describe('cluster changes', () => {
+    it('should clear cached history when the cluster changes without changing the RPC URL', async () => {
+        mockResult([sig('rpc-2', 10)], null);
+
+        const { result, rerender } = renderHook(
+            () => ({
+                fetch: useFetchAccountHistory(25, {}),
+                history: useAccountHistory(ADDRESS),
+            }),
+            { wrapper },
+        );
+
+        await act(async () => {
+            result.current.fetch(new PublicKey(ADDRESS));
+        });
+        await waitFor(() => expect(result.current.history?.data?.fetched?.[0]?.signature).toBe('rpc-2'));
+
+        mockCluster = Cluster.Custom;
+        rerender();
+
+        expect(result.current.history).toBeUndefined();
+
+        mockConnection.getSignaturesForAddress.mockResolvedValueOnce([sig('legacy', 5)]);
+        await act(async () => {
+            result.current.fetch(new PublicKey(ADDRESS));
+        });
+
+        await waitFor(() => expect(result.current.history?.data?.fetched?.[0]?.signature).toBe('legacy'));
+        expect(fetchMock).toHaveBeenCalledTimes(1);
+    });
+
+    it('should discard an in-flight response from the previous cluster at the same RPC URL', async () => {
+        const pending = deferred<ReturnType<typeof envelope>>();
+        fetchMock.mockReturnValueOnce(pending.promise);
+
+        const { result, rerender } = renderHook(
+            () => ({
+                fetch: useFetchAccountHistory(25, {}),
+                history: useAccountHistory(ADDRESS),
+            }),
+            { wrapper },
+        );
+
+        act(() => {
+            result.current.fetch(new PublicKey(ADDRESS));
+        });
+        await waitFor(() => expect(fetchMock).toHaveBeenCalledTimes(1));
+
+        mockCluster = Cluster.Custom;
+        rerender();
+
+        mockConnection.getSignaturesForAddress.mockResolvedValueOnce([sig('legacy', 5)]);
+        await act(async () => {
+            result.current.fetch(new PublicKey(ADDRESS));
+        });
+        await waitFor(() => expect(result.current.history?.data?.fetched?.[0]?.signature).toBe('legacy'));
+
+        await act(async () => {
+            pending.resolve(envelope([sig('stale', 1)], null));
+            await pending.promise;
+        });
+
+        expect(result.current.history?.data?.fetched).toHaveLength(1);
+        expect(result.current.history?.data?.fetched[0].signature).toBe('legacy');
+    });
+});
+
 describe('getSignaturesForAddress fallback', () => {
     it('should use getSignaturesForAddress directly on a custom cluster', async () => {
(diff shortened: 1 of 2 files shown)


app/providers/accounts/history.tsx
```diff
@@ -154,21 +154,29 @@ export const MethodSupportContext: React.Context<MethodSupport | undefined> = Re
 
 type HistoryProviderProps = { children: React.ReactNode };
 export function HistoryProvider({ children }: HistoryProviderProps) {
-    const { url } = useCluster();
+    const { cluster, url } = useCluster();
+
+    // Two cluster identities can share an RPC URL while selecting different history
+    // methods. Remount the cache boundary so neither cached nor in-flight results cross over.
+    return (
+        <ClusterHistoryProvider key={`${cluster}:${url}`} cluster={cluster} url={url}>
```
askov, line 162: **thought**: the keyed remount unmounts the whole page subtree:

Keying `ClusterHistoryProvider` on `${cluster}:${url}` remounts every child below it, not just the provider. That is `RewardsProvider` plus the entire address or tx page (`index.tsx:202`, `app/address/[address]/layout.tsx:212`) — all local state lost, all mount effects re-run. It fires on every cluster and URL change, not only the same-URL case this PR targets. I confirmed this with a local probe.

The cache fix itself is correct, the same-URL leak is real.

A cheaper fix gives the same isolation: make the reducer identity `${cluster}:${url}` instead of `url`, and pass it as the `url` field on every dispatch. The existing guard at `cache.tsx:87` then rejects stale cross-cluster actions, including in-flight ones. It needs a `cacheKey` parameter next to `url`, since `url` stays the real RPC target.

If you keep the remount, please say so in the PR body — "remount the cache boundary" reads as if only the provider remounts.

@greptile-apps wdyt?

app/providers/accounts/history.tsx
```diff
@@ -378,21 +386,29 @@ async function fetchAccountHistory(
...
-            paginationToken: options.paginationToken,
-        });
-        history = {
-            fetched: result.data,
-            // A null/absent paginationToken is the canonical end-of-stream signal; a
-            // short page is not, since the server may still hand back a token for more.
-            foundOldest: !result.paginationToken,
-            paginationToken: result.paginationToken,
-        };
+        if (cluster === Cluster.Custom) {
```
askov, line 389: **question/issue**: this decides filter support from the cluster enum instead of asking the endpoint. `fetchAccountHistory` short-circuits on `cluster === Cluster.Custom`, so `getTransactionsForAddress` is never sent, and `supported` is forced `false`. A custom endpoint that implements the method loses filtering.

#1024 built this deliberately the other way: "Detection is behavioral ... no provider allowlist is hardcoded."

The tooltip is now wrong for those users. `HistoryFilterBar.tsx:263` tells someone on a Helius URL that filtering "requires a Triton- or Helius-compatible RPC endpoint". This repo's own fixtures use a Helius URL as the canonical custom RPC (`Navbar.stories.tsx:34`).

What failure does this fix? No issue is linked, and the `-32601` fallback already handles unsupported endpoints in one request. If the problem is a non-standard rejection (HTTP 404/405, a proxy error page), widen the fallback instead of skipping the probe:

```ts
if (isMethodNotFound(error) || cluster === Cluster.Custom) {
    onMethodNotFound?.();
```
That fixes broken endpoints and keeps filtering for capable ones. @greptile-apps wdyt?

app/providers/accounts/__tests__/history.spec.tsx
```diff
@@ -319,7 +322,98 @@ describe('useResetAccountHistory', () => {
...
+
+        await act(async () => {
+            result.current.fetch(new PublicKey(ADDRESS));
+        });
+
+        await waitFor(() => expect(result.current.history?.data?.fetched?.[0]?.signature).toBe('legacy'));
+        expect(fetchMock).not.toHaveBeenCalled();
+        expect(mockConnection.getSignaturesForAddress).toHaveBeenCalledWith(expect.any(PublicKey), { limit: 25 });
+    });
+
```
rogaldh, line 416: Fallback was added, and it looks like we need a test to ensure that fallback resolves to the `FetchFailed` state in case of an error. Smth like that:

```suggestion
it('should mark history as failed when getSignaturesForAddress rejects on a custom cluster', async () => {
    mockCluster = Cluster.Custom;
    mockConnection.getSignaturesForAddress.mockRejectedValue(new Error('connection refused'));
    // drive fetch, then:
    await waitFor(() => expect(result.current.history?.status).toBe(FetchStatus.FetchFailed));
    expect(fetchMock).not.toHaveBeenCalled(); // no fallback retry of the Triton method
});
```

rogaldh: @greptile-apps thoughts?

sporicle:
Understood, dug into it more - the problem is that the RPC I am testing is returning the non-standard -32700, causing it to not enter the fallback flow. 

I will change this on our end. Thanks for the review!

PR #1171, opened 2026-07-29
refactor(associated-token): normalise parser, unify duplicated cards

askov:
## Description

Refactors the Associated Token instruction card. The starting point was a parser that returned `info: unknown` in three incompatible shapes, and everything downstream was a workaround for it.

**The root cause.** `parseAssociatedTokenInstruction` (byte path, inspector) emitted `@solana-program/token`'s kit account names, `parseAssociatedTokenRpcInstruction` (tx page) emitted the RPC names, and `create` was passed through with no validation:

| Instruction | kit name | RPC name | Label |
| -- | -- | -- | -- |
| create / createIdempotent | `payer` / `ata` / `owner` | `source` / `account` / `wallet` | Source / Account / Wallet |
| recoverNested | `nestedAssociatedAccountAddress` | `nestedSource` | Nested Source |
| recoverNested | `ownerAssociatedAccountAddress` | `nestedOwner` | Nested Owner |

The parser's own doc comment described this as a "transitional period".

**What it forced.** Because the two paths disagreed, the inspector gave up on `info` and read `raw.keys[0..5]` by hardcoded position (Mint/Wallet at indexes 3/2 to match display order). That required a complete duplicate four-file card tree, which required injecting the card shell, and both card trees then re-decoded what the dispatcher had already decoded.

**This PR.** Both decode paths now normalise onto one canonical named shape, run through the same superstruct validator — which is what guarantees they agree — and the eight card files collapse into one presentational card driven by a `(label, field)` table. The inspector uses that same card, injecting only what genuinely differs: the card shell, the address renderer, and whether the program row shows.

### Also fixed along the way

- **`create` is now validated.** It previously rendered straight off unvalidated RPC `info`.
- **Malformed payloads no longer throw.** When the parser rejects a payload, `createInstructionParserDispatcher` falls back to RPC's raw view — a familiar `type` carrying un-coerced base58 strings. The card verifies the shape and degrades to the unknown card. Previously `create` threw (`Cannot read properties of undefined (reading '_bn')`); `createIdempotent`/`recoverNested` degraded. Now all three degrade.
- **Dead `message` prop removed** from `InspectorInstructionCard` and its callers. It was declared *required* and never read; `TokenDetailsCard` didn't even declare it and a generic swallowed it silently.
- **Deleted `app/components/common/inspector/UnknownDetailsCard.tsx`**, which had no remaining callers once the duplicate card tree went.

### Behaviour change to be aware of

`createIdempotent` rendered **Wallet before Mint** while `create` rendered **Mint before Wallet**, for no reason. Unified on Mint→Wallet (what 2 of the 3 old renderers did). This is the only intentional visual change; everything else renders identically.

I verified that by A/B-ing the old components against the new one on the same fixtures — all three instruction types produce identical label→address maps on both surfaces, with `createIdempotent`'s row order as the sole difference.

## Type of change

-   [x] Bug fix
-   [x] Other (please describe): refactor — parser normalisation and de-duplication of the card

## Testing

- [Inspector — `createIdempotent`, v0 message with an address lookup table](https://explorer-o68shp76n-solana-foundation.vercel.app/tx/Nwa7VZEfhbPV2MmW4dQNebj6Pw1TrJH9j2G3pL7Vt4Hr2rTzyEPdh1LBd4SKtzEBhC1MKMDEeaQuz96FZzQSyYk/inspect) — the surface that used to index accounts positionally, so the one worth checking closely
- [Inspector — `create`](https://explorer-o68shp76n-solana-foundation.vercel.app/tx/4F3kU8ssrWwn7t4YCyVNNbxbP2D3YgHmPL6efGycLoxM3aRFo2Fi1D6UgzxcK8ZMTmgmBqavUtchA7S4dewSbnfs/inspect)
- [Inspector — `recoverNested`](https://explorer-o68shp76n-solana-foundation.vercel.app/tx/inspector?message=AQADByTPMvZ5NhbwY7GzM3bmF6aUB0Es9utyRgN3KoaqxFltNfKjDEAu3mQ7ldMPRzdZ2rwfown8mXJVsLSe%252FIoWPQObM34V%252BKEVoZ%252FbyZ2YI49FXsL%252FHdFCreG7S85NtrCJdK1H1URKpBmcuP98y%252Be8a7uDW4LI8LKAiAZBUh7wxywejJclj04kifG7PRApFI4NgwtaE5na%252FxCEBI572Nvp%252BFnG%252BnrzvtutOj1l82qryXQxsbvkwtL24OR8pgIDRS9dYQbd9uHXZaGT2cvhRs7reawctIXtX1s3kTqM9YV%252B%252FwCpsOI2EcJQ6duyss4%252B%252BR%252BYbQUQYEI4NS2%252Bk6O1be30VWMBBAcDBQECBQAGAQI%253D) — no real transaction to link; a scan of 250 recent Associated Token Program signatures found zero `recoverNested`, which is why its fixture is a hand-built message
- [Transaction page — `createIdempotent`](https://explorer-o68shp76n-solana-foundation.vercel.app/tx/Nwa7VZEfhbPV2MmW4dQNebj6Pw1TrJH9j2G3pL7Vt4Hr2rTzyEPdh1LBd4SKtzEBhC1MKMDEeaQuz96FZzQSyYk) — **Mint/Wallet order intentionally changed here**, now matching `create`
- [Transaction page — `create`](https://explorer-o68shp76n-solana-foundation.vercel.app/tx/4F3kU8ssrWwn7t4YCyVNNbxbP2D3YgHmPL6efGycLoxM3aRFo2Fi1D6UgzxcK8ZMTmgmBqavUtchA7S4dewSbnfs)
- [Transaction page — `createIdempotent` with inner cards](https://explorer-o68shp76n-solana-foundation.vercel.app/tx/XzQ1B1vAM775GMFdYTbs34VMBr25QxNh2BbHQkur2zUFFxke1AcsWtTk5omZwuF8As3zFZiXJADS3yfinU6v6Ef)

Same paths on [explorer.solana.com](https://explorer.solana.com) render the pre-change output, so each link can be compared side by side. Rows and addresses should match, with `createIdempotent`'s Mint/Wallet order as the one expected difference.

## Related Issues

Closes [HOO-916](https://linear.app/solana-fndn/issue/HOO-916/refactor-associated-token-card)

## Checklist

-   [ ] Screenshots included — required for protocol screens, recommended for other UI changes

## Additional Notes

Test changes are worth a look, since two specs were pinning the behaviour this PR removes:

- **`instruction-parser-contract.spec.ts`** previously asserted "mapped parity" for Associated Token — it hand-mapped `accounts.payer` → `source` to work around the two paths disagreeing. It now asserts the **same-validator parity** the other slices use: both paths validate against one struct and are compared field-for-field.
- **`associated-token-parser.spec.ts`** asserted the kit shape (`info.data`, `info.programAddress`, `info.accounts`). It now pins the canonical field names against real fixture addresses, because the kit→canonical mapping is the load-bearing logic.
- Added a regression test covering all three types for the raw-RPC-fallback case described above.

One pre-existing issue I noticed while testing, **not** introduced here and not addressed: on transaction pages with token accounts, `@solflare-wallet/utl-sdk` throws `chain(...).flatten is not a function` (`lib/esm/utils.js:113`) because it uses lodash explicit chaining and `chain` resolves to the standalone `lodash/chain` module, whose wrapper lacks the sequence methods. Effect is that token metadata lookup silently fails, so tokens fall back to raw mint addresses. It reproduces identically on `master` — I confirmed by running the same page on a clean checkout. Worth its own issue.


app/features/decode-instruction-associated-token/lib/associated-token-parser.ts
```diff
@@ -51,30 +61,86 @@ export function parseAssociatedTokenInstruction(ix: KitInstruction): AssociatedT
...
                 return { info: create(ix.parsed.info, RecoverNestedInfo), type: 'recoverNested' };
             default:
                 return undefined;
         }
-    } catch {
+    } catch (error) {
+        // The program label already matched, so a validation failure here means
+        // the RPC sent a payload we don't model — worth surfacing. Returning
+        // undefined makes the dispatcher fall back to the unknown-instruction card.
+        Logger.error(error, { instructionType: ix.parsed.type, program: ix.program });
```
mikhd, line 82: **question**: do we want this to be logged in sentry?

mikhd (approved):

app/features/decode-instruction-associated-token/lib/associated-token-parser.ts
askov: Not sure. I would say no, because if we start to log all the errors of such type, sentry could be overloaded

Woody4618 (approved):

PR #1177, opened 2026-07-31
fix(transaction-history): recover history on limited-retention RPC indexes

askov:
## Description

The account page rendered "Fetched full history" with **zero rows** for accounts whose transactions predate the RPC endpoint's index window.

**Root cause.** We fetch history with Triton's `getTransactionsForAddress` (gTFA) and fall back to `getSignaturesForAddress` (gSFA) only on a JSON-RPC method-not-found. Some endpoints back gTFA with a **limited-retention index** and answer `HTTP 200` with `{"data": [], "paginationToken": null}` for anything older than their floor. That is a *successful* empty page, not an error — so no error path ran, the fallback never fired, `foundOldest` was set from `!paginationToken`, and the empty page was cached and rendered as a complete history.

Measured on a Helius devnet endpoint: the retention floor sits between slot 477,000,000 and 478,000,000 against a tip of ~480,158,000 — roughly the last 2.5M slots (~11 days). `EKpQGSJtjMFqKZ9KQanSqYXRcF8fBopzLHYxdM65zcjm` has its newest transaction at slot 459,981,298, about 20M slots below that floor. gSFA on the same endpoint returns its transactions.

### Behaviour changes

-   Confirm an empty **unfiltered** gTFA page against gSFA before accepting it as the end of history; prefer gSFA's rows when the two disagree.
-   Skip the confirmation while a filter is active. gSFA cannot honour filters, so its rows would answer a different question, and an empty *filtered* page is a legitimate "no matches".
-   Latch a confirmed-uncovered address onto the gSFA path for its remaining unfiltered pages. The two methods page with incompatible cursors (opaque `paginationToken` vs trailing `before` signature); returning to gTFA with a null token re-requests the head and never advances. Per address, cleared on cluster change.
-   Recognise `-32603 "Method not found"` as a missing method. Some endpoints return it instead of `-32601`, and the previous numeric-code check rejected it before reading the message — so the fallback would not have fired even for an endpoint genuinely lacking gTFA. Every other structured code still short-circuits on the code alone, so `-32603 "Slot <= N not found"` stays a real error.
-   Validate the gTFA response with `superstruct` instead of `json.result as GetTransactionsForAddressResult`. The schema is loose (an endpoint may add fields) but strict on the fields the table reads. An unrecognised `confirmationStatus` is dropped per row rather than failing the page.

### Structural changes

`app/providers/accounts/history.tsx` (771 lines) held six concerns in one file: JSON-RPC transport, filter→param mapping, error classification, cache-merge rules, five React contexts, and five hooks. Split into the existing `transaction-history` feature slice:

-   `lib/` — pure, no React and no I/O: types, filter mapping, error classification, disabled-address list, cache-merge rules
-   `api/` — transport only: the three RPC calls
-   `model/` — contexts, method-selection policy, read hooks, write path

The pure logic is transcribed verbatim (verified by textual diff against the deleted file — only an added `export`).

Also in scope: deleted the unused `useFetchTransactionsForHistory`; normalised `paginationToken` to `string | undefined` internally with `null` confined to the wire; dropped the deleted file from the `unicorn/no-null` exemption list and justified the two contract-mandated nulls inline (web3.js types `err`/`memo` as `T | null`; the RPC requires an explicit null cursor).

The feature barrel deliberately exports **UI only** — re-exporting the data layer pulled `TransactionHistoryCard`'s render tree into every route that only wanted a hook, measured at **+30 kB** first-load on `/address/[address]/{instructions,tokens,transfers}`. Non-UI consumers deep-import from `model/` and `lib/`. `bench/BUILD.md` is unchanged for those routes.

## Type of change

-   [x] Bug fix
-   [x] Other: refactor — extract the account-history data layer into the feature slice

## Testing

-   [Devnet — the regression](https://explorer-git-fork-hoodieshq-fix-transa-162424-solana-foundation.vercel.app/address/EKpQGSJtjMFqKZ9KQanSqYXRcF8fBopzLHYxdM65zcjm?cluster=devnet) — Transaction History should list this account's transactions (newest at slot 459,981,298). On `master` it shows "Fetched full history" with no rows.
-   [Devnet — busy account, inside the retention window](https://explorer-git-fork-hoodieshq-fix-transa-162424-solana-foundation.vercel.app/address/TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA?cluster=devnet) — should still be served by `getTransactionsForAddress`: rows load, **Load More** keeps paging, and the slot / block-time / status filters stay enabled and apply.
-   [Devnet — wrapped SOL, gTFA disabled up front](https://explorer-git-fork-hoodieshq-fix-transa-162424-solana-foundation.vercel.app/address/So11111111111111111111111111111111111111112?cluster=devnet) — rows should load via `getSignaturesForAddress`, and filters should still be offered for *other* accounts in the same session.
-   [Mainnet — regression check on the same path](https://explorer-git-fork-hoodieshq-fix-transa-162424-solana-foundation.vercel.app/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v) — same code path, unfiltered and filtered.

Worth checking on the first entry: applying a filter is expected to show no rows, since the filterable index does not cover that account and `getSignaturesForAddress` cannot honour filters. That trade-off is described in the Description.

## Related Issues

Closes [HOO-957](https://linear.app/solana-fndn/issue/HOO-957/fix-empty-transaction-history-on-rpc-endpoints-with-a-limited)

## Checklist

## Additional Notes

**Not addressed here.** A method-not-found is not latched: on an endpoint without gTFA, every page still attempts gTFA, takes the `-32601`, then falls back — one wasted round-trip per page. This is pre-existing behaviour, unchanged by this PR. Fixing it needs a separate endpoint-wide flag, since the latch introduced here is per-address and deliberately filter-dependent.


app/features/transaction-history/model/use-fetch-account-history.ts
```diff
@@ -0,0 +1,237 @@
...
+            filters,
+            forceSignatures,
+            limit,
+            onMethodNotFound,
+            paginationToken,
+            pubkey,
+            url,
+        });
+        history = page.history;
+        if (page.signaturesOnly) onSignaturesOnly?.();
```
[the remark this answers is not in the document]
askov: Fixed

app/features/transaction-history/model/fetch-history-page.ts
```diff
@@ -0,0 +1,103 @@
...
+    // `data: []` for anything older than its floor — indistinguishable from a real
+    // end-of-history, and cached as `foundOldest: true`. getSignaturesForAddress reads the
+    // full ledger index, so it settles the question.
+    //
+    // Skip the check when a filter is active: getSignaturesForAddress can't honour filters,
+    // so its rows would answer a different question than the one that was asked, and an
+    // empty filtered page is a legitimate "no matches" answer anyway.
+    if (hasActiveFilters(filters)) return { history, signaturesOnly: false };
+
+    const confirmed = await fetchViaSignatures({ before, limit, pubkey, url });
```
mikhd: **question**: is there a case when gtfa succeed but gsfa fails, like:
- We have an empty account
- First fetch via `getTransactionsForAddress` returns correct empties `{data: [], paginationToken: null}`.
- Then at current line:96 `gSFA` starts fetching but fails with some error (e.g rate limit)
- So we've got an error `Failed to fetch transaction history` instead of empty `No transactions`.

mikhd: I mean should we try/catch and return the response from gtfa?

```ts
try {
    await fetchViaSignatures({ before, limit, pubkey, url });
} catch (error) {
    // Second try. If it fails, keep the gTFA answer rather than failing with error.
    Logger.warn('[transaction-history] gSFA fallback failed', {
        address: pubkey.toBase58(), error, url,
    });
    return { history, signaturesOnly: false };
}
```

app/features/transaction-history/model/fetch-history-page.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const history: AccountHistory = {
+        fetched: result.data,
+        // An absent paginationToken is the canonical end-of-stream signal; a short page
+        // is not, since the server may still hand back a token for more.
+        foundOldest: !result.paginationToken,
+        paginationToken: result.paginationToken,
+    };
+
+    if (result.data.length > 0) return { history, signaturesOnly: false };
```
mikhd: **question**: Regarding comment `An absent paginationToken is the canonical end-of-stream signal`.

Should we take it into account?
- Fetch without filters. (or first fetch with data, then do second fetch)
- Response returns an empty data and paginationToken  `({data: [], paginationToken: 'tok2'})`
- `result.data.length > 0` fails
- Should `paginationToken` signal us that this is the end-of-stream?

I.e.
```
if (result.data.length > 0 || result.paginationToken) return { history, signaturesOnly: false }
```

mikhd (approved):

app/features/transaction-history/model/fetch-history-page.ts
askov: Half right (Load More case), gated on the request cursor, added two tests. Thanks

app/features/transaction-history/model/fetch-history-page.ts
askov: Can't really imagine this situation, but you are right that it's better to handle anyway, added

Woody4618 (approved):

PR #1179, opened 2026-08-03
refactor(token-info): drop @solflare-wallet/utl-sdk, resolve mints server-side

askov:
## Description

`app/utils/token-info.ts` instantiated `@solflare-wallet/utl-sdk`'s `Client` in the browser. Under Next 16 + Turbopack that crashes: Next's unconditional `modularizeImports.lodash` rewrites `import { chain, chunk } from 'lodash'` in `utl-sdk/lib/esm/utils.js` into `import chain from 'lodash/chain'`, and `lodash/chain.js` never registers the lazy `.flatten()` / `.map()` methods on the wrapper prototype. The result was `chain(...).flatten is not a function` — an unhandled rejection on **every transaction with token movements**, so token symbols never loaded in `TokenBalancesCard`.

Both halves of the lookup now live behind `/api/token-info`, which accepts an array of addresses:

1. **UTL list** — the existing server-only `getTokenInfos` (`entities/token-info/api/fetch-token-mints.ts`).
2. **On-chain Metaplex fallback** — for mints the list does not carry, built on the Umi pipeline already in `entities/nft` (added in #911). It keeps the SDK's behaviour: `Fungible`-only token standard, `verified: false`, null-padding stripped from on-chain names, and the 5s per-mint `metaplexTimeout`.

The fallback is **opt-in** (`includeOnChainFallback`). The SDK's `fetchMints` included it, so batch callers get it; `getTokenInfoWithoutOnChainFallback` does not, keeping the contract each function already documented and avoiding new RPC cost on single-mint lookups.

The RPC endpoint is resolved server-side via `serverClusterUrl` rather than sent from the browser — a client-supplied URL would let a caller point the route at an arbitrary host.

### Why now

The SDK's last release was **2023-01-11** and it dragged `axios@0.27` plus the deprecated `@metaplex-foundation/js@0.11.7` (1.4 MB unminified) into the client bundle. Dropping it cuts **First Load JS by ~500–600 kB on every route** — see the regenerated `bench/BUILD.md` (e.g. `/address/[address]` 1.42 MB → 910 kB). No dependency changes landed between the previous `BUILD.md` baseline and `master`, so the delta is this branch.

### Also in here

- The SDK's `ChainId` and `Token` types are replaced by local equivalents, since removing the package takes its type exports with it.
- The two fetchers moved from `lib/` to the **`api/` segment** to match the convention used by `cluster`, `domain` and `idl`; `lib/` keeps only pure helpers.
- Three `boundaries/dependencies` allowlist entries are **retired** by routing `chain-id` access through a new `@x/token-info` public API instead of importing across entities directly.
- `getMetadataJson` stays browser-only, and it now checks `response.ok`. It passed every response to `processJson`, which only rejects a payload naming no artwork — so the proxy's `{ error }` body for a blocked or dead URI was handed back as an NFT's metadata. Its write-only `localStorage` cache is removed too: nothing ever read it, and it competed for the ~5 MB origin quota with `explorer:savedClusters`, whose writes are unguarded.
- **`ipfs://` metadata URIs resolve on the server path as well.** `fetchResource` speaks only http(s), so handing it a mint's raw on-chain `uri` cost every IPFS-hosted mint its logo. The gateway mapping `getProxiedUri` already did in the browser moved to **`app/shared/lib/ipfs.ts`** and is applied on both paths, rather than duplicated. An unparseable URI now short-circuits instead of failing inside `fetchResource`.
- The route validates its body with **superstruct**, as the other API routes here do, and so does the UTL response in `fetch-token-mints.ts`, which previously cast. Validation there is per token rather than per batch, so one malformed entry does not blank every symbol in a transaction. `decimals` and `logoURI` are `optional(nullable(…))`, not just `nullable`: UTL omits those keys outright for some tokens, and requiring them would drop the whole token — losing its name and symbol over a missing logo. `normalizeToken` puts back the `null` that `TokenInfo` declares. This matches `features/search/api/discover-with-utl.ts`, which reads the same API.
- The off-chain logo reads in the fallback run at **8 in flight** with a shared 10s per-batch deadline, instead of one socket per mint. Past the deadline the queued mints report `logoURI: null`, so names, symbols and decimals still arrive. The route carries `maxDuration = 30` as a platform backstop above that.

## Type of change

-   [x] Bug fix
-   [x] Other: removes an abandoned dependency and moves the lookup server-side

## Testing

Manual testing on the preview deployment. `PREVIEW` below is `https://explorer-git-fork-hoodieshq-refactor-r-f07d5c-solana-foundation.vercel.app`.

The on-chain fallback is reached from **one** surface: the transaction **Tokens** card. It is the only caller of `app/utils/token-info.ts#getTokenInfos`, which is where `includeOnChainFallback` is set. Every `<Address>`-based surface — a wallet's **Tokens** tab, token history, and a mint's *own* address page — resolves through `useTokenInfo` and the batched server action in `fetch-token-mints.ts`, which reads the UTL list only and does not opt in. Extending the fallback to those is follow-up work, not part of this PR.

-   [Transaction with an unlisted mint](https://explorer-git-fork-hoodieshq-refactor-r-f07d5c-solana-foundation.vercel.app/tx/3tRve9HV9xBthpQEKEM8ne4aRQWfhNoUXef33AQvCT69S4nHV7tB18mxrFzWe7W9T3iLy7JwLWjBvJTdhWM1wDhN) — the **Tokens** card should read `DECK` and `SOL`. `HWDHz1yh…afE7X` is absent from the UTL list, so its symbol comes from the new on-chain Metaplex read. Also covers de-duplication: 7 token-balance rows across only 2 distinct mints, so the request must collapse to 2 lookups.
-   [Wallet holding that unlisted mint](https://explorer-git-fork-hoodieshq-refactor-r-f07d5c-solana-foundation.vercel.app/address/94YQq71S3f8ERutz6JYw9XxJpgDRCBRbEs2JWXsyC4re/tokens) — **expected to be identical to production**: still the bare mint, with no symbol or logo. This surface goes through the batched server action, which does not opt into the fallback, so the link is a check that the transport change did not regress it rather than a place to look for a difference.
-   [Transaction with listed mints only](https://explorer-git-fork-hoodieshq-refactor-r-f07d5c-solana-foundation.vercel.app/tx/oeBBkD3HxhirfwesS7jg6CJFqZAHj2kBWLxWU672oC3V5DEFXBa7vkoqpcPfZUQPX9kBCL5sKHebkPTkfEz8DxP) — `USDC` / `SOL` from the UTL list, exercising the transport change without the fallback.
-   [A mint's own page](https://explorer-git-fork-hoodieshq-refactor-r-f07d5c-solana-foundation.vercel.app/address/EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v) — **expected to be identical to production.** It resolves through `getTokenInfoWithoutOnChainFallback`, which deliberately does not opt in, so this page is a check that nothing regressed rather than a place to look for a difference.

### Checking the route directly

The clearest way to see the change is the route itself, against the same unlisted mint. Production has no array form and no fallback:

```
curl -s https://explorer.solana.com/api/token-info -X POST -H 'Content-Type: application/json' \
  -d '{"address":"HWDHz1yhoHMSUGwectxheAa7L1XLnwsBWTRd1cXafE7X","cluster":0}'
→ {}

curl -s https://explorer.solana.com/api/token-info -X POST -H 'Content-Type: application/json' \
  -d '{"addresses":["HWDHz1yhoHMSUGwectxheAa7L1XLnwsBWTRd1cXafE7X"],"cluster":0,"includeOnChainFallback":true}'
→ {"error":"Invalid request"}
```

On the preview the array form is accepted, and opting in resolves the mint on chain:

```
curl -s $PREVIEW/api/token-info -X POST -H 'Content-Type: application/json' \
  -d '{"addresses":["HWDHz1yhoHMSUGwectxheAa7L1XLnwsBWTRd1cXafE7X"],"cluster":0}'
→ {"content":[]}

curl -s $PREVIEW/api/token-info -X POST -H 'Content-Type: application/json' \
  -d '{"addresses":["HWDHz1yhoHMSUGwectxheAa7L1XLnwsBWTRd1cXafE7X"],"cluster":0,"includeOnChainFallback":true}'
→ {"content":[{"address":"HWDHz1yh…afE7X","decimals":6,"logoURI":"https://…/QmQhZNEPuG…",
               "name":"Chain Deck","symbol":"DECK","verified":false}]}
```

The `logoURI` confirms the off-chain JSON read works through the proxy's `fetchResource` in process, and `verified: false` marks data that did not come from a curated list.

## Related Issues

Closes [HOO-610](https://linear.app/solana-fndn/issue/HOO-610/remove-solflare-walletutl-sdk-from-client)

## Checklist

-   [x] All checks pass locally (`pnpm test`, `pnpm lint`, `pnpm exec tsc --noEmit`, `pnpm format`)

## Additional Notes

Two behaviours differ from the old SDK on purpose:

- **Decimals of `0` are preserved.** `transformMetaplexToken` used `decimalsMap[mint] || 6`, turning a legitimate `0` into `6`.
- **No jsdelivr CDN fallback** when the UTL backend errors. The server-only `fetch-token-mints.ts` this reuses returns `[]` instead; the on-chain fallback now partly covers that case. Restoring the CDN path belongs in `fetch-token-mints.ts` as separate work.

The `Fungible`-only filter is carried over deliberately — it means unlisted legacy mints with no `tokenStandard` set are still skipped, exactly as before.

One pre-existing issue this brushes against but does not fix: `app/features/custom-cluster/lib/cluster-storage.ts` writes to `localStorage` unguarded, so a full quota throws. Removing the dead metadata cache makes that far less likely to trigger, but the fragility remains. Happy to open a follow-up.

### DIFF
--- a/app/api/token-info/__tests__/route.spec.ts
+++ b/app/api/token-info/__tests__/route.spec.ts
@@ -1,13 +1,17 @@
-import { PublicKey } from '@solana/web3.js';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { gen } from '@/app/__fixtures__/gen';
 import type { TokenInfo } from '@/app/entities/token-info/server';
 import { Logger } from '@/app/shared/lib/logger';
 import { Cluster } from '@/app/utils/cluster';
 
-const MINT_A = 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v';
-const MINT_B = 'So11111111111111111111111111111111111111112';
-const MINT_C = 'mSoLzYCxHdYgdzU16g5QSh3i5K3z3KZK7ytfqcJm7So';
+import { MAX_ADDRESSES } from '../config';
+
+// The route treats a mint as an opaque, well-formed address, so generated ones say more about
+// what is under test than real mints would.
+const MINT_A = gen.address(1);
+const MINT_B = gen.address(2);
+const MINT_C = gen.address(3);
 
 const mocks = vi.hoisted(() => ({
     getTokenInfos: vi.fn(),
@@ -79,9 +83,32 @@ describe('POST /api/token-info', () => {
         expect(mocks.getTokenInfos).not.toHaveBeenCalled();
     });
 
+    it('should return 400 when the cluster is not a number', async () => {
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A], cluster: '0' }));
+
+        expect(res.status).toBe(400);
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+
+    it('should return 400 when the genesis hash is not a string', async () => {
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A], cluster: Cluster.Custom, genesisHash: 42 }));
+
+        expect(res.status).toBe(400);
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+
+    it('should ignore keys it does not know', async () => {
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A], cluster: Cluster.MainnetBeta, verbose: true }));
+
+        expect(res.status).toBe(200);
+    });
+
     it('should return 400 when more distinct addresses than the cap are requested', async () => {
         const { POST } = await importRoute();
-        const addresses = Array.from({ length: 129 }, (_, i) => addressAt(i));
+        const addresses = Array.from({ length: MAX_ADDRESSES + 1 }, (_, i) => gen.address(i));
         const res = await POST(createRequest({ addresses, cluster: Cluster.MainnetBeta }));
 
         expect(res.status).toBe(400);
@@ -222,14 +249,6 @@ describe('POST /api/token-info', () => {
         expect(mocks.getTokenInfos).not.toHaveBeenCalled();
     });
 });
-
-/** Builds a distinct, well-formed mint address for cap tests. */
-function addressAt(index: number): string {
-    const bytes = new Uint8Array(32);
-    bytes[0] = index + 1;
-    bytes[1] = Math.floor((index + 1) / 256);
-    return new PublicKey(bytes).toBase58();
-}
 
 function createRequest(body: Record<string, unknown>) {
     return new Request('http://localhost:3000/api/token-info', {
--- a/app/api/token-info/__tests__/route.spec.ts
+++ b/app/api/token-info/__tests__/route.spec.ts
@@ -170,10 +170,12 @@ describe('POST /api/token-info', () => {
             }),
         );
 
+        // The endpoint is resolved from server config, and nothing is derived from the incoming
+        // request — a caller-controlled Host must not be able to redirect the fallback's fetches.
         expect(mocks.getTokenInfosFromMetaplex).toHaveBeenCalledWith(
             [MINT_B, MINT_C],
             expect.any(String),
-            expect.objectContaining({ baseUrl: 'http://localhost:3000' }),
+            expect.not.objectContaining({ baseUrl: expect.anything() }),
         );
         expect(await res.json()).toEqual({ content: [token(MINT_A), token(MINT_C, { verified: false })] });
     });
--- a/app/api/token-info/__tests__/route.spec.ts
+++ b/app/api/token-info/__tests__/route.spec.ts
@@ -1,0 +1,242 @@
+import { PublicKey } from '@solana/web3.js';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { TokenInfo } from '@/app/entities/token-info/server';
+import { Logger } from '@/app/shared/lib/logger';
+import { Cluster } from '@/app/utils/cluster';
+
+const MINT_A = 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v';
+const MINT_B = 'So11111111111111111111111111111111111111112';
+const MINT_C = 'mSoLzYCxHdYgdzU16g5QSh3i5K3z3KZK7ytfqcJm7So';
+
+const mocks = vi.hoisted(() => ({
+    getTokenInfos: vi.fn(),
+    getTokenInfosFromMetaplex: vi.fn(),
+}));
+
+// The UTL list lookup and the Metaplex fallback each have their own spec. The route is the transport
+// edge, so both are mocked here to exercise body parsing, the address cap, response shaping, and the
+// opt-in fallback policy. `isValidCluster` stays real so cluster handling is exercised end to end.
+vi.mock('@entities/token-info/server', async () => {
+    const actual = await vi.importActual<typeof import('@entities/token-info/server')>('@entities/token-info/server');
+    return {
+        ...actual,
+        getTokenInfos: mocks.getTokenInfos,
+        getTokenInfosFromMetaplex: mocks.getTokenInfosFromMetaplex,
+    };
+});
+
+function token(address: string, overrides: Partial<TokenInfo> = {}): TokenInfo {
+    return {
+        address,
+        decimals: 6,
+        logoURI: null,
+        name: `Token ${address.slice(0, 4)}`,
+        symbol: 'TKN',
+        verified: true,
+        ...overrides,
+    };
+}
+
+describe('POST /api/token-info', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+        vi.spyOn(Logger, 'warn').mockImplementation(() => {});
+        mocks.getTokenInfos.mockResolvedValue([]);
+        mocks.getTokenInfosFromMetaplex.mockResolvedValue([]);
+    });
+
+    it('should return 400 when the body is not valid JSON', async () => {
+        const { POST } = await importRoute();
+        const res = await POST(new Request('http://localhost:3000/api/token-info', { body: 'nope', method: 'POST' }));
+
+        expect(res.status).toBe(400);
+        expect(await res.json()).toEqual({ error: 'Invalid request' });
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+
+    it('should return 400 when neither address nor addresses is given', async () => {
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ cluster: Cluster.MainnetBeta }));
+
+        expect(res.status).toBe(400);
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+
+    it('should return 400 for a malformed address', async () => {
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A, 'not-a-pubkey'], cluster: Cluster.MainnetBeta }));
+
+        expect(res.status).toBe(400);
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+
+    it('should return 400 for an unsupported cluster', async () => {
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A], cluster: 999 }));
+
+        expect(res.status).toBe(400);
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+
+    it('should return 400 when more distinct addresses than the cap are requested', async () => {
+        const { POST } = await importRoute();
+        const addresses = Array.from({ length: 129 }, (_, i) => addressAt(i));
+        const res = await POST(createRequest({ addresses, cluster: Cluster.MainnetBeta }));
+
+        expect(res.status).toBe(400);
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+
+    it('should accept a request whose repeats push it over the cap only before de-duplication', async () => {
+        const { POST } = await importRoute();
+        // A single transaction can move one mint through far more than `MAX_ADDRESSES` token
+        // accounts; that must not be rejected, because it is one distinct lookup.
+        const addresses = Array.from({ length: 400 }, () => MINT_A);
+        const res = await POST(createRequest({ addresses, cluster: Cluster.MainnetBeta }));
+
+        expect(res.status).toBe(200);
+        expect(mocks.getTokenInfos).toHaveBeenCalledWith([MINT_A], Cluster.MainnetBeta, undefined, expect.anything());
+    });
+
+    it('should accept a single address and still answer with an array', async () => {
+        mocks.getTokenInfos.mockResolvedValueOnce([token(MINT_A)]);
+
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ address: MINT_A, cluster: Cluster.MainnetBeta }));
+
+        expect(res.status).toBe(200);
+        expect(await res.json()).toEqual({ content: [token(MINT_A)] });
+        expect(mocks.getTokenInfos).toHaveBeenCalledWith([MINT_A], Cluster.MainnetBeta, undefined, expect.anything());
+    });
+
+    it('should resolve several addresses in one request', async () => {
+        mocks.getTokenInfos.mockResolvedValueOnce([token(MINT_A), token(MINT_B)]);
+
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A, MINT_B], cluster: Cluster.MainnetBeta }));
+
+        expect(await res.json()).toEqual({ content: [token(MINT_A), token(MINT_B)] });
+        expect(mocks.getTokenInfos).toHaveBeenCalledWith(
+            [MINT_A, MINT_B],
+            Cluster.MainnetBeta,
+            undefined,
+            expect.anything(),
+        );
+    });
+
+    it('should de-duplicate repeated addresses', async () => {
+        const { POST } = await importRoute();
+        await POST(createRequest({ addresses: [MINT_A, MINT_A, MINT_B], cluster: Cluster.MainnetBeta }));
+
+        expect(mocks.getTokenInfos).toHaveBeenCalledWith(
+            [MINT_A, MINT_B],
+            Cluster.MainnetBeta,
+            undefined,
+            expect.anything(),
+        );
+    });
+
+    it('should forward the genesis hash so a custom cluster can resolve a chain id', async () => {
+        const genesisHash = '5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d';
+
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A], cluster: Cluster.Custom, genesisHash }));
+
+        expect(res.status).toBe(200);
+        expect(mocks.getTokenInfos).toHaveBeenCalledWith([MINT_A], Cluster.Custom, genesisHash, expect.anything());
+    });
+
+    it('should not read on-chain metadata unless the caller opts in', async () => {
+        mocks.getTokenInfos.mockResolvedValueOnce([]);
+
+        const { POST } = await importRoute();
+        const res = await POST(createRequest({ addresses: [MINT_A], cluster: Cluster.MainnetBeta }));
+
+        expect(await res.json()).toEqual({ content: [] });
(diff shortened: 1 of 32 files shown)


app/entities/nft/lib/get-metadata-json.ts
```diff
@@ -6,52 +6,70 @@ import type { NftJson } from './types';
...
+    // Building the request URI is synchronous, and a failure means bad input (a malformed
+    // `baseUrl`), so it is reported to the caller.
+    let requestUri: string;
+    try {
+        requestUri = resolveRequestUri(getProxiedUri(uri), deps?.baseUrl);
+    } catch (error) {
+        deps?.onError?.(error);
+        return undefined;
+    }
+
```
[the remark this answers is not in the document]
askov: Good catch — this is real, and it was introduced by this PR. Fixed in e3065bd5.

You are right about the mechanism: `/api/metadata/proxy` returns 404 unless `NEXT_PUBLIC_METADATA_ENABLED === "true"`, and in exactly that case `getProxiedUri` falls through to `return uri` — the raw, attacker-controlled on-chain URI. Client-side that is the user's own network and predates this branch, but moving `getMetadataJson` server-side made it an SSRF from our infrastructure, reachable unauthenticated via `POST /api/token-info` with `includeOnChainFallback: true` and a mint whose metadata URI points at `169.254.169.254` or a private host.

Worth adding to your report: it is not only the disabled-proxy case. Even with the flag on, `getProxiedUri` returns the URI unchanged for any protocol outside `SAFE_EXTERNAL_PROTOCOLS`, so those bypassed the proxy too.

The fix inverts the default for server callers rather than patching the two known holes: a call passing `baseUrl` now fetches **only** URIs that resolve to the `/api/metadata/proxy` path and refuses everything else, so it fails closed if `getProxiedUri` grows another pass-through branch. Browser behaviour is untouched.

The tradeoff is a missing logo when metadata proxying is disabled. Names and symbols come from the metadata account rather than the off-chain JSON, so they are unaffected. Three tests pin the boundary, including the metadata-endpoint case and one asserting the browser path still fetches directly.

app/api/token-info/route.ts
```diff
@@ -1,22 +1,112 @@
...
+    requestUrl: string,
+): Promise<TokenInfo[]> {
+    const found = new Set(listed.map(token => token.address));
+    const missing = addresses.filter(address => !found.has(address));
+    if (missing.length === 0) return listed;
+
+    // Resolved server-side on purpose: a client-supplied RPC URL would let a
+    // caller point this route at an arbitrary host.
+    const rpcEndpoint = serverClusterUrl(cluster, '');
+    if (!rpcEndpoint) return listed;
```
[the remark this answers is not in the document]
askov: Not taking this one, and I want to lay out why — partly because the suggested remedy is the vulnerability in the sibling comment.

**It is not a regression.** `getChainId` returns `undefined` for `Cluster.Custom` unless the caller supplies a `genesisHash` matching a known network, and neither caller does — `TokenBalancesCard` calls `getTokenInfos(mints, cluster)` and `TokensProvider` calls it with `(mints, cluster)`. On master that meant `makeUtlClient` returned `undefined`, so `getTokenInfos` resolved to `undefined` and custom-cluster users already saw no names, symbols or logos. This branch short-circuits on the same `getChainId` check before issuing the request, so the observable result is identical.

Where a chain id *does* resolve (a custom cluster with a matching `genesisHash`, currently unreachable from either caller), the UTL list is still served in full — only the on-chain fallback is skipped, so the degradation is partial rather than data "disappearing".

**The fix would open an SSRF.** The only way to give the server a custom endpoint is to forward the browser's RPC URL in the request body, which lets any caller point this route at an arbitrary host — the same class of bug flagged at `get-metadata-json.ts:65`, which I have just fixed in e3065bd5. Trading that for a code path nothing currently reaches is not a good deal.

I have expanded the comment at that line in e3065bd5 to record the tradeoff and the reasoning, since this is a fair question for the next reader. If custom clusters should get on-chain token metadata, the way there is an explicit server-side allowlist or validation of user-supplied endpoints — worth its own issue rather than falling out of this refactor.

app/api/token-info/route.ts
```diff
@@ -1,22 +1,118 @@
...
+    // and the on-chain fallback is skipped. No regression: a custom cluster has no chain id
+    // unless the caller supplies a matching `genesisHash`, and neither `TokenBalancesCard` nor
+    // `TokensProvider` does, so `getTokenInfos` already returned nothing there. Where a chain
+    // id does resolve, the UTL list is still served — only the on-chain fallback is dropped.
+    const rpcEndpoint = serverClusterUrl(cluster, '');
+    if (!rpcEndpoint) return listed;
+
+    try {
+        const onChain = await getTokenInfosFromMetaplex(missing, rpcEndpoint, {
+            baseUrl: new URL(requestUrl).origin,
```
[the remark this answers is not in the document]
askov: Valid, and worse than a leak — fixed in 4feda1c8.

You are right that `baseUrl: new URL(request.url).origin` trusts the incoming host. Two consequences, the second being the one that worried me: with a forwarded host of `169.254.169.254` the server fetches `http://169.254.169.254/api/metadata/proxy?uri=...`, which **reopens the SSRF from the sibling comment by a different route** — my previous fix guaranteed the request went through `/api/metadata/proxy`, but not that the origin was ours.

So rather than harden the origin, I removed the need for one: the fallback now calls the proxy's `fetchResource` **in process** instead of requesting `/api/metadata/proxy` over HTTP. That is the same code the route delegates to, so it keeps per-hop private-IP rejection, protocol and redirect validation, and the size cap — while the self-directed HTTP hop and every request-derived value disappear. Strictly less surface, and one less network round trip per mint.

`getMetadataJson` is browser-only again, which is what it was before this PR: `baseUrl`, `signal` and the proxy-path guard are gone, and it is no longer exported from nft's `@x/token-info`. Worth naming the pattern, since both P1s trace to it: the root cause was making one client helper serve the server as well, which silently changed its threat model without changing its call sites. The two paths are separate now.

Logos still resolve for unlisted mints, so the fallback keeps the `logoURI` behaviour HOO-610 specified. Four tests cover the new path, including a regression guard asserting no request-derived origin reaches it and that a `fetchResource` rejection degrades to `logoURI: null`.

app/api/token-info/__tests__/route.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        const res = await POST(createRequest({ addresses: [], cluster: Cluster.MainnetBeta }));
+
+        expect(res.status).toBe(200);
+        expect(await res.json()).toEqual({ content: [] });
+        expect(mocks.getTokenInfos).not.toHaveBeenCalled();
+    });
+});
+
+/** Builds a distinct, well-formed mint address for cap tests. */
+function addressAt(index: number): string {
```
mikhd: **nit**: can we use `app/__fixtures__/gen.ts` instead?

app/api/token-info/route.ts
```diff
@@ -1,22 +1,110 @@
...
+import { Logger } from '@/app/shared/lib/logger';
+
 const CACHE_MAX_AGE = 3600; // 1 hour
 
+/**
+ * Caps how many distinct mints one request may resolve. `TokensProvider` sends
+ * up to 101 token accounts, so this leaves headroom without letting a caller
+ * fan out an unbounded number of RPC lookups.
+ */
+const MAX_ADDRESSES = 128;
```
mikhd: **question**: I'm wondering if this could be exported and re-used in cap tests, i.e MAX_ADDRESSES + 1?

app/api/token-info/route.ts
```diff
@@ -1,22 +1,110 @@
...
 export async function POST(request: Request) {
-    const { address, cluster, genesisHash } = await request.json();
+    let body: RequestBody;
+    try {
+        body = await request.json();
+    } catch {
+        return NextResponse.json({ error: 'Invalid request' }, { status: 400 });
+    }
 
+    const { cluster, genesisHash } = body;
```
mikhd: **question**: will this make sense to validate body via superstruct?

And the response from `getTokenInfos`.
I think currently it typecasts -   https://github.com/hoodieshq/explorer/blob/4feda1c883b208c6df38ed14e1793b9037c82da7/app/entities/token-info/api/fetch-token-mints.ts#L36

app/entities/nft/lib/get-metadata-json.ts
```diff
@@ -11,47 +11,54 @@ export type GetMetadataJsonDeps = {
...
+    // A dead link, a timeout, or an unparseable body is an everyday outcome for third-party
+    // metadata. Those stay quiet: the NFT page maps `onError` to `Logger.error`, so reporting them
+    // would escalate the ordinary case.
+    try {
+        const response = await fetch(requestUri);
```
mikhd, line 59: **question:** I think we should check for `response.ok`?

e.g. https://github.com/solana-foundation/explorer/pull/1179/changes#diff-089a3e1c45204dd7f38b63f7e17bd0010655def156b17fbba5210c7332b40f65R69

app/api/token-info/route.ts
```diff
@@ -1,22 +1,110 @@
...
+
+type RequestBody = {
+    address?: unknown;
+    addresses?: unknown;
+    cluster?: unknown;
+    genesisHash?: unknown;
+    includeOnChainFallback?: unknown;
+};
+
 export async function POST(request: Request) {
```
mikhd, line 37: **question**: do we need `maxDuration` - same guard as in `app/api/metadata/proxy/route.ts` to keep route safe?

app/entities/token-info/api/fetch-token-metaplex.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    if (fungible.length === 0) return [];
+
+    const [decimals, logoUris] = await Promise.all([
+        fetchDecimals(
+            fungible.map(metadata => metadata.mint.toString()),
+            rpcEndpoint,
+            options.onError,
+        ),
+        Promise.all(fungible.map(metadata => fetchLogoUri(metadata, options))),
```
mikhd: **question:** Here we don't have cap, so we may have 128 calls, if i'm not mistaken?
Does it make sense to batch them or limit to 8-10 in flight? (but to keep result order)

app/api/token-info/__tests__/route.spec.ts
```diff
@@ -0,0 +1,244 @@
+import { PublicKey } from '@solana/web3.js';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { TokenInfo } from '@/app/entities/token-info/server';
+import { Logger } from '@/app/shared/lib/logger';
+import { Cluster } from '@/app/utils/cluster';
+
+const MINT_A = 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v';
+const MINT_B = 'So11111111111111111111111111111111111111112';
+const MINT_C = 'mSoLzYCxHdYgdzU16g5QSh3i5K3z3KZK7ytfqcJm7So';
```
mikhd: Do we want fixture/gen.ts helpers for addresses?

mikhd (approved):

app/api/token-info/route.ts
askov: Done in 243173d7 — the cap now lives in `app/api/token-info/config.ts` alongside `CACHE_MAX_AGE`, and the spec asserts `MAX_ADDRESSES + 1`.

Not a bare export from `route.ts`: Next reads route segment config as literal route exports, and the same split is already documented in `app/api/metadata/proxy/config.ts`. So `maxDuration` stays inline and everything else moves to `config.ts`.

app/api/token-info/route.ts
askov: Both taken, in 243173d7.

**Request body.** `RequestStruct` now covers it, so the hand-rolled `typeof` checks are gone and the address format is a `refine` on `string()`. Two rules stay outside the struct because they are not shape checks: the cap applies *after* de-duplication (400 unique-once repeats of one mint is one lookup, not 400 slots), and a cluster is `number()` here because `isValidCluster` decides whether it resolves a chain id. Two things got stricter as a side effect — a non-string `genesisHash` and a non-number `cluster` are now 400 rather than silently ignored; both are pinned by tests.

**`getTokenInfos` response.** You are right that the cast was doing no work — a UTL shape change reached the UI as `undefined` fields. It is a superstruct schema now, matching what the jupiter/coingecko routes already do.

One deliberate difference from those routes: validation is **per token, not per batch**. They answer 502 on any mismatch, which is right when the payload is one token; here a single malformed entry would blank every symbol in a transaction. So a bad entry is dropped and the rest are kept, with the drop reported through `onError` so upstream drift stays visible. `type()` also ignores fields the app does not read (`chainId`, `tags`, `holders`), so an upstream addition is not a mismatch — tested.

Happy to switch it to all-or-nothing for consistency with the other routes if you would rather have that.

app/entities/nft/lib/get-metadata-json.ts
askov: Good catch — that was a real bug, fixed in 0bc008cd.

Without the check, `processJson` received the error body. It only rejects a payload that names no artwork, and `{ error: "Not Found" }` fails that test the wrong way: `extended.image` is `undefined` but `extended?.properties?.files?.length === 0` is `false`, so the guard does not fire and the error body was handed back **as the NFT metadata**.

It fires on two real inputs, not just a hypothetical: the proxy returns `{ error }` with 400/403/404/413/415/502 for a blocked, dead or oversize URI, and an upstream can serve its own JSON error page. Test added.

app/api/token-info/route.ts
askov: Yes — added in 243173d7 as `export const maxDuration = 30`.

Sizing it, per invocation: the UTL list lookup, then at most four RPC calls (`MAX_ADDRESSES` keys chunked 100 at a time, for metadata PDAs and mint accounts), then the off-chain logo reads. The reads are now bounded as a group by `LOGO_BUDGET_MS` (10s, see the sibling thread on `fetch-token-metaplex.ts`), so 30 is headroom above the work rather than a ceiling the fallback can hit — being killed mid-flight would drop the listed tokens too, which is the opposite of what the fallback is for.

Higher than the proxy's 15 for that reason: the proxy fetches one resource, this one aggregates a batch.

app/entities/token-info/api/fetch-token-metaplex.ts
askov: You are not mistaken — 128 concurrent reads to third-party hosts, one per mint. Limited in 9b3349b3.

It now goes through `fetchAll` (`app/utils/fetch-all.ts`, `p-limit`) at 8 in flight, which keeps the results in the order the mints were given, so the `logoUris[index]` lookup below still lines up with `fungible[index]`. A test pins both the peak in flight and the ordering under out-of-order completion.

One thing the limit alone does not fix, which is why the change is slightly larger than asked: it converts a wide fan-out into a long queue. 128 mints at 8 in flight is 16 waves, and at the 5s per-read timeout that is 80s worst case — past any sane `maxDuration`, so the function would be killed and the listed tokens lost along with the logos. So the reads also share a per-batch deadline (`LOGO_BUDGET_MS`, 10s): once it lapses the queued mints report `logoURI: null` and the response still carries their names, symbols and decimals, which come from the metadata and mint accounts rather than the logo read. Degraded logos beat no response.

Batching in the `getMultipleAccounts` sense is already there for the RPC side (`ACCOUNTS_CHUNK_SIZE`, 100) — these are individual HTTP reads of arbitrary third-party URIs, so there is nothing to coalesce.

app/api/token-info/__tests__/route.spec.ts
askov: Done in 243173d7 — the local `addressAt` is gone, replaced by `gen.address(i)`.

I removed the same duplicated helper from `fetch-token-metaplex.spec.ts` too (9b3349b3), which is where I had copied it from in the first place.

app/api/token-info/__tests__/route.spec.ts
askov: Yes — the three constants are `gen.address(1..3)` now (243173d7).

The route treats a mint as an opaque, well-formed address; it never looks at *which* mint. Real ones (USDC/wSOL/mSOL) implied a significance that is not there, and one of them was doing double duty as "the mint the UTL list does not carry", which a generated address states more plainly.

I left the real mints in `fetch-token-metaplex.spec.ts`, since a couple of assertions there read better against a recognisable USDC.

mikhd (approved):

askov:
@greptile-apps Please review one more time, all the commits are signed an verified

Woody4618 (approved):

PR #1185, opened 2026-08-10
feat(stake): show lifetime total reward on the stake delegation card

askov:
## Description

Adds a **Total Reward (SOL)** row to the Stake Delegation card, showing a stake account's lifetime staking rewards.

The figure cannot come from account state. Each epoch's reward is added to **both** the lamport balance and `delegation.stake`, and the principal at delegation time is never stored, so the two are indistinguishable on chain. Two plausible formulas are wrong: `lamports − rentExemptReserve − delegation.stake` yields the inactive stake, and `lamports − initial_deposit` overshoots by MEV tip claims and dust. No RPC method returns the total either — `getInflationReward` serves one epoch per call. So it comes from Solscan's per-epoch reward index.

- A `stake-rewards` entity pages `GET /v2.0/account/stake/reward` — the only Solscan endpoint whose `address` parameter is documented as a stake account — and sums `amount`. The range is floored at mainnet genesis, passed explicitly so Solscan's one-month default window cannot pass as a lifetime total. Cost is `ceil(epochs / 100)` requests, bounded by cluster age: at most 10, one for a 70-epoch account.
- `GET /api/stake-rewards/[address]` keeps the key server-side and shares one result through the CDN for 4 hours. Mainnet only — Solscan indexes no other cluster, and answering for devnet would serve a mainnet figure under a devnet address. A rate limit passes through as 429; other upstream failures return an uncached 502.
- **The row fails closed.** A failed page, the paging bound, a row not denominated in lamports, or a sum beyond exact integer arithmetic all render `Unavailable` rather than a number we cannot stand behind — and never `0`, which would be a claim about the account rather than about the request.
- **Off by default.** `NEXT_PUBLIC_STAKE_TOTAL_REWARD_ENABLED` gates both ends: with it unset the row is not rendered at all and the route refuses before validating the address. It is client-readable on purpose — otherwise the browser cannot tell an unprovisioned deployment from a broken one, and renders `Unavailable` on every stake page.
- **The metered request is gated on a stake-account check.** The route is public, so one `getAccountInfo` (owner only) confirms the address is owned by the stake program before any Solscan call. Enumerating arbitrary addresses costs one cheap RPC call rather than up to ten metered ones.

Rationale and the rejected alternatives are in `openspec/changes/add-stake-total-reward-via-solscan/proposal.md`.

## Type of change

-   [x] New feature

## Screenshots

No screenshot yet: the feature is off by default, and with the flag on but no provisioned `solscan_api` plan the row renders `Unavailable`. A capture would show the failure state rather than the feature. See Additional Notes.

## Testing

Manual testing on the preview deployment. Set `NEXT_PUBLIC_STAKE_TOTAL_REWARD_ENABLED=true` first — the feature is off by default, so without it no row renders anywhere.

- [Stake account, 70 epochs](https://explorer-git-feat-stake-total-reward-from-api-solana-foundation.vercel.app/address/116vkzEjoTpFEd3x12XL9HbYJ5EpoC4cZ9a1N5A5mUt) — Stake Delegation card should carry a **Total Reward (SOL)** row reading **0.004263041 SOL**.
- [Stake account, re-delegated, 169 epochs](https://explorer-git-feat-stake-total-reward-from-api-solana-foundation.vercel.app/address/MCrWxQJgT3VogbgM5sy78K4uCEmwPQiKeXG2Bna51zs) — should read **0.055208535 SOL**, not 0.010482997 SOL. See Additional Notes.
- [Not a stake account](https://explorer-git-feat-stake-total-reward-from-api-solana-foundation.vercel.app/address/So11111111111111111111111111111111111111112) — no Stake Delegation card, so no Total Reward row.
- [Delegated stake account on devnet](https://explorer-git-feat-stake-total-reward-from-api-solana-foundation.vercel.app/address/5CSTrn37yJEpwhVkEuHMrzY4JuFkDjwSgTQNDXzNHKmk?cluster=devnet) — the card renders and the row reads `Unavailable`: the route refuses non-mainnet clusters rather than serving a mainnet figure. This account exists only on devnet.

Until a Solscan plan is provisioned, every mainnet row above reads `Unavailable`.

## Related Issues

Closes [HOO-898](https://linear.app/solana-fndn/issue/HOO-898/add-total-reward-field-to-stake-account-overview)

## Checklist

## Additional Notes

Two accepted limitations, both recorded in the proposal:

- **Inflation rewards only.** MEV tips reach a delegator's stake account as a claim transfer rather than an epoch reward, so they are absent from this endpoint and the row reads lower than the account's actual growth.
- **The first visitor after a cache miss waits for every page** — up to 10 sequential upstream calls before the row resolves.


.env.example
```diff
@@ -55,6 +55,11 @@ SEARCH_JUPITER_API_KEY=
 RUGCHECK_API_KEY=
 ### Configuration for CoinGecko API
```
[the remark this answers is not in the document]
askov: That is not true, both commits are verified

app/features/stake/model/use-total-reward.ts
```diff
@@ -0,0 +1,54 @@
...
+    const { cluster } = useCluster();
+    // Solscan indexes mainnet-beta only, so other clusters skip the request rather than fail it.
+    const isSupported = cluster === Cluster.MainnetBeta;
+
+    const { data, error, isLoading } = useSWRImmutable(
+        isSupported && (['stake-total-reward', stakeAccountAddress] as const),
+        () => fetchTotalReward(stakeAccountAddress),
+        // fetchTotalReward throws rather than returning undefined, so cap the retries.
+        { errorRetryCount: 3 },
+    );
```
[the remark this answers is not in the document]
askov: That is not p1 and there is a 4-hour CDN window anyway. But replaced with useSWR

app/api/stake-rewards/[address]/route.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    try {
+        // Gate the metered request behind one cheap RPC call: the route is public, and every cache
+        // miss spends shared paid quota across up to ten upstream calls.
+        if (!(await isStakeAccount({ address, rpcUrl: serverClusterUrl(Cluster.MainnetBeta, '') }))) {
+            return NextResponse.json({ error: 'Not a stake account' }, { headers: ERROR_CACHE_HEADERS, status: 404 });
+        }
+
+        const { epochs, lamports } = await fetchTotalStakeReward({ address, apiKey });
```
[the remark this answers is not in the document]
Comment (askov) [reply app/api/stake-rewards/[address]/route.ts:66]: It is valid, but it also matches the project style: every request that uses a third-party API is not gated by auth. Added a gate for non-stake address.

app/api/stake-rewards/[address]/route.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (!apiKey) {
+        Logger.warn('[api:stake-rewards] SOLSCAN_API is not configured', { sentry: true });
+        return NextResponse.json(
+            { error: 'Stake rewards are not configured' },
+            { headers: NO_STORE_HEADERS, status: 503 },
+        );
+    }
+
+    try {
+        const { epochs, lamports } = await fetchTotalStakeReward({ address, apiKey });
```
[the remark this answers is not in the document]
Comment (askov) [reply app/api/stake-rewards/[address]/route.ts:66]: Answered here https://github.com/solana-foundation/explorer/pull/1185#discussion_r3748070822

.env.example
```diff
@@ -55,6 +55,12 @@ SEARCH_JUPITER_API_KEY=
...
```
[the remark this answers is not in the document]
askov: They are signed

askov:
@greptile-apps Review again, all the commits are signed and verified

kejonbrown27-byte (approved):

Woody4618 (dismissed):

app/features/stake/ui/StakeAccountSection.tsx
```diff
@@ -237,6 +261,26 @@ function DelegationCard({
...
 }
 
+// Lifetime rewards load separately from the account, so the row carries its own state. A failure
+// shows a quiet message rather than 0 — zero is a claim about the account, not about the request.
+// `disabled` is excluded because the caller drops the whole row, leaving this switch exhaustive.
+function TotalReward({
+    state,
+    solPrice,
+}: {
+    state: Exclude<TotalRewardState, { status: 'disabled' }>;
```
rogaldh: nit: Maybe it is worth specifying `DisabledTotalRewardState` to exclude it by providing the exact shape via an alias

app/features/stake/ui/StakeAccountSection.tsx
```diff
@@ -237,6 +261,26 @@ function DelegationCard({
...
+    solPrice: number | null;
+}) {
+    switch (state.status) {
+        case 'loading':
```
rogaldh: nit: going further, we could use an enum variant here instead of a literal. To ensure each case is checked against the existing value

app/entities/stake-rewards/api/fetch-total-stake-reward.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Solscan answered, but not in a shape we can trust — their contract changed, not our logic.
+ *
+ * Separate from our own invariant failures (the paging bound, the safe-integer guard) because the
+ * two need different owners: this one is read-their-changelog-and-fix-the-parser, and until someone
+ * does, every stake account's Total Reward row quietly reads `Unavailable`.
+ */
+export class SolscanResponseError extends Error {
```
rogaldh: It is not stated anywhere, but several modules already have separate `error.ts` files that expose possible errors. As we define a specific constructor for the error, which is great, I think it is worth moving it into a separate file for clarity

app/entities/stake-rewards/api/fetch-total-stake-reward.ts
```diff
@@ -0,0 +1,162 @@
...
+    // a partial sum is wrong by an unknown amount and still looks authoritative.
+    const response = await fetchUpstream(`${SOLSCAN_BASE_URL}/account/stake/reward?${query}`, {
+        headers: { token: apiKey },
+    });
+
+    if (!response.ok) {
+        throw new SolscanRequestError(`Stake reward page ${page} failed`, { status: response.status });
+    }
+
+    return parseRewardPage(await response.json()).data;
```
rogaldh: If I understand correctly, https://github.com/solana-foundation/explorer/pull/1185/changes#diff-3f2acad9ad6045786f0b9ffe16d0a3bf508417c7e9fe3ad8bb1f4da037467b8dR20-R23 could return success: false. In that case, returning just the data hides possible errors.

app/features/stake/model/use-total-reward.ts
```diff
@@ -0,0 +1,69 @@
...
+    const { data, error, isLoading } = useSWR(
+        // A falsy key disables the fetch, so a disabled deployment never asks for the total.
+        isEnabled && isSupported && (['stake-total-reward', stakeAccountAddress] as const),
+        () => fetchTotalReward(stakeAccountAddress),
+        // Revalidating (rather than `useSWRImmutable`) keeps a long-lived tab from showing a total
+        // frozen from before the last epoch boundary. It costs no upstream quota: a revalidation
+        // inside the route's 4 h CDN window is served by the CDN and never reaches Solscan.
+        //
+        // fetchTotalReward throws rather than returning undefined, so cap the retries.
+        { errorRetryCount: 3 },
```
rogaldh: Potential improvement: here we retry every error, even one that does not make sense to retry, like a 400, for example. Maybe it would be worth adding a handler that decides which error to retry.

app/features/stake/model/use-total-reward.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+
+    if (!isEnabled) {
+        return { status: 'disabled' };
+    }
+    if (!isSupported || error) {
```
rogaldh: `|| error` here means that we may be stuck with an "unavailable" status for any error, even if not connected to the cluster support

rogaldh (changes requested):
Other than that - LGTM

app/features/stake/ui/StakeAccountSection.tsx
askov: DisabledTotalRewardState is now a named export

app/features/stake/ui/StakeAccountSection.tsx
askov: Added TotalRewardStatus

app/entities/stake-rewards/api/fetch-total-stake-reward.ts
askov: Sure, moved

app/entities/stake-rewards/api/fetch-total-stake-reward.ts
askov: I've never faced false with 200, but you are right that it's better to check anyway

app/features/stake/model/use-total-reward.ts
askov: Added retry only for 429, 502, 504

app/features/stake/model/use-total-reward.ts
askov: Split into Unsupported (cluster Solscan does not index) and Unavailable (a request that failed)

rogaldh (approved):
LGTM

PR #1189, opened 2026-08-12
fix(ci): build workspace packages before the feature-gates script

askov:
## Description

The daily `Update Feature Gates` workflow has failed on every run since 2026-08-06:

```
Error: Cannot find module '/home/runner/work/explorer/explorer/node_modules/@explorer/parsers/dist/compat/index.js'
```

The job installs dependencies but never builds the workspace packages. `@explorer/parsers` is a
`workspace:*` dependency whose `exports` map points at `./dist/*`, and in `node_modules` it is only a
symlink to `packages/parsers` — so `dist/` does not exist on a fresh CI checkout.

`scripts/update-feature-gates.ts` reaches that package transitively:

```
scripts/update-feature-gates.ts
  → scripts/feature-gates/lib/feature-store.ts        (value import, not type-only)
    → app/entities/feature-gate/server.ts
      → app/entities/feature-gate/lib/feature-gates-schema.ts
        → app/validators/pubkey.ts
          → export { PublicKeyFromString } from '@explorer/parsers/compat'
```

The schema only needs `AddressFromString`, but ESM evaluates all of `pubkey.ts`, so the unused
`@explorer/parsers/compat` re-export is still resolved at runtime and resolution fails.

This became a runtime dependency in #1175, which moved `PublicKeyFromString` into
`@explorer/parsers/compat` and left a re-export behind in `app/validators/pubkey.ts`. The last green
run was 2026-08-05, the day that PR merged; the first failure was the next cron run. It is invisible
locally because `dev`, `build`, `test`, and `typecheck` all run `pnpm build:packages` first, so
contributors always have a populated `dist/`.

The fix adds a `Build workspace packages` step between install and the script, mirroring what
`ci.yaml` already does for `scripts/update-sitemap.ts` for the same reason (`ci.yaml:133-135`).

`Update Verified Programs` is unaffected — its only app import (`app/utils/verified-builds-url`) does
not reach a workspace package, and it is currently green.

## Type of change

-   [x] Bug fix

## Testing

Reproduced and verified locally, since the failure only appears when `dist/` is absent as it is in CI:

1. `rm -rf packages/parsers/dist` → importing `scripts/feature-gates/lib/feature-store.ts` fails with
   the exact CI error above.
2. `pnpm build:packages` (the step this PR adds) → the same import resolves.

No preview links: the change touches only a workflow file, so there is no UI surface to check.

## Checklist

-   [x] All checks pass locally (`pnpm test`, `pnpm lint`, `pnpm exec tsc --noEmit`, `pnpm format`)
-   [x] I have run `build:info` script to update build information (no `bench/BUILD.md` diff — the change is workflow-only)

### DIFF
--- a/.github/workflows/update-feature-gates.yml
+++ b/.github/workflows/update-feature-gates.yml
@@ -31,6 +31,11 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      # scripts/update-feature-gates.ts imports app/entities/feature-gate/server, which reaches the
+      # built @explorer/parsers through app/validators/pubkey.ts.
+      - name: Build workspace packages
+        run: pnpm build:packages
+
       - name: Update feature gates
         run: pnpm exec tsx scripts/update-feature-gates.ts
 
--- a/app/api/stake-rewards/[address]/route.ts
+++ b/app/api/stake-rewards/[address]/route.ts
@@ -59,7 +59,7 @@ export async function GET(request: Request, props: Params) {
     try {
         // Gate the metered request behind one cheap RPC call: the route is public, and every cache
         // miss spends shared paid quota across up to ten upstream calls.
-        if (!(await isStakeAccount({ address, rpcUrl: serverClusterUrl(Cluster.MainnetBeta, '') }))) {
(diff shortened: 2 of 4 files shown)


rogaldh (dismissed):
LGTM

Woody4618 (approved):

PR #1191, opened 2026-08-12
feat(mcp): add landing page

rogaldh:
Adds an MCP landing page at `/mcp/start`: a public page that documents the Explorer's MCP endpoint and gets a client connected. The `POST /mcp` endpoint had no entry point in the app — no nav entry, no page, no link — so agent developers had no way to discover it.

- _Live status_ — the page reports whether the endpoint it documents actually answers, by polling it directly.
- _Setup snippets_ — copy-paste config for Claude Code, Cursor, VS Code, Codex and Windsurf, each checked against current vendor docs.
- _Enabled clusters_ — `app/mcp/clusters.ts` becomes the single source for which clusters `/mcp` serves, so the tool schema, the tool description and the page cannot disagree. Listing them rather than aliasing `SUPPORTED_CLUSTERS` makes a newly supported cluster opt-in here instead of going live with a package bump.
- _Hydration fix_ — the cluster modal rendered a localStorage-backed custom RPC URL into SSR'd markup, leaving anyone with a saved custom URL a stale link. Pre-existing; details in Additional Notes.
- _Docs_ — `packages/entity-inspector/TELEMETRY.md` records the GA4 dimensions that must be registered before the property can report on MCP usage; `app/mcp/README.md` gains the two `MCP_GA_*` variables it was missing.

Also adds a shared `CodeBlock` primitive, and wraps the two mcp docs at `printWidth` via a prettier override scoped to them.

## Type of change

-   [x] Bug fix
-   [x] New feature
-   [ ] Protocol integration
-   [x] Documentation update

## Screenshots

image

## Testing

Automated — `pnpm test`, `pnpm vitest --project storybook run`, `pnpm lint`, `pnpm typecheck`, all clean. New coverage:

- `BaseCodeBlock` / `CodeBlock` — caption and code rendering, no control without a handler, clipboard payload, and the per-state accessible name and live-region text.
- `setup-clients` — every snippet points at the canonical endpoint under the shared server key, and none contains `Authorization`/`Bearer`. That last one is a real regression guard: the operator README's snippets do carry a bearer header.
- `mcp-health` — `503` → `disabled`, non-ok → `unreachable`, network failure → `unreachable`, `pong` → `ready`, and both required `Accept` media types on the request.
- `clusters` — drift canary: the enabled set stays a subset of the inspector's `SUPPORTED_CLUSTERS`, contains the cluster the schema defaults to, and lists each cluster exactly once.
- `schemas` (package) — a cluster the deployment has not enabled is rejected, and `defaultCluster` falls back to the first enabled entry so `.default()` can never name a value its own enum rejects.
- `constants` — the advertised endpoint URL is absolute, since snippets are copied into other people's config files.
- `McpSetupTabs` / `McpLandingView` — tab switching, and that each hero CTA fragment matches a section id that exists.
- `useHydrated` — reports false on the first render, which is what makes the first client render match server markup.

Manual, against a local dev server:

- `MCP_ENDPOINT_ENABLED=true` → status block reaches `Ready` with a latency reading; flag off → `Disabled`, page still renders.
- `POST /mcp` and `OPTIONS /mcp` still answer 200/204, confirming the child segment does not disturb the handler.
- `inspect_entity` on the USDC mint returned `spl-token:mint` — that reply is what the Tools section shows, so the documented response shape is real rather than invented.
- Desktop (1280px) and mobile (390px): tab strip wraps, long commands wrap, JSON blocks scroll inside their own container, no page-level horizontal scroll, no console errors.

## Related Issues

[HOO-1088](https://linear.app/solana-fndn/issue/HOO-1088)

## Checklist

-   [x] Screenshots included — required for protocol screens, recommended for other UI changes

## Additional Notes

- The snippets always name the canonical `explorer.solana.com/mcp`, while the status block probes the current origin. Intentional: an agent should be pointed at production, but the reader wants to know whether the deployment they are looking at actually answers.
- Every visitor's health poll emits an `mcp_tool_call`, so MCP usage analytics will pick up non-agent traffic. Registering the `tool` dimension makes it filterable; a parameter that distinguishes it outright needs whoever owns the GA4 property. Volume rests on SWR defaults: hidden tabs do not poll, but refocusing one triggers a check on top of the 5-minute interval.
- A browser check cannot authenticate, so if `MCP_ACCESS_KEYS` is ever set, the block will report `unreachable` on a healthy endpoint. Fine under the current open-access setup, worth revisiting if that changes.
- `MCP_ENDPOINT_ENABLED` is inlined at build time for a statically prerendered page, so the seeded label follows deploys rather than live env edits. The first client-side check corrects it on load.
- The sitemap picked `/mcp/start` up automatically at the default priority; no `STATIC_OVERRIDES` entry was needed.
- `resolveRpcEndpoints` gives every cluster the package supports an entry whether or not `MCP_ENABLED_CLUSTERS` lists it, so enabling one is a change in `clusters.ts` alone.
- `enabledClusters` defaults to `SUPPORTED_CLUSTERS`, so nothing changes for package consumers that do not pass it.
- `pnpm format` runs `format:md` first so the main check stays last in the chain and `pnpm format <path>` still forwards its argument.
- The hydration fix is pre-existing, just hit while testing the page: `rememberedCustomUrlAtom` reads localStorage during render (`getOnInit: true`), so SSR emitted `localhost:8899` and the first client render disagreed. React leaves mismatched attributes stale, so the link kept pointing at localhost. `CustomClusterInput` holds the default for one render instead; the atom keeps `getOnInit`, which the param-strip effect needs.
- Left as out of scope: `Button`'s `ui="tw"` focus ring is `ring-neutral-950`, invisible on dark and unfixable from a call site because `cn` is clsx-only — hence `BaseCodeBlock`'s copy control carrying its own ring. Separately, `isCustomUrlAllowed` whitelist-checks the remembered URL rather than the param being honored, so the trusted-host allowance never fires on a first visit.


rogaldh:
@greptile-apps review pls

Woody4618 (approved):

PR #1192, opened 2026-08-12
feat(tx): read transaction details with kit so v1 renders

amilz:
Agave 4.2 Local Validator accepts [v1 transactions (SIMD-0385)](https://solana.com/upgrades/larger-transaction-sizes) so they can be tested today. The detail page fetched through web3.js `Connection` with `maxSupportedTransactionVersion: 0`, so a v1 transaction did not render at all — and raising that ceiling is not an option, because web3.js validates the response against `union([literal(0), literal('legacy')])` and throws on the first v1 the RPC returns.

This moves the transaction detail and raw fetch paths to `@solana/kit` (already a dependency at 6.5.0), which decodes v1 and exposes the message-level config.

## Approach

kit stops at the provider boundary. Roughly twenty components downstream are written against web3.js types, so rather than migrate them all, `adaptParsedTransaction` converts kit's response into the web3.js shape and consumers keep compiling. Migrating those consumers is the right long-term direction, but not as part of a v1 change.

The v1 resource limits come from decoding the base64 wire bytes, not from `message.transactionConfig` — kit's RPC types do not model that field, so reading it there would need a hand-written validator. Decoding goes through kit's typed decoder instead. It costs a second round trip, paid only by v1.

## Changes

- New `app/entities/transaction-data/` entity: `fetchTransactionDetails`, `fetchRawTransaction`, the `adaptParsedTransaction` boundary, and `decodeWireTransaction` / `decodeTransactionConfig`.
- Providers drop `Connection`. `Details.raw` gains `messageBytes` and `version`; `message` and `transaction` become optional, populated for legacy and v0 only.
- Download reads `messageBytes` rather than re-serializing, so it is byte-exact on every version.
- Summary card renders the v1 limits, labels the priority fee `Priority fee (total)` to distinguish it from v0's per-CU price, and shows versions as `legacy` / `v0` / `v1`.
- The Inspect page is gated with an explicit "not supported" card rather than adapted — its cards are built on `MessageV0`, and synthesizing one would make Download emit bytes that are not the transaction on chain.

## Notes for review

- `version` is absent from kit's `getTransaction` integer allow-list, so it arrives as a bigint while `GetTransactionApi` types it as a number. Every `version === 1` check is silently false. Converted in the adapter and tracked upstream in DEV-858.
- The version row now reads `v0` where it previously read `0`, so `v1` does not sit next to a bare `0`.
- Still on `maxSupportedTransactionVersion: 0`: the block page, address history, and the server-side receipt fetch. They omit v1 entries rather than erroring; each needs the same move and is worth its own change.

## Testing

`pnpm lint` and `pnpm typecheck` clean. 19 new unit tests covering the adapter and the config decoder. The 46 failing specs on this branch fail identically on `master` (localStorage and cluster-param suites) and are unrelated.

Here's a script for running v1 txns using Kit: https://gist.github.com/amilz/832fb14baf7dc2bdc5ec2f42d7c3f280

Closes DEV-856

### DIFF
--- a/app/components/instruction/InstructionCard.tsx
+++ b/app/components/instruction/InstructionCard.tsx
@@ -41,7 +41,7 @@ export function InstructionCard({
     // Use provided raw prop, or fetch from transaction details
     let raw: TransactionInstruction | undefined = rawProp;
     if (!raw && rawDetails && childIndex === undefined) {
-        raw = rawDetails?.data?.raw?.transaction.instructions[index];
+        raw = rawDetails?.data?.raw?.transaction?.instructions[index];
     }
 
     const fetchRaw = useFetchRawTransaction();
--- a/app/components/instruction/InstructionCard.tsx
+++ b/app/components/instruction/InstructionCard.tsx
@@ -1,4 +1,5 @@
 import { BaseInstructionCard } from '@components/common/BaseInstructionCard';
+import { FetchStatus } from '@providers/cache';
 import { useFetchRawTransaction, useRawTransactionDetails } from '@providers/transactions/raw';
 import { ParsedInstruction, SignatureResult, TransactionInstruction } from '@solana/web3.js';
 import React, { useCallback, useContext } from 'react';
@@ -47,8 +48,11 @@ export function InstructionCard({
     const fetchRaw = useFetchRawTransaction();
     const fetchRawTrigger = useCallback(() => fetchRaw(signature), [signature, fetchRaw]);
 
-    // Only allow fetching raw data if we have a valid signature (not in inspector mode)
-    const canFetchRaw = signature && !raw;
+    // Only allow fetching raw data if we have a valid signature (not in inspector mode), and only
+    // while a fetch could still produce it: a v1 transaction has no web3.js instruction view, so
+    // once its raw data has arrived, asking again would refetch on every open of the Raw view.
+    const rawUnavailable = rawDetails?.status === FetchStatus.Fetched && raw === undefined;
+    const canFetchRaw = signature && !raw && !rawUnavailable;
 
     return (
         <BaseInstructionCard
--- a/app/components/instruction/InstructionCard.tsx
+++ b/app/components/instruction/InstructionCard.tsx
@@ -51,8 +51,11 @@ export function InstructionCard({
     // Only allow fetching raw data if we have a valid signature (not in inspector mode), and only
     // while a fetch could still produce it: a v1 transaction has no web3.js instruction view, so
     // once its raw data has arrived, asking again would refetch on every open of the Raw view.
-    const rawUnavailable = rawDetails?.status === FetchStatus.Fetched && raw === undefined;
-    const canFetchRaw = signature && !raw && !rawUnavailable;
+    const rawFetched = rawDetails?.status === FetchStatus.Fetched;
+    const canFetchRaw = signature && !raw && !rawFetched;
+    // Inner instructions never carry raw wire data, so their Raw view is the same with or without
+    // it; only the top-level list has rows to lose.
+    const rawUnavailable = rawFetched && raw === undefined && childIndex === undefined;
 
     return (
         <BaseInstructionCard
@@ -66,6 +69,7 @@ export function InstructionCard({
             childIndex={childIndex}
             raw={raw}
             onRequestRaw={canFetchRaw ? fetchRawTrigger : undefined}
+            rawUnavailable={rawUnavailable}
             headerButtons={headerButtons}
             collapsible={collapsible}
         >
--- a/app/entities/transaction-data/api/fetch-raw-transaction.ts
+++ b/app/entities/transaction-data/api/fetch-raw-transaction.ts
@@ -2,9 +2,11 @@ import { createSolanaRpc, signature as createSignature } from '@solana/kit';
 import { type DecompileArgs, type Finality, PublicKey, TransactionMessage, VersionedMessage } from '@solana/web3.js';
 
 import { fromBase64 } from '@/app/shared/lib/bytes';
+import { Logger } from '@/app/shared/lib/logger';
 
-import { decodeWireTransaction } from '../lib/decode-wire-transaction';
-import type { RawTransaction, TransactionVersion } from '../model/types';
+import { decodeTransactionConfig } from '../lib/decode-transaction-config';
+import { decodeWireTransaction, type WireTransaction } from '../lib/decode-wire-transaction';
+import type { RawTransaction, TransactionConfig } from '../model/types';
 import { MAX_SUPPORTED_TRANSACTION_VERSION } from './max-supported-transaction-version';
 
 /**
@@ -36,10 +38,9 @@ export async function fetchRawTransaction(
 
     const [base64Transaction] = response.transaction;
     const { compiledMessage, messageBytes, signatures } = decodeWireTransaction(fromBase64(base64Transaction));
-    const version = compiledMessage.version as TransactionVersion;
     const meta = response.meta;
 
-    return {
+    const base = {
         messageBytes,
         meta: meta
             ? {
@@ -57,9 +58,32 @@ export async function fetchRawTransaction(
               }
             : undefined,
         signatures,
-        version,
-        ...(version === 1 ? undefined : decodeWithWeb3(messageBytes, meta?.loadedAddresses)),
+        transactionConfig: readTransactionConfig(compiledMessage, signature),
     };
+
+    if (compiledMessage.version === 1) {
+        return { ...base, version: 1 };
+    }
+
+    return {
+        ...base,
+        ...decodeWithWeb3(messageBytes, meta?.loadedAddresses),
+        version: compiledMessage.version,
+    };
+}
+
+// The resource-limit rows are supplemental; the wire bytes still render, download, and inspect
+// without them.
+function readTransactionConfig(
+    compiledMessage: WireTransaction['compiledMessage'],
+    signature: string,
+): TransactionConfig | undefined {
+    try {
+        return decodeTransactionConfig(compiledMessage);
+    } catch (error) {
+        Logger.error(error, { module: '[transaction-data]', signature });
+        return undefined;
+    }
 }
 
 type RpcLoadedAddresses = Readonly<{ readonly: readonly string[]; writable: readonly string[] }>;
@@ -67,7 +91,7 @@ type RpcLoadedAddresses = Readonly<{ readonly: readonly string[]; writable: read
 function decodeWithWeb3(
     messageBytes: Uint8Array,
     loadedAddresses: RpcLoadedAddresses | undefined,
-): Pick<RawTransaction, 'message' | 'transaction'> {
+): { message: VersionedMessage; transaction: TransactionMessage } {
     const message = VersionedMessage.deserialize(messageBytes);
     const accountKeysFromLookups = loadedAddresses && {
         readonly: loadedAddresses.readonly.map(address => new PublicKey(address)),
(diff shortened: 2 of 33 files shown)


app/entities/transaction-data/api/fetch-transaction-details.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+async function fetchTransactionConfig(url: string, signature: string): Promise<TransactionConfig | undefined> {
+    const response = await createSolanaRpc(url)
+        .getTransaction(createSignature(signature), {
+            commitment: 'confirmed',
+            encoding: 'base64',
+            maxSupportedTransactionVersion: MAX_SUPPORTED_TRANSACTION_VERSION,
+        })
+        .send();
```
askov: **issue**: greptile is correct, this should be guarded: any fail here (like 429) leads to "Failed to fetch details". Needs a try-catch with a sensible handler

app/components/instruction/InstructionCard.tsx
```diff
@@ -41,7 +41,7 @@ export function InstructionCard({
     // Use provided raw prop, or fetch from transaction details
     let raw: TransactionInstruction | undefined = rawProp;
     if (!raw && rawDetails && childIndex === undefined) {
-        raw = rawDetails?.data?.raw?.transaction.instructions[index];
+        raw = rawDetails?.data?.raw?.transaction?.instructions[index];
```
askov, line 45: **issue**: For v1, `raw?.transaction?.instructions[index]` is permanently undefined, so canFetchRaw never goes false. rawClickHandler fires onRequestRaw() on every open of the Raw view - a fresh getTransaction plus a re-render each time. Line 111 then renders only the JSON row; the program-account and raw-data rows disappear. The new ?. made this type-safe but not behaviour-safe. Gate canFetchRaw on fetched-but-unrepresentable.

app/entities/transaction-data/api/fetch-transaction-details.ts
```diff
@@ -0,0 +1,62 @@
...
+        // The cache providers distinguish "fetched, not found" from "not fetched yet" by null vs undefined.
+        // eslint-disable-next-line unicorn/no-null
+        return null;
+    }
+
+    const transactionWithMeta = adaptParsedTransaction(response);
+    if (transactionWithMeta.version !== 1) {
+        return transactionWithMeta;
+    }
+
```
askov: **suggestion**: SummaryCard already calls fetchRaw, which requests the same base64 encoding and already runs decodeWireTransaction to get compiledMessage - precisely what fetchTransactionConfig re-fetches and re-decodes. So it seems it could decode the config from the message you already have without extra rpc call

app/entities/transaction-data/api/max-supported-transaction-version.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * The newest transaction version Explorer asks the RPC for.
+ *
+ * Without it the RPC answers `-32015` for anything newer than legacy and the transaction fails to
+ * load at all. Nodes that predate v1 compare this ceiling against the versions they actually hold,
+ * so requesting 1 is safe against them.
+ */
+export const MAX_SUPPORTED_TRANSACTION_VERSION = 1;
```
askov: **question**: Can we just import it from kit?

app/entities/transaction-data/lib/__tests__/adapt-parsed-transaction.spec.ts
```diff
@@ -0,0 +1,191 @@
+import { describe, expect, it } from 'vitest';
+
+import { adaptParsedTransaction, type RpcParsedTransaction } from '../adapt-parsed-transaction';
+
+const FEE_PAYER = '11111111111111111111111111111112';
```
askov: **nit**: We have a special set of utilities to not hardcode addresses in test (see fixtures gen.* helpers)

app/entities/transaction-data/model/types.ts
```diff
@@ -0,0 +1,60 @@
...
+    TransactionMessage,
+    VersionedMessage,
+} from '@solana/web3.js';
+
+/**
+ * Transaction message versions Explorer can render.
+ *
+ * web3.js stops at v0, so its `TransactionVersion` cannot describe a v1 transaction.
+ */
+export type TransactionVersion = 'legacy' | 0 | 1;
```
askov: **question**: Can we just import it from kit?

app/entities/transaction-data/api/fetch-transaction-details.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (response === null) {
+        return undefined;
+    }
+
+    const [base64Transaction] = response.transaction;
+    try {
+        return decodeTransactionConfig(decodeWireTransaction(fromBase64(base64Transaction)).compiledMessage);
+    } catch {
+        // The resource-limit rows are supplemental; the rest of the page renders without them.
```
askov: **suggestion**: It would be nice to have Logger.error here at least

app/entities/transaction-data/model/types.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+/**
+ * A transaction's wire bytes and the parts of its metadata the inspector needs.
+ *
+ * `message` and `transaction` are the web3.js views of the bytes and are absent on v1, whose
+ * message web3.js cannot represent. `messageBytes` is always present, so anything that only
+ * needs the bytes — the download button — works on every version.
+ */
+export type RawTransaction = {
```
askov: **suggestion/issue**: I would highly recommend to have this as a discriminated union. Hard to read/understand in its current state, something like:
```ts
type RawTransactionBase = {
    messageBytes: Uint8Array;
    meta?: { innerInstructions?: CompiledInnerInstruction[]; postBalances: number[]; preBalances: number[] };
    signatures: (string | undefined)[];
};

export type RawTransaction = RawTransactionBase &
    (
        | { version: 'legacy' | 0; message: VersionedMessage; transaction: TransactionMessage }
        | { version: 1; message?: undefined; transaction?: undefined }
    );
```

app/entities/transaction-data/api/fetch-raw-transaction.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Fetches a transaction's wire bytes and metadata for the inspector and the download button.
+ *
+ * `base64` encoding rather than `json` so the bytes are the ones the network holds: they are what
+ * the download button writes, and the only representation a v1 transaction has here — web3.js
+ * `VersionedMessage` tops out at v0, so `message` and `transaction` are populated for legacy and
+ * v0 only.
+ */
+export async function fetchRawTransaction(
```
askov, line 19: **suggestion**: It would be useful to have tests for the new api things (fetch-transaction-details, fetch-raw-transaction), they are not that simple to leave untested (for example, the main issue with try-catch for send would have been caught). get-transactions-for-address could be a reference.

app/entities/transaction-data/lib/decode-wire-transaction.ts
```diff
@@ -0,0 +1,40 @@
...
+    const base58Decoder = getBase58Decoder();
+
+    return {
+        compiledMessage: getCompiledTransactionMessageDecoder().decode(transaction.messageBytes),
+        // The decoded message bytes are a view over `bytes`; copy so callers own their buffer.
+        messageBytes: new Uint8Array(transaction.messageBytes),
+        // SignaturesMap is insertion-ordered, matching the message's signer order. A signer that has
+        // not signed yet gets the all-zero signature, which renders as a placeholder rather than a gap.
+        signatures: Object.values(transaction.signatures).map(signature =>
+            base58Decoder.decode(signature ?? new Uint8Array(SIGNATURE_BYTES)),
```
askov: **suggestion/issue**: Unsigned signer slots now render a false "Invalid" badge, could use:

```ts
signatures: Object.values(transaction.signatures).map(signature =>
    signature ? base58Decoder.decode(signature) : undefined,
),
```

askov (changes requested):

Woody4618:
Tested using https://gist.github.com/amilz/832fb14baf7dc2bdc5ec2f42d7c3f280 

image

app/entities/transaction-data/api/fetch-transaction-details.ts
```diff
@@ -0,0 +1,62 @@
...
+        // eslint-disable-next-line unicorn/no-null
+        return null;
+    }
+
+    const transactionWithMeta = adaptParsedTransaction(response);
+    if (transactionWithMeta.version !== 1) {
+        return transactionWithMeta;
+    }
+
+    return { ...transactionWithMeta, transactionConfig: await fetchTransactionConfig(url, signature) };
```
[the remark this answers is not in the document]
amilz: Fixed in f945d531 — the supplemental fetch is gone entirely; the config now decodes from the bytes `fetchRawTransaction` already retrieves.

app/entities/transaction-data/api/fetch-transaction-details.ts
amilz: Fixed in f945d531 — no supplemental call left to fail; the surviving decode is guarded with `Logger.error`.

app/entities/transaction-data/api/fetch-transaction-details.ts
amilz: Good call — done in f945d531. `transactionConfig` rides the existing base64 response, so v1 is back to no extra round trip.

app/entities/transaction-data/api/max-supported-transaction-version.ts
amilz: You were right — kit re-exports it from `@solana/transaction-messages`; local module deleted in d84bf130.

app/entities/transaction-data/model/types.ts
amilz: Yes — now imported from kit in f945d531.

app/entities/transaction-data/api/fetch-transaction-details.ts
amilz: Added in f945d531, on the config decode in `fetch-raw-transaction.ts`.

app/entities/transaction-data/model/types.ts
amilz: Adopted your shape in f945d531, with `transactionConfig` moved onto the v1 arm in 18977e9e.

app/entities/transaction-data/lib/__tests__/adapt-parsed-transaction.spec.ts
amilz: My bad! thanks. Switched to `gen.*` in f945d531, and in the new wire-transaction fixtures in 18977e9e.

app/entities/transaction-data/api/fetch-raw-transaction.ts
amilz: Added in f945d531 — specs for both fetchers against a stubbed `fetch`, following `get-transactions-for-address`; one pins the single round trip.

app/entities/transaction-data/lib/decode-wire-transaction.ts
amilz: Took your snippet in f945d531 — unsigned slots are `undefined` now, so they render as Missing rather than Invalid.

app/components/instruction/InstructionCard.tsx
amilz: Refetch gate landed in f945d531; for the disappearing rows, take a look at ab7aa23d — I think that's what you're after, lmk. (Inner instructions never carry raw data, so they make no version claim.)

askov (approved):

PR #1201, opened 2026-08-14
refactor(deps): decode legacy layouts with kit codecs

amilz:
## Description

Replaces `@solana/buffer-layout` 3.0.0 with `@solana/kit` decoders and drops the dependency. It was used in exactly three modules — the feature account, the NFToken accounts, and the Pyth instruction layouts — and every construct in use has a direct kit equivalent, so no bridge or compat shim is needed. The two hand-written `Layout` subclasses (`LPString`, `GreedyList`) become `addDecoderSizePrefix` and `getArrayDecoder(..., { size: 'remainder' })`.

Porting surfaced two live bugs the old layouts had been hiding:

- **The NFToken NFT struct was off by one byte.** It declared three unused padding bytes where the account has four, so every field from `metadata_url_length` onward was read one byte early. The length came out as 17152 instead of 67 (unused downstream, so invisible) and `metadata_url` began one byte early with a stray leading `NUL` — which the `.replace(/\0/g, '')` that existed to strip the account's *trailing* padding also removed. The URL was correct by accident. `buffer-layout`'s `utf8(maxSpan)` reads greedily to the end of the buffer, which is what let a mis-declared field be absorbed silently; a size-prefixed kit decoder reads exactly the declared bytes and fails otherwise, so it surfaced immediately.
- **Address fields were rendering as raw bytes.** `blob(32)` yields a `Buffer`, and five call sites assigned it straight into `string`-typed fields — `authority` on the collection card, and all four of `authority`/`collection`/`delegate`/`holder` in the collection-grid fetcher. `getAddressDecoder()` produces the base58 strings the types already promised.

Kit decoders read exact byte counts and throw where `buffer-layout` clamped, so two call sites needed hardening: the NFToken discriminator checks move *ahead* of the decode (the NFT and collection accounts share an owner and a discriminator prefix but not a size, so the NFT parser would otherwise throw and log on every render of any collection account), and `getNftsInCollection` now guards each account so one undecodable entry drops itself instead of emptying the whole grid.

Kit-migration task 0.4. Also included as a separate commit: deleting the stale `.pnpmfile.cjs` (task 0.3), whose rewrites targeted `@solana-program/compute-budget` 0.6.1 and `lighthouse-sdk` 2.0.1 — we install 0.15.0 and 2.1.0, so it was a confirmed no-op.

## Type of change

-   [x] Bug fix
-   [x] Other: dependency removal / migration groundwork

## Testing

-   Full local CI green on Node 22: `format:ci`, `lint`, `openspec:validate` (17/17), `build`.
-   `test:ci`: 4070 passing, 3 failures — all in `app/og/receipt/[signature]/__tests__/route.spec.ts`, whose mocks broke in master's Vitest 4 migration (#1122). Nothing in this diff is imported by that route. This branch adds 12 passing tests.
-   New `app/components/instruction/pyth/__tests__/program.test.ts` — the Pyth decoders had no coverage at all. Covers header/type resolution, the version and index-mismatch guards, address decoding, a negative `i32` exponent, the `bigint`→`number` narrowing, and the remainder-sized attribute list.
-   `isNFTokenAccount-test.ts` renamed to `isNFTokenAccount.test.ts`: vitest's default include matches `*.test.*` / `*.spec.*`, so under its old name this file had **never run**. It passes as written, and now runs. Added cases for the collection path, the collection-account-to-NFT-parser rejection, and short data.
-   `pnpm why @solana/buffer-layout -r` reports no 3.0.0 resolution afterward. The 4.0.1 line under `@solana/web3.js` is untouched, and nothing declares a peer on 3.x — so the peer re-resolution trap from #1200 does not apply here.
-   Bundle impact is small by design: buffer-layout 3.0.0's only dependency is `buffer`, which stays (36 other paths pull it via web3.js, buffer-layout 4.0.1, and rpc-websockets). Measured 8224 KB of client JS against 8244 KB recorded on the base. The value of this change is the dependency removal and the two fixes above, not the KB.

## Related Issues

Closes DEV-892
Closes DEV-890

## Additional Notes

Rebased onto `master` now that #1200 has landed, so the diff is just this change. Per review feedback the OpenSpec proposal was dropped — `openspec/README.md` exempts dependency removals and mechanical refactors.

Deliberately out of scope: `pyth/program.ts` still takes web3.js `TransactionInstruction` and returns `PublicKey` params (kit-migration task 2.4 rewrites those cards onto `KitInstruction`), and `nftoken.ts` still builds a `new Connection` (task 1.8).


app/components/account/nftoken/nftoken.ts
C0mberry: nit: i know that you didn't change this code, but i see two sortings. can we remove it?

C0mberry (approved):

openspec/changes/decode-legacy-layouts-with-kit-codecs/proposal.md
```diff
@@ -0,0 +1,51 @@
+# Proposal: Decode legacy account and instruction layouts with kit codecs
```
Woody4618: Do we really need these?

openspec/changes/decode-legacy-layouts-with-kit-codecs/specs/legacy-layout-decoding/spec.md
```diff
@@ -0,0 +1,80 @@
+## ADDED Requirements
```
Woody4618: Do we really need this?

Woody4618:
Hey, could you please merge the merge conflicts and check if we really need that proposal.md to remove a legacy component. Looks a bit like a big file for smth not 100% needed imo

amilz:
Both addressed — thanks @Woody4618.

**Conflicts:** rebased onto `master` now that #1200 has landed. 
**proposal.md:** agreed, dropped it (and its spec delta).

Woody4618 (approved):

PR #1204, opened 2026-08-14
feat(inspector): support v1 transactions via a kit message bridge

amilz:
## Summary
- Un-gates v1 transactions in the inspector across all three entry paths: permalink fetch, `?message=` URL param, and pasted raw input.
- New `app/shared/lib/v1-message-bridge.ts`: decodes v1 message bytes with kit 7 and exposes them as a `MessageV0` view whose `serialize()` returns the **original v1 wire bytes** — so simulation, signature verification, size math, downloads, and cache fingerprints all operate on the bytes the network sees, never a v0 re-encoding. This closes the serialize hazard that motivated the original gate.
- The overview card renders the v1 version row and the message-level resource limits (compute unit limit, total priority fee, loaded accounts data size limit, heap size), mirroring the detail page. The address table lookups card is hidden for v1 — kit 7's `V1CompiledTransactionMessage` carries static accounts only, with no lookup support.
- Zero behavior change for legacy/v0: the bridge sits behind a wire-prefix check (`0x81`) that only matches bytes which previously failed to decode.

Stacked on #1193 (kit 7 introspection); retarget to `master` once that merges. The openspec change records why the bridge was chosen over re-rooting the inspector on kit's `CompiledTransactionMessage`, which remains the deferred migration destination and deletes this shim when it lands.

## Test Plan
- `app/shared/lib/__tests__/v1-message-bridge.spec.ts` — field mapping, config extraction, and the regression that matters: `serialize()` is byte-equal to the input and `new VersionedTransaction(view).serialize()` round-trips through kit's transaction decoder to the same v1 message bytes.
- `PermalinkView.spec.tsx` — a v1 cache entry renders the loaded view with config rows and no lookups card; undecodable v1 bytes keep the error card.
- `InspectorPage-V1.spec.tsx` — full-page render from a v1 `?message=` param with real providers.
- Full suite shows the same 46 pre-existing failures across the same 7 files as the base branch (documented in the #1193 openspec).
- Note: whether an RPC accepts a v1 `simulateTransaction` depends on cluster runtime support; a rejection surfaces through the simulator's existing error card.

Closes DEV-898

### DIFF
--- a/app/components/inspector/InspectorPage.tsx
+++ b/app/components/inspector/InspectorPage.tsx
@@ -508,16 +508,17 @@ export function PermalinkView({
 
     const { message, messageBytes, signatures, meta } = transaction;
     // The inspector renders a web3.js `VersionedMessage`; a v1 message gets there through a
-    // bridged view over the wire bytes.
-    let bridgedMessage: VersionedMessage | undefined;
+    // bridged view over the wire bytes, which also carries the message's resource limits so
+    // every entry path derives them from the same decode.
+    let bridged: ReturnType<typeof bridgeV1MessageBytes> | undefined;
     if (!message && transaction.version === 1) {
         try {
-            bridgedMessage = bridgeV1MessageBytes(messageBytes).message;
+            bridged = bridgeV1MessageBytes(messageBytes);
         } catch (_err) {
             // Fall through to the error card below.
         }
     }
-    const resolvedMessage = message ?? bridgedMessage;
+    const resolvedMessage = message ?? bridged?.message;
     if (!resolvedMessage) {
         return (
             <ErrorCard
@@ -534,9 +535,7 @@ export function PermalinkView({
         message: resolvedMessage,
         rawMessage: messageBytes,
         signatures,
-        ...(transaction.version === 1
-            ? { transactionConfig: transaction.transactionConfig, version: 1 as const }
-            : undefined),
+        ...(bridged ? { transactionConfig: bridged.transactionConfig, version: 1 as const } : undefined),
     };
     return <LoadedView transaction={tx} onClear={reset} showTokenBalanceChanges={showTokenBalanceChanges} />;
 }
--- a/app/components/inspector/InspectorPage.tsx
+++ b/app/components/inspector/InspectorPage.tsx
@@ -29,6 +29,7 @@ import { Button } from '@/app/components/shared/ui/button';
 import { useCluster } from '@/app/providers/cluster';
 import { DownloadDropdown } from '@/app/shared/components/DownloadDropdown';
 import { toBase64 } from '@/app/shared/lib/bytes';
+import { bridgeV1MessageBytes, isV1MessageBytes, type V1TransactionConfig } from '@/app/shared/lib/v1-message-bridge';
 import { Card, CardHeader, CardTitle } from '@/app/shared/ui/Card';
 import { PageContainer } from '@/app/shared/ui/page-container/PageContainer';
 import { BaseTable } from '@/app/shared/ui/Table';
@@ -71,6 +72,13 @@ export function vaultMessageToVersionedMessage(message: typeof VaultTransaction.
 export type TransactionData = {
     rawMessage: Uint8Array;
     message: VersionedMessage;
+    /**
+     * Set when `rawMessage` holds a v1 message. `message` is then a bridged view whose
+     * `version` getter still reports 0, so version-dependent rendering must read this field.
+     */
+    version?: 1;
+    /** Message-level resource limits; v1 only, and only when the message sets at least one. */
+    transactionConfig?: V1TransactionConfig;
     signatures?: (string | undefined)[];
     accountBalances?: {
         preBalances: number[];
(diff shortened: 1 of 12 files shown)


app/components/inspector/InspectorPage.tsx
```diff
@@ -489,23 +502,36 @@ export function PermalinkView({
...
 
     const { message, messageBytes, signatures, meta } = transaction;
-    // The inspector renders a web3.js `VersionedMessage`, which cannot represent a v1 message.
-    if (!message) {
+    // The inspector renders a web3.js `VersionedMessage`; a v1 message gets there through a
+    // bridged view over the wire bytes.
+    let bridgedMessage: VersionedMessage | undefined;
+    if (!message && transaction.version === 1) {
+        try {
+            bridgedMessage = bridgeV1MessageBytes(messageBytes).message;
```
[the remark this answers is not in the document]
amilz: Addressed in 006d2b90 — `PermalinkView` now keeps the full `bridgeV1MessageBytes` result and reads `transactionConfig` from it, so all three entry paths derive the resource limits from the same decode. The cached field stays on `RawTransaction` for the detail page's summary card, which is its remaining consumer. The PermalinkView spec now omits `transactionConfig` from the cache entry to pin that the rows render from the message bytes.

app/components/inspector/__tests__/InspectorPage-V1.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+        expect(await screen.findByText('Transaction Overview')).toBeInTheDocument();
+        expect(screen.queryByText('Inspector Input')).toBeNull();
+        expect(screen.getByText('v1')).toBeInTheDocument();
+        expect(screen.getByText('Compute unit limit')).toBeInTheDocument();
+        expect(screen.getByText('300,000')).toBeInTheDocument();
+        expect(screen.getByText('Priority fee (total)')).toBeInTheDocument();
+        // fee payer, recipient, program
+        expect(screen.getByText('Account List (3)')).toBeInTheDocument();
+        // v1 messages carry static accounts only, so the lookups card is not rendered.
+        expect(screen.queryByText('Address Table Lookups')).toBeNull();
```
askov: **todo**: Please check that this assert really does the work, because the card name is "Address Table Lookup(s)"

app/shared/lib/v1-message-bridge.ts
```diff
@@ -0,0 +1,151 @@
...
+
+import { Logger } from '@/app/shared/lib/logger';
+
+/**
+ * Message-level resource limits carried by a v1 transaction.
+ *
+ * kit models this as `V1TransactionConfig` but does not export that type by name, so it is read
+ * off the v1 arm of kit's `TransactionMessage` to stay in step with kit's definition.
+ */
+export type V1TransactionConfig = NonNullable<Extract<KitTransactionMessage, { version: 1 }>['config']>;
```
askov, line 23: **suggestion**: TransactionConfig and readTransactionConfig are already presented in transaction-data entity with different names. So, maybe move it to entity (or use shared helper in the entity) to not duplicate.

askov (dismissed):

app/components/inspector/__tests__/InspectorPage-V1.spec.tsx
amilz: Good catch — the card renders `Address Table Lookup(s)`, so the old string could never match and the assertion was vacuous. `AddressTableLookupsCard` now exports `ADDRESS_TABLE_LOOKUPS_CARD_TITLE` and both the card and the spec use it, so the query cannot drift from the title again. The spec also asserts the `Address Table Lookup` account badge is absent, which is the other place a lookup-derived account would show up.

app/shared/lib/v1-message-bridge.ts
amilz: Agreed. Deduped in the direction FSD layering allows: `readV1TransactionConfig` and `V1TransactionConfig` are now exported from `app/shared/lib/v1-message-bridge.ts`, the entity keeps `TransactionConfig` as an alias of that type for its public API, and `fetch-raw-transaction` calls the shared reader (passing its own `{ module, signature }` log context) instead of its local copy.

app/shared/lib/v1-message-bridge.ts
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * The wire version prefix of a v1 transaction message: the version flag bit (0x80) with
+ * version number 1. Legacy messages never set the flag bit, and a v0 message's first byte
+ * is 0x80, so this single byte identifies v1 unambiguously.
+ */
+const V1_MESSAGE_PREFIX = 0x81;
```
Woody4618, line 30: Does this really need to be a constant here in explorer? Can we move that to some other package?

app/shared/lib/v1-message-bridge.ts
```diff
@@ -0,0 +1,171 @@
...
+    );
+
+    return { message, transactionConfig: readV1TransactionConfig(compiled) };
+}
+
+/**
+ * The maximum size of a v1 transaction in bytes. kit defines the same constant but
+ * intentionally does not export it.
+ */
+export const V1_TRANSACTION_SIZE_LIMIT = 4096;
```
Woody4618, line 114: Does this really need to be a constant here in explorer? Can we move that to some other package?

Woody4618 (approved):

app/shared/lib/v1-message-bridge.ts
Woody4618: The whole v1 message bridge could maybe move somewhere else?

amilz: hopefully most of this will go away when we do kit sprint. this was in the spirit of adding support w/o breaking/refactoring everything existing. we'll do that later. @Woody4618

app/shared/lib/v1-message-bridge.ts
amilz: It's not exported from kit atm.

Issue #458, opened 2025-02-07
search priority for programs and domains


nickfrosty:
When searching for anything inside the Explorer search box, we should force alter the order of results that get displayed to the user. Specifically to show programs and domain results first before tokens.

So the end goal of order should be:
- domains
- programs
- tokens

ngundotra:
Fixed in #578

PR #457, opened 2025-02-06
Add support for emit-based metadata on t22 tokens

ngundotra:
Also rips out the need for respecting token metadata pointer. 

This change however, by default assumes metadata pointer is pointing to mint. 

So on a conflict of extensions (metadataPointer points away from mint, but `MetadataExtension` is set, `MetadataExtension` takes higher priority).

Spec design for `emit` here: https://forum.solana.com/t/srfc-00017-token-metadata-interface/283

PR #1213, opened 2026-08-17
refactor(wallet): replace Solana Wallet Adapter with the Kit wallet plugin

amilz:
## Summary

- Removes `@solana/wallet-adapter-*`, the only runtime path to Dependabot [#457](https://github.com/solana-foundation/explorer/security/dependabot/457) and [#458](https://github.com/solana-foundation/explorer/security/dependabot/458) (high-severity DoS in `image-size`, no patched version). It arrived via `@solana/wallet-adapter-react` → `@solana-mobile/wallet-adapter-mobile` → `react-native` → `metro`. After this PR `pnpm why image-size` shows only a devDependency path through Storybook.
- Wallet discovery, connection, and signing move to `@solana/kit-plugin-wallet` + `@solana/react` 7.0.0 (`walletSigner()` behind `ClientProvider`). `@solana/kit` stays at 7.0.0 — `@solana/react` 7.0.0 peers that exact version.
- Transaction building and broadcasting stay on web3.js, so the change is confined to the wallet seam. `sign-web3js-transaction.ts` bridges Kit signers to web3.js transactions through the wire format; `use-wallet.ts` keeps the four interactive-IDL consumers on a single import. Both are written to be deleted when transaction building goes Kit-native, not extended.
- `WalletModalProvider`'s modal is replaced by an in-repo `WalletPickerDialog`; `BaseConnectWallet` is unchanged.

Wallet state is mounted in exactly one place (`IdlSection.tsx`), so the blast radius is the interactive-IDL tab on program pages.

## Notable details

- **Signatures are rebuilt from the wallet's returned bytes.** Wallets exposing `solana:signTransaction` surface as `TransactionModifyingSigner`s and may rewrite the message, so copying a signature onto the original transaction would sign bytes we don't broadcast. The wire-format round trip also carries versioned transactions, which Anchor's wallet interface requires.
- **`disposablestack` is added.** `@solana/plugin-core` registers plugin cleanup through `DisposableStack`, which Next.js doesn't polyfill and browsers only gained in 2025; without it the IDL section throws on browsers older than Chrome/Edge 134, Firefox 134, Safari 18.4.
- **Custom and SIMD-296 clusters map to `solana:localnet`.** The chain identifier is sent with every signature request and drives which network the wallet simulates against, so an arbitrary endpoint must not be labelled mainnet. Trade-off: wallets that don't advertise `solana:localnet` aren't offered on those clusters.
- **A watch-only wallet reports as not connected**, so Execute stays disabled rather than enabling and failing at signing time. Such a wallet couldn't have executed under wallet-adapter either.
- **`skipToast` is removed** from `WalletProvider`. The Kit plugin rejects the action promise instead of calling `onError`, and the sole call site already passed `skipToast`, so no toast was reachable. Failures now go to `Logger.error`.
- `publicKey` is `undefined` rather than `null` when disconnected, so `isEnabled`, `createGetAutocompleteItems`, and `createWalletPrefillDependency` widen to match.

Rationale and rejected alternatives are recorded in `openspec/changes/connect-wallets-with-kit-plugin/`.

## Test Plan

- `pnpm format:ci`, `pnpm lint`, `pnpm openspec:validate`, `pnpm build` — clean
- `pnpm test:ci` — 360 files, 3867 passed, 0 type errors
- `pnpm test:sb` — 322 files, 1182 passed
- `pnpm why image-size` — no runtime path; `pnpm why @solana/wallet-adapter-base -r` and `pnpm why react-native -r` — empty
- Client chunks 8716 → 8712 KB
- New specs cover the signing bridge (legacy and versioned round trips, sign-and-send-only rejection, empty-batch short circuit) and the cluster-to-chain mapping

Manual verification needs `NEXT_PUBLIC_INTERACTIVE_IDL_ENABLED=true` and a real RPC in `.env.local` (the public mainnet endpoint 403s the browser, and `modifyUrl` rewrites server-side calls to `explorer-api.*`). Program IDL card → Interact tab on e.g. `whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc`. Wallet discovery and the picker are confirmed working; auto-reconnect, disconnect, cluster switching, and a devnet execute are still worth a reviewer pass.

Closes HOO-1128


openspec/changes/connect-wallets-with-kit-plugin/proposal.md
```diff
@@ -0,0 +1,48 @@
+# Proposal: Connect wallets through the Kit wallet plugin instead of Solana Wallet Adapter
+
```
askov, line 2: **thought**: Hm, I'm not sure we need an open spec doc for every change. I do not insist on not having it here, but I wouldn't create it for this change. Or maybe trim the open spec docs to make them really short. Right now they are out of sync with the work: see disposablestack and confirmBroadcastTransaction, for example

app/features/idl/interactive-idl/model/transaction/use-execute-transaction.ts
```diff
@@ -110,6 +112,57 @@ export function useExecuteTransaction(opts: {
...
+    signature: string,
+    signed: Transaction,
+    requested: { blockhash: string; lastValidBlockHeight: number },
+): Promise<RpcResponseAndContext<SignatureResult>> {
+    const blockhash = signed.recentBlockhash;
+    if (!blockhash || blockhash === requested.blockhash) {
+        return connection.confirmTransaction({ ...requested, signature }, commitment);
+    }
+
+    const lastValidBlockHeight = (await connection.getBlockHeight()) + MAX_BLOCKHASH_AGE_IN_BLOCKS;
```
askov: **issue**: If this fails the already broadcast transaction gets reported as failed. Fix + test case would be great

openspec/changes/connect-wallets-with-kit-plugin/proposal.md
amilz: cut this quite a bit. might be worth just adding a little more context to AGENT.md for lvl of depth

app/features/idl/interactive-idl/model/transaction/use-execute-transaction.ts
amilz: i'm just going to revert back to the legacy confirmation strategy. updating confirm strategy is arguably out of scope for this pr. will update to @solana/transaction-confirmation as a part of the kit migration.

app/features/idl/interactive-idl/model/transaction/use-execute-transaction.ts
```diff
@@ -110,6 +112,77 @@ export function useExecuteTransaction(opts: {
...
+ * that height cannot be read, no bound is imposed and the validity poll ends the wait on its own:
+ * every height known before signing may already have passed while the wallet held the prompt, and
+ * confirming against one would expire a transaction that can still land.
+ */
+async function substitutedBlockhashDeadline(connection: Connection, signature: string): Promise<number> {
+    try {
+        return (await connection.getBlockHeight()) + MAX_BLOCKHASH_AGE_IN_BLOCKS;
+    } catch (error) {
+        Logger.error(error, { signature });
+        return UNBOUNDED_BLOCK_HEIGHT;
```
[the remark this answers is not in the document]
amilz: @greptile-apps removed confirmation strategy from scope of this PR, will address in another pr later. for now reverting `use-execute-transaction.ts` to it's previous state

askov:
@greptile-apps please review again

app/providers/wallet/use-wallet.ts
```diff
@@ -0,0 +1,53 @@
...
+    const status = useWalletStatus(client);
+    // `dispatch` rather than `dispatchAsync`: a superseded call rejects with an AbortError even
+    // though the newer one succeeded, so awaiting it would report a double-click as a failure.
+    const { dispatch: disconnect, error: disconnectError } = useDisconnect(client);
+    useLoggedWalletError(disconnectError);
+
+    const signer = connectedWallet?.signer;
+
+    const publicKey = useMemo(
+        () => (connectedWallet ? new PublicKey(connectedWallet.account.address) : undefined),
```
askov: **suggestion**: It would be nice to have it guarded with try/catch to not crash idl tab. Maybe also use account.address as a key. Tests for the hook would be great

app/providers/wallet/use-logged-wallet-error.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Reports a wallet action's error state.
+ *
+ * The Kit wallet plugin has no `onError` callback — failures land on the action hook's `error`
+ * instead, so something has to watch it for the failure to reach logging at all.
+ */
+export function useLoggedWalletError(error: unknown) {
+    useEffect(() => {
+        if (error) {
+            Logger.error(error);
```
askov: **thought**: Maybe send it to sentry? Not sure

app/providers/wallet/use-wallet.ts
```diff
@@ -0,0 +1,53 @@
...
+            signAllTransactions: <T extends Web3Transaction>(transactions: T[]) =>
+                signWeb3jsTransactions(signer, transactions),
+            signTransaction: <T extends Web3Transaction>(transaction: T) => signWeb3jsTransaction(signer, transaction),
+        };
+    }, [signer]);
+
+    return {
+        // A watch-only wallet connects without a signer. Reporting it as connected would enable
+        // Execute, which then fails at signing time with a misleading "wallet not connected".
+        connected: status === 'connected' && Boolean(signer),
```
askov: **suggestion**: connected conflates "connected" with "can sign", and handleConnect calls disconnect() only when connected is true - so a wallet with no signer cannot be disconnected. Maybe derive connected from status === 'connected' alone, expose a separate canSign.

app/providers/wallet-provider.tsx
```diff
@@ -1,46 +1,46 @@
...
+ * Wallet clients are cached per chain because each one runs its own Wallet Standard discovery and
+ * silent reconnect. Rebuilding on every mount — or on every hop between two clusters — would restart
+ * both and flash a disconnected UI. The cache is bounded by the number of chains
+ * {@link clusterToWalletChain} can return, so the clients are held for the life of the page rather
+ * than disposed; disposing one would defeat the reconnect it exists to preserve.
+ */
+const walletClients = new Map<string, WalletClient>();
+
+function getWalletClient(chain: WalletChain, autoConnect: boolean): WalletClient {
+    const key = `${chain}:${autoConnect}`;
```
askov: **suggestion**: cache keyed ${chain}:${autoConnect} while storageKey is only kit-wallet:${chain}, so two live clients can share one persisted selection. Both comments hide the second key dimension — :19-21 claims the bound is chains alone. Also a module-level mutable cache, against your recorded preference. The solution is to drop autoConnect from the cache key, fix both comments.

app/providers/wallet/disposable-stack-polyfill.ts
```diff
@@ -0,0 +1,81 @@
+/**
+ * Installs `Symbol.dispose` and `DisposableStack` where the runtime does not provide them.
+ *
+ * Kit registers plugin cleanup through both, and browsers only gained them in 2025 (Chrome 134,
+ * Firefox 137, Safari 26). Importing this module before any Kit client is created keeps Explorer
```
askov: **todo**: Safari 26.3.1 here, still no Symbol.dispose, so the comment must be misleading.

app/providers/wallet/__tests__/sign-web3js-transaction.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function sign(messageBytes: ReadonlyUint8Array): SignatureBytes {
+    return nacl.sign.detached(new Uint8Array(messageBytes), keypair.secretKey) as SignatureBytes;
+}
+
+const modifyingSigner = {
+    address: signerAddress,
+    modifyAndSignTransactions: (transactions: readonly KitTransaction[]) =>
+        Promise.resolve(
+            transactions.map(transaction => ({
```
askov, line 89: **suggestion**: The modifyingSigner fake spreads ...transaction and only adds a signature, so messageBytes is identical and the modifying path - the module's whole justification, and scenario spec.md:53-57 - is never exercised. The round trip is currently correct, but rests on a web3.js internal (Transaction.populate storing _message) that no test pins. Make the fake prepend an instruction, then assert the modification survives and verifySignatures() is true.

app/providers/wallet/use-wallet.ts
amilz: added `canSign`

app/providers/wallet/use-logged-wallet-error.ts
amilz: done

app/providers/wallet/disposable-stack-polyfill.ts
amilz: fixed comment

app/providers/wallet/__tests__/sign-web3js-transaction.spec.ts
amilz: Fake now prepends an instruction and signs the rebuilt message

app/providers/wallet-provider.tsx
amilz: dropped the prop, hardcoded `autoConnect: true`, and the map is now `Map<WalletChain, WalletClient>`-- lmk what you think.

app/providers/wallet/use-wallet.ts
amilz: added new test file w/ guard

askov (approved):

Woody4618 (approved):

PR #1227, opened 2026-08-19
fix(cluster): keep the consent question on screen while the dialog closes

askov:
## Description

`CustomUrlConsentDialog` branched with a two-way ternary over a three-state prop, so the `undefined` that closes the dialog rendered the developer-bypass copy:

```tsx
{request?.kind === 'endpoint' ? <EndpointConsent … /> : <BypassConsent … />}
```

Radix keeps `DialogContent` mounted until its exit animation ends (`duration-200`), so answering the endpoint question swapped it for "Stop asking about custom RPC servers?" — red **Turn on anyway** button included — for ~200ms on the way out. Every close path hits it: Cancel, Connect, Escape and the X. On Cancel it is camouflaged: `router.replace` repaints the navbar, cluster and stats at that same instant, so the eye follows the page, not the box that is already dismissing.

No consent can be granted in that window. The visible button is wired to the pending-consent mount, whose `onConfirm` is guarded on the request, and `customUrlEnabledAtom` is only ever written from the developer-settings switch. The bug is the message, not the gate: the last thing shown after refusing one server was an offer to stop asking about all of them.

The fix retains the last request, so a closing dialog keeps the question it asked — the endpoint's host stays on screen while the box fades, rather than collapsing onto nothing — and branches per `kind`, so a future third kind cannot fall into the bypass copy either.

From #1209, which added the dialog.

**Why the existing specs were green.** jsdom runs no animations, so `getComputedStyle().animationName` is `none` and Radix `Presence` unmounts on the tick that closes the dialog — the closing render never happens, and the specs assert the dialog is simply gone. The new spec reports the animation name the stylesheet would for the `data-state` Radix has just written, which puts the dialog into `unmountSuspended` exactly as a browser does, then asserts the closing dialog still shows the question it asked and never the other one. Reverting the component fails it.

## Type of change

-   [x] Bug fix

## Testing

- [Root page, endpoint prompt — Cancel, then reload and Connect: the question must keep its own copy while the box fades](https://explorer-git-fork-hoodieshq-fix-consen-15c88c-solana-foundation.vercel.app/?cluster=custom&customUrl=https://api.devnet.solana.com)
- [Address page, the same prompt from a deep link — Escape and the X close it the same way](https://explorer-git-fork-hoodieshq-fix-consen-15c88c-solana-foundation.vercel.app/address/SysvarC1ock11111111111111111111111111111111?cluster=custom&customUrl=https://api.devnet.solana.com)
- [Cluster switcher → Developer Settings → "Trust any custom url param" → Cancel: the bypass question keeps its copy on the way out](https://explorer-git-fork-hoodieshq-fix-consen-15c88c-solana-foundation.vercel.app/)

## Related Issues

Closes [HOO-1155](https://linear.app/solana-fndn/issue/HOO-1155/closing-the-endpoint-consent-dialog-flashes-the-developer-bypass)

## Checklist

-   [x] All checks pass locally (`pnpm test`, `pnpm lint`, `pnpm exec tsc --noEmit`)

## Additional Notes

`bench/BUILD.md` is regenerated in its own commit. The `/address/*` size deltas there are pre-existing drift on `master`, not from this change: two consecutive `build:info` runs on this tree produce the committed table byte for byte, and the fix touches one client component in the cluster switcher.

### DIFF
--- a/app/features/cluster-switcher/ui/CustomUrlConsentDialog.tsx
+++ b/app/features/cluster-switcher/ui/CustomUrlConsentDialog.tsx
@@ -11,6 +11,7 @@ import {
 } from '@components/shared/ui/dialog';
 import type { RpcEndpoint } from '@entities/cluster';
 import { Alert } from '@shared/ui/Alert';
+import { useRef } from 'react';
 
 // Two things share this dialog: a link supplying an endpoint nobody has agreed to, and the developer
 // toggle that stops asking altogether.
@@ -28,6 +29,10 @@ type Props = {
 };
 
 export function CustomUrlConsentDialog({ request, onConfirm, onCancel }: Props) {
+    // What the dialog shows, which outlives what it is asking: the same `undefined` that closes it would
+    // otherwise blank the body, and the content stays on screen for the whole exit animation.
+    const shown = useRetainedRequest(request);
+
     // Anything that closes the dialog — Escape, the X, Cancel — is a decline: the safe outcome has to be
     // the default one. A click outside closes nothing, though: this asks a security question, and a stray
     // click on the backdrop is not an answer to it.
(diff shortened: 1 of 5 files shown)


rogaldh (dismissed):
LGTM

Woody4618 (approved):

PR #1228, opened 2026-08-19
feat(MCP): support idl for frozen programs

rogaldh:
`/mcp` reported `idl: not_found` for programs that do publish an IDL. A frozen program has no upgrade
authority, so the canonical PMP metadata PDA cannot exist for it by construction — every IDL the
Foundation publishes lives under a fallback authority instead, and `@explorer/idl-decode` only ever
looked at the canonical one. Every affected program was invisible — Token, Stake, Config, Address
Lookup Table and Feature Gate when this started, joined by System, Associated Token Account, Compute
Budget and Memo v2 as the Foundation kept publishing — which also cost `decoded` / `source: "idl"` on
their accounts and instructions.

- `@explorer/idl-decode/fetch` now resolves through `@solana/idl` behind a `resolve/` submodule: PMP
  canonical PDA → fndn fallback authority → Anchor PDA, one leg per file, upstream's `IdlSource`
  passed through rather than re-derived.
- _Own leg ordering, not upstream's `fetchIdlWrapped`_ — the wrapper cannot be told to skip the
  Anchor leg (native programs must not read the derived PDA at all; some RPCs answer a transient
  error there instead of `null`), and a valid Anchor IDL there masks a corrupt PMP one.
- `PROGRAMS_WITHOUT_ANCHOR_IDL` in the inspector skips that leg for programs that definitionally
  cannot have an Anchor IDL — loaders included, since a deploy instruction's `program_id` IS the loader.
- _MCP wire vocabulary_ — `source` (`pmp` / `anchor`) plus `source_type` refined to `pmp_canonical` /
  `pmp_fallback` / `anchor`, and PMP results carry the `authority` whose PDA served the IDL.
  `anchor_on_chain` is renamed to `anchor`.
- New capability: `buffer` reads the IDL one named account holds — staged (`anchor idl write-buffer`,
  PMP `write` before `setData`) or committed, either family, dispatched on the account's owner.
- _An IDL decode now carries what the bundled parsers used to own_ — `decoded.accounts` maps every
  IDL-declared role to its address (the codama engine does that pairing itself, so optional and
  remaining accounts stay the engine's problem, and the instruction's positional `accounts` list stays
  complete), and token decodes carrying their own `decimals` add `decoded.ui_amount`. This closes the
  information gap on the IDL path instead of reordering the cascade around it.
- The app is untouched: it keeps `@solana/idl` directly, and the inspector stays the only consumer of
  this seam.

## Type of change

-   [x] Bug fix
-   [x] New feature
-   [ ] Protocol integration
-   [x] Documentation update

## Screenshots

No UI surface — package internals and MCP wire fields only.

## Testing

-   Unit: one case per error-mapping row, the discovery mapping, and a real Codama root driven through
    the cascade for the named accounts and ui amount. Both packages stay at 100% coverage.
-   `pnpm test` (375 files, 4094 tests), `pnpm lint`, `pnpm typecheck` clean locally.
-   Live mainnet via a local MCP: every fndn-published program answers `found` + `pmp_fallback` — the
    complete set, enumerated by `getProgramAccounts` on the PMP program with a `memcmp` on the fndn
    authority at offset 33, re-run as it grew from five to nine.
-   All 31 baseline entity fixtures still match, and both transaction fixtures keep their exact source
    attribution, so the added fields moved no instruction between cascade rungs.
-   Not covered by a fixture: an `unknown`-kind account owned by an fndn program provably does not exist
    on mainnet today — Stake and Feature declare no accounts, Config's carries no discriminator, and
    Token/ALT accounts classify before the unknown-kind gate. Unit specs are that leg's only coverage.

## Related Issues

n/a

## Checklist

## Additional Notes

-   _Known behaviour change, shipped deliberately_ — `runDecodeCascade` is IDL-first, so making these
    IDLs resolvable flips instruction decoding for every fndn-published program, System included.
    Measured on mainnet tx
    `5sYUDzH7L1CsgVFWfDg9MKNZX455ecz8cAnHvw2WWj6SGDRmpxCzUCg8NAE3U9WvDqmzAdJBr7cNzMpviuwomYjc`:
    `transferChecked` moves from `source: "bundled"` to `source: "idl"`. The information that move used
    to cost is now carried on the IDL path (named accounts, `ui_amount`), so what remains is wording —
    the `type` reads `Transfer Checked` rather than `transferChecked`, and `program` the IDL's own name
    rather than `spl-token`. Feature Gate improves outright (`raw` → decoded). Excluding programs from
    tx-path IDL resolution was the alternative and was dropped: the fndn set grew from five to nine
    inside this branch's lifetime, so any static exclusion list is stale by construction.
-   `decode_sources` analytics shift with that flip — an instruction previously counted `bundled` now
    counts `idl`.
-   Inspector-layer decode coverage was the gap that hid the shape change; the cascade spec now drives
    a real Codama root (a `transferChecked` node the spec owns) end to end, so account naming and the
    ui amount are pinned where they ship. A mainnet-snapshot IDL still runs only in
    `@explorer/idl-decode`, which is where those fixtures live.
-   `@solana/idl@0.2.0` declares `@solana/kit ^6.4.0` and 0.2.0 is the latest published, so on master's
    kit 7 it installs as an unmet peer. Master already carries the same for `@solana-program/`
    `program-metadata`, `system` and `compute-budget`; this adds one more. Types and tests are clean
    against kit 7.
-   `createLatestIdlFetcher` / `fetchLatestIdlClient` remain as aliases for the pre-rename names.
-   `no-restricted-imports` added to the inspector's oxlint config, making the "never import app code"
    constraint enforced rather than conventional — it is why `PROGRAMS_WITHOUT_ANCHOR_IDL` duplicates
    the app's list instead of importing it.


packages/entity-inspector/src/enrichments/idl-clients.ts
```diff
@@ -31,12 +37,13 @@ function createIdlRpc(cluster: SupportedCluster, rpcEndpoints: Record<SupportedC
...
 
-type FetchOutcome = { fetched: FetchedIdlClient } | { error: IdlError } | { rejected: unknown };
+type FetchOutcome = { fetched: PublishedIdlClient } | { error: IdlError } | { rejected: unknown };
 
-async function fetchLatest(programAddress: string, rpc: IdlFetcherRpc): Promise<FetchOutcome> {
+async function fetchOnChain(programAddress: string, rpc: IdlFetcherRpc): Promise<FetchOutcome> {
     try {
-        const [error, fetched] = await fetchLatestIdlClient(programAddress, {
+        const [error, fetched] = await fetchOnChainIdlClient(programAddress, {
             abortSignal: AbortSignal.timeout(RPC_REQUEST_TIMEOUT_MS),
```
askov: **issue**: The timeout only covers the RPC reads. A PMP entry with dataSource: Url ends in a plain fetch(url) with no signal and no timeout. resolveIdlClients awaits these in Promise.all, so one slow URL blocks the whole inspect_entity response. The old fetch/pmp.ts refused non-Format.Json entries before downloading; @solana/idl has no such gate, and there are now two PMP lookups instead of one, so more entries reach the fetch than before. Maybe just use raceWithTimeout

packages/entity-inspector/src/enrichments/idl-clients.ts
```diff
@@ -46,16 +53,36 @@ async function fetchLatest(programAddress: string, rpc: IdlFetcherRpc): Promise<
...
 }
 
-function toFoundDiscovery(fetched: FetchedIdlClient): IdlDiscoveryResult {
+// A switch, not a ternary: the declared return type makes an unhandled IdlSource a compile error rather
+// than a silent misattribution.
+function toSourceType({ authority, source }: PublishedIdlClient): IdlSourceType {
+    switch (source) {
+        case IdlSource.Pmp:
+            // an authority key means a fallback PDA served it; the canonical PDA derives from no authority
+            return authority ? 'pmp_fallback' : 'pmp_canonical';
```
askov: **issue**: A buffer read has no authority key at all, so this reports it as pmp_canonical — the one thing it is not. Check 'authority' in fetched instead of truthiness. Not reachable today since we never pass buffer, but buffer is public API and the package test already pins the state (fetch-idl-client.spec.ts:774).

packages/entity-inspector/src/transactions/token-ui-amount.ts
```diff
@@ -0,0 +1,46 @@
...
+
+function readCheckedAmount(info: unknown): { amount: bigint; decimals: number } | undefined {
+    if (typeof info !== 'object' || info === null || !('amount' in info) || !('decimals' in info)) {
+        return undefined;
+    }
+    const { amount, decimals } = info;
+    if (typeof decimals !== 'number' || !Number.isInteger(decimals) || decimals < 0) {
+        return undefined;
+    }
+    if (typeof amount === 'bigint') {
```
askov: **issue**: The string branch rejects a negative, the bigint branch does not. { amount: -5n, decimals: 6 } returns "0.0000-5" — padStart puts zeros left of the minus. One line fixes it: if (amount < 0n) return undefined;. The spec covers '-100' but not -100n.

packages/entity-inspector/src/mcp/server.ts
```diff
@@ -24,11 +24,15 @@ const buildInspectEntityDescription = (clusterNames: EnabledClusterNames) =>
...
+        '  Note "verification.evidence.is_frozen" is registry metadata recorded when the build was attested, so it can disagree with the current on-chain state; "upgradeable" is read live.',
         '- "bpf-loader" / "bpf-loader-2" / "loader-v4": Legacy-loader programs — currently unsupported; return a CURRENTLY_UNSUPPORTED error.',
+        '- "program-metadata:metadata" / "program-metadata:buffer" / "program-metadata:empty": Accounts of the program-metadata (PMP) program, subtyped by the discriminator they carry. Both metadata and buffer report "program" (the program the entry describes, NOT this account), "seed" (e.g. "idl"), "authority" (null when none) and "canonical"; metadata adds "mutable", "format", "encoding", "compression", "data_source" and "data_length". This is where a program\'s PMP-published IDL physically lives, so "canonical": false matches the "pmp_fallback" source_type above.',
```
askov: **suggestion**:  The buffer branch also emits data_length (program-metadata.ts:57), so it is not metadata-only. They also measure different things — metadata declares a length, buffer counts the bytes staged so far. Worth saying which is which. Also program can be null on a buffer; the "(null when none)" note is only on authority.

packages/entity-inspector/src/accounts/account-kinds/program-metadata.ts
```diff
@@ -0,0 +1,78 @@
...
+            const header = getBufferDecoder().decode(bytes);
+            return {
+                authority: unwrapOption(header.authority),
+                canonical: header.canonical,
+                data_length: header.data.length,
+                program: unwrapOption(header.program),
+                seed: header.seed,
+            };
+        }
+    } catch {
```
askov: **issue**: Returning {} gives the same output as an empty account, so a broken metadata account is indistinguishable from an intentionally empty one. The sibling builders use unknownMarker(...) for exactly this. The account only got here because the owner and the discriminator both matched, so a decode failure is a real anomaly worth surfacing.

packages/entity-inspector/src/enrichments/idl-clients.ts
rogaldh: Fixed

askov (dismissed):

askov (dismissed):

Woody4618 (approved):

PR #1229, opened 2026-08-20
feat(transaction-data): resolve instruction names for CU charts and history

askov:
## Description

The CU profiling chart labelled every row by position — `Instruction #1`, `Instruction #2` — and the
simulator and transaction-history rows had no shared way to name an instruction. This adds one naming
pipeline in the `transaction-data` entity and routes all three surfaces through it.

Naming runs in three stages:

1. **Stage 1 — sync, no I/O.** `resolveInstructionNames` reads an RPC-parsed instruction;
   `resolveNamesFromData` reads raw bytes from a compiled message. Anything neither can name carries a
   `nameLookup` onward.
2. **Stage 2 — the only I/O.** `useProgramIdlNames` fetches the IDLs those rows need, once per
   transaction, in a single SWR entry.
3. **Stage 3 — sync, no I/O.** `NAME_SOURCES` runs in order, first hit wins: Compute Budget, Memo, ZK
   ElGamal, Lighthouse, Mango, Serum, the generated `@solana-program/*` clients, and finally the IDL —
   which is the default for any program not in that list.

Notable pieces:

- `useResolvedInstructionNames` / `useResolvedSummaryNames` wrap stages 2 and 3, so each surface wires
  naming in one line.
- `applyNameSources` is the single owner of the `nameLookup` rule: present means "still unnamed", dropped
  the moment a source names the row.
- `rpcTypeOverrides` keeps generated-client wording matching the RPC's `parsed.type`, so a simulated
  instruction reads the same as a fetched one (`Transfer`, not `Transfer Sol`). Verified against 11
  distinct `(program, parsed.type)` pairs from real mainnet transactions — 0 mismatches.
- `NON_ANCHOR_PROGRAMS` keeps builtins out of the IDL fetch.

`app/entities/transaction-data/README.md` documents the flow, the seven invariants it holds to, how to
add a name source, and why the built-in sources are not simply replaced by IDLs.

## Type of change

- [x] New feature

## Testing

- [CU chart names Compute Budget, an Anchor program and a parsed instruction](https://explorer-git-fork-hoodieshq-feat-instr-6afc2d-solana-foundation.vercel.app/tx/5i8nFf2kFDjXJ13ixoV8QXVfMYtbK7bYDemuQATrzDCNL1V4Q8jitmQHBZ2G91j7bavBMaMbaHoePKC3wM3sSjyZ) — four rows, in order: `Set Compute Unit Limit`, `Set Loaded Account Data Size Limit`, `Route` (Jupiter, appears once the IDL lands), `Transfer`.
- [Row order survives a trailing Compute Budget instruction](https://explorer-git-fork-hoodieshq-feat-instr-6afc2d-solana-foundation.vercel.app/tx/2UP5PpKrvGn85nhVvv3keGiQY8vp2aYvrHNGhc9BeQTp9FUpNULNE2yTCvoxug3FsuxqGqEbSypCEBh3AhwMUHEW) — a marginfi flashloan: `Lending Account Start Flashloan` / `Lending Account End Flashloan` either side of row 4, which stays `Instruction #4` because that program has no IDL and its data is the ASCII marker `SCANNER1`. Row 6 must be the Compute Budget row; rows pair with log entries by index, so a shift here misattributes every later CU figure.
- [A built-in source resolves with no IDL fetch](https://explorer-git-fork-hoodieshq-feat-instr-6afc2d-solana-foundation.vercel.app/tx/5sW2kLTJmSAVTjsqs7jWsCi5y1W9JjBLdXxjJqBphEUTVEWYBP9k6kUyH8ALgGjktma6XK8qGjCUzEb37iTXX5e9) — row 3 reads `Assert Account Info` (Lighthouse) on first paint, with no `Instruction #3` flicker.
- [Memo on the transaction page](https://explorer-git-fork-hoodieshq-feat-instr-6afc2d-solana-foundation.vercel.app/tx/5AC2ujRqweyFKZ7bSVsGBifYG11vfGaZ52X9VCmMh3ydcDAiYBMpU3RTt7L73vyehRwzWq8LKLmMAgiTfTC4jyEW) — `Transfer` then `Memo`, both from the RPC parse.
- [The same transaction in the simulator](https://explorer-git-fork-hoodieshq-feat-instr-6afc2d-solana-foundation.vercel.app/tx/5AC2ujRqweyFKZ7bSVsGBifYG11vfGaZ52X9VCmMh3ydcDAiYBMpU3RTt7L73vyehRwzWq8LKLmMAgiTfTC4jyEW/inspect) — press **Simulate**. There is no RPC parse here, so both names come from `NAME_SOURCES` and must match the page above exactly: `Transfer`, not `Transfer Sol`, and `Memo`, not `Instruction #2`.
- [Transaction-history rows](https://explorer-git-fork-hoodieshq-feat-instr-6afc2d-solana-foundation.vercel.app/address/capsFrvPsaCQ6y7RUk3TYkB5D1oNCNjRb4uYuB5UCsZ) — instruction names in the history card. Compute Budget is deliberately absent: `getInstructionSummaries` filters it out of summary rows while the CU chart keeps it.

## Related Issues

Closes [HOO-805](https://linear.app/solana-fndn/issue/HOO-805/parse-transaction-names-in-cu-profiler)

## Checklist

- [x] All checks pass locally (`pnpm test`, `pnpm lint`, `pnpm typecheck`)
- [x] I have updated documentation as needed

## Additional Notes

Naming never fetches outside stage 2, and never throws — `identifyInstruction` converts every
generated-client throw into `undefined`, because these run during render with no error boundary.

Row count and order are preserved through every stage. The CU chart pairs row `i` with
`instructionLogs[i]`, so a dropped row would shift every later CU figure onto the wrong instruction. That
is why Compute Budget rows are kept in the chart even though `getInstructionSummaries` filters them out
of history rows, and why a simulated message with an out-of-range `programIdIndex` yields no rows at all
rather than a partial set.


app/entities/transaction-data/lib/name-sources.ts
```diff
@@ -0,0 +1,113 @@
...
+ * The single owner of the `nameLookup` rule; both row shapes go through here, so neither can drift.
+ * @param names - The names resolved from the transaction alone
+ * @param idlNames - IDL-derived names per program id, for the programs in this transaction
+ */
+export function applyNameSources(names: InstructionNames, idlNames: Map<string, ProgramIdlNames>): InstructionNames {
+    if (!names.nameLookup) return { name: names.name, programName: names.programName };
+
+    const resolved = resolveNamesFromLookup(names.nameLookup, idlNames);
+    const name = resolved.name ?? names.name;
+    const programName = resolved.programName ?? names.programName;
```
rogaldh: This part is a bit tricky.

Theoretically the could not be a name, but the programName could be resolved. Also, taking line 101 into account, we will use the names provided in the function and compare strings to return the summary (if the name and program name match).

I'd suggest rewriting this part for clarity

rogaldh (dismissed):
Other than that, LGTM

app/entities/transaction-data/lib/name-sources.ts
askov: Fixed

Woody4618 (approved):

app/entities/compute-unit/ui/__stories__/BaseCUProfilingCard.stories.tsx
```diff
@@ -0,0 +1,193 @@
+import type { Meta, StoryObj } from '@storybook-config/types';
+
+import { BaseCUProfilingCard } from '../BaseCUProfilingCard';
+
+const meta: Meta<typeof BaseCUProfilingCard> = {
+    component: BaseCUProfilingCard,
+    tags: ['autodocs', 'test'],
+    // TODO(dashkit): rename to match BaseCUProfilingCard once the Dashkit migration lands. Kept as-is
```
Woody4618: Dashkit migration we did already right?

PR #1233, opened 2026-08-21
Refactor/hoo 1186 read v1 blocks with kit

amilz:
Closes HOO-1186

Block pages sent `maxSupportedTransactionVersion: 0`. `getBlock` has no partial mode, so one v1 transaction failed the whole block and all four subpages showed the error state. Reproduced against a v1 transaction landed on the SIMD-296 surfnet.

Pinning to `1` fails at runtime on our 1.98.4: web3.js types the option as `number`, but its schema admits only `0`/`legacy` and has nowhere to put `transactionConfig`. kit is the migration direction, so blocks follow the transaction and inspector paths (#1192, #1204) and reuse that v1 message bridge. The four cards change types only.

Second, silent fix: `estimateRequestedComputeUnits` scanned for ComputeBudget instructions, which v1 doesn't carry, so it reported 200k for a transaction that requested 10,000 — wrong Reserved CUs, sort, and block total, no error.

Steps to test: 
- Open [Preview](https://explorer-git-refactor-hoo-1186-read-v1-ef230d-solana-foundation.vercel.app/)
- Run Local test validator (v4.2.1+)
- Change network to local
- Run a V1 transaction ([example](https://gist.github.com/amilz/832fb14baf7dc2bdc5ec2f42d7c3f280))
- Open the block the transaction was set in on block explorer
- It should render (run the same test on production explorer site, and it should crash the block page)

### DIFF
--- a/app/entities/block-data/api/fetch-block.ts
+++ b/app/entities/block-data/api/fetch-block.ts
@@ -44,6 +44,8 @@ export async function fetchBlock(url: string, slot: number): Promise<BlockWithV1
         parentSlot: Number(response.parentSlot),
         previousBlockhash: response.previousBlockhash,
         rewards: response.rewards?.map(reward => ({
+            // Only staking and voting rewards carry a commission.
+            commission: 'commission' in reward ? reward.commission : undefined,
             lamports: Number(reward.lamports),
             postBalance: reward.postBalance === null ? null : Number(reward.postBalance),
             pubkey: reward.pubkey,
--- a/app/entities/block-data/api/fetch-block.ts
+++ b/app/entities/block-data/api/fetch-block.ts
@@ -6,12 +6,18 @@ import {
     MAX_SUPPORTED_TRANSACTION_VERSION,
     type Slot,
     type Transaction,
-    type TransactionError,
 } from '@solana/kit';
-import { PublicKey, type TokenBalance, VersionedMessage } from '@solana/web3.js';
+import { PublicKey, VersionedMessage } from '@solana/web3.js';
+import { create } from 'superstruct';
 
+import { Logger } from '@/app/shared/lib/logger';
 import { bridgeV1MessageBytes, isV1MessageBytes } from '@/app/shared/lib/v1-message-bridge';
 
+import {
+    BlockResponseSchema,
+    type BlockTransactionResponse,
+    BlockTransactionResponseSchema,
+} from '../model/block-response-schema';
 import type { BlockTransaction, BlockWithV1 } from '../model/types';
 
 /**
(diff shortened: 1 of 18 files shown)


app/entities/block-data/api/fetch-block.ts
```diff
@@ -0,0 +1,160 @@
...
+        blockTime: response.blockTime === null ? null : Number(response.blockTime),
+        blockhash: response.blockhash,
+        parentSlot: Number(response.parentSlot),
+        previousBlockhash: response.previousBlockhash,
+        rewards: response.rewards?.map(reward => ({
+            lamports: Number(reward.lamports),
+            postBalance: reward.postBalance === null ? null : Number(reward.postBalance),
+            pubkey: reward.pubkey,
+            rewardType: reward.rewardType,
+        })),
```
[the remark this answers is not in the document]
amilz: 696da9c05c3cd87f1d238df50ac3bdac9ac90ea4

app/entities/block-data/api/fetch-block.ts
```diff
@@ -0,0 +1,162 @@
...
+        previousBlockhash: response.previousBlockhash,
+        rewards: response.rewards?.map(reward => ({
+            // Only staking and voting rewards carry a commission.
+            commission: 'commission' in reward ? reward.commission : undefined,
+            lamports: Number(reward.lamports),
+            postBalance: reward.postBalance === null ? null : Number(reward.postBalance),
+            pubkey: reward.pubkey,
+            rewardType: reward.rewardType,
+        })),
+        transactions: response.transactions.map(adaptTransaction),
```
askov: **suggestion**: It has no per-item guard, and adaptTransaction has five throw sites, any rejects fetchBlock and contradicts with the PR spec. Shouldn't happen, but try/catch just in case would be nice

app/entities/block-data/api/fetch-block.ts
```diff
@@ -0,0 +1,162 @@
...
+            })) ?? undefined,
+        loadedAddresses: meta.loadedAddresses
+            ? {
+                  readonly: meta.loadedAddresses.readonly.map(address => new PublicKey(address)),
+                  writable: meta.loadedAddresses.writable.map(address => new PublicKey(address)),
+              }
+            : undefined,
+        logMessages: meta.logMessages === null ? null : [...meta.logMessages],
+        postBalances: meta.postBalances.map(Number),
+        postTokenBalances: meta.postTokenBalances === undefined ? undefined : [...meta.postTokenBalances],
```
askov: **question**:  postTokenBalances could be null if I'm not mistaken?

app/entities/block-data/api/fetch-block.ts
```diff
@@ -0,0 +1,162 @@
...
+        .getBlock(BigInt(slot) as Slot, {
+            commitment: 'confirmed',
+            encoding: 'base64',
+            maxSupportedTransactionVersion: MAX_SUPPORTED_TRANSACTION_VERSION,
+            rewards: true,
+            transactionDetails: 'full',
+        })
+        .send();
+
+    if (response === null) {
```
askov, line 42: **suggestion**: We usually try to validate all responses with superstruct. Was really helpful a couple of times

app/entities/block-data/api/fetch-block.ts
amilz: addressed in new schema

app/entities/block-data/api/fetch-block.ts
amilz: added `app/entities/block-data/model/block-response-schema.ts` lmk if this is what you are thinking

app/entities/block-data/api/fetch-block.ts
amilz: wrapped adapt in try/catch

askov (approved):

PR #1238, opened 2026-08-21
Add transaction size to transaction page

Woody4618:
## Description
add transaction size to transaction page: 

image

For TV1 
image

### DIFF
--- a/app/entities/transaction-data/api/fetch-raw-transaction.ts
+++ b/app/entities/transaction-data/api/fetch-raw-transaction.ts
@@ -2,13 +2,13 @@ import {
     createSolanaRpc,
     decodeTransactionFromRpcResponse,
     getBase58Decoder,
+    getTransactionSize,
     MAX_SUPPORTED_TRANSACTION_VERSION,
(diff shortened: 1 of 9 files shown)


askov (dismissed):

app/entities/transaction-data/api/fetch-raw-transaction.ts
```diff
@@ -62,6 +64,7 @@ export async function fetchRawTransaction(
                   preBalances: meta.preBalances.map(Number),
               }
             : undefined,
+        serializedSize: base64ByteLength(base64Transaction),
```
rogaldh: Could we use getTransactionSize here?

rogaldh (dismissed):
Except for this, LGTM

app/entities/transaction-data/api/fetch-raw-transaction.ts
Woody4618: Good call! Thx 
image

rogaldh (approved):
LGTM

PR #1239, opened 2026-08-21
refactor(instruction-card): share card chrome via surface context

askov:
## Description

Instruction cards took their surface chrome as props. `InstructionCardComponent` swapped the card frame in 7 files, `AssociatedTokenDetailsCard` additionally took `AddressComponent` and `showProgramField`, `innerCards` was threaded through four layers to reach `BaseInstructionCard`, and the inspector invented `INSPECTOR_RESULT` / `INSPECTOR_SIGNATURE` purely to satisfy card signatures. Row markup was copied per field: the 13 System cards were 815 lines, almost all of it the same `<BaseTable.Row>` shape repeated.

This adds an `instruction-card` entity that supplies chrome ambiently and lets cards declare their rows as data:

- **`InstructionSurface` via Context** — card frame, address renderer, `showProgramField`, and the transaction's `SignatureResult`. `useInstructionSurface` throws outside a provider, matching `useInstructionParser`.
- **`InstructionNode`** — one prop replacing the `ix` + `index` + `childIndex` + `raw` + `innerCards` spread. `programId` is its own field and the only one a card reads; `node.ix` exists for the frame and has exactly one reader in the tree.
- **Field descriptors** — `address`, `sol`, `bytes`, `seed`, `text`, `custom`. `InstructionFields` is the only place that knows row markup, apart from `ProgramField`, which owns the leading `Program` row on both surfaces.
- **`InstructionCardView`** — the escape hatch for cards that need real markup, and the carrier for card-derived program events.

All 13 System cards migrate to the `defineInstructionCard` factory: **815 → 155 lines**. No file under `app/components/instruction/system/` imports `Address`, `BaseTable`, or `InstructionCard` any more.

Scope is deliberately one program. 58 frame-drawing cards still take chrome as props and migrate program by program — a partially migrated program would render two different frames side by side in the inspector. `openspec/changes/unify-instruction-card-rendering/proposal.md` carries the rationale, the audit behind the design, the rejected alternatives, and the follow-ups.

### Behaviour

**Transaction page: no change.** Verified by rendering against a `HEAD` worktree and diffing the resulting DOM — cell text, button set, wrapper elements and `<code>` counts are identical for the Transfer and TransferWithSeed fixtures. Every piece is the same component receiving the same props, which is the argument that carries the other 11 instructions.

**Inspector: four chrome deltas on System cards**, which previously used the tx-page frame on both surfaces — a collapse chevron, an address wrapper div, and in Raw mode the hex/accounts block plus `AddressWithContext` on the Program row. Row labels, order and count are unchanged, and all four align System with what Token, Token-2022 and ATA already do there.

One caveat worth flagging to reviewers, because it bounds the claim above: the inspector consults its dynamic IDL tier *before* its program switch — `inspector/InstructionsSection.tsx` returns `IdlInstructionCard` as soon as `useIdlInstructionDecode` resolves, and reaches its `SYSTEM_PROGRAM_LABEL` case only when it does not — so once an IDL resolves for the System program the IDL card replaces `SystemDetailsCard` outright, measured at roughly 1s against a dev server. The migrated card is therefore what users read on the **tx page**, whose `InstructionsSection` runs its program switch first and falls through to the IDL tier, while in the inspector it is currently reachable only when no IDL resolves. **This ordering is pre-existing, identical on `master`, and untouched by this PR**, but it means every program migrated after System is shadowed the same way wherever an IDL resolves for it. Tracked as [HOO-1194](https://linear.app/solana-fndn/issue/HOO-1194/inspector-prefers-a-fetched-idl-over-curated-instruction-cards).

## Type of change

-   [x] Other (please describe): refactor — no user-facing feature change; adds a rendering layer and migrates one program onto it

## Testing

Manual testing on the preview deployment. For the inspector links, **click Raw** — that is where the chrome deltas show.

-   [Tx page — System Transfer](https://explorer-git-fork-hoodieshq-refactor-u-49ca17-solana-foundation.vercel.app/tx/VLf51Wds42nr49qTw19aKjKN8pvt6PDWLJFf8aiveCxsZW6FZX8JxPDz79sD9xTg3CTWvEfKTZpytPwp6Y5jWzR) — a single-instruction mainnet transfer. **Start here:** this is the surface where the migrated card is what users actually read. Check row labels, ordering, the single `Program` row, SOL formatting, and the Raw toggle.
-   [Inspector — System Transfer](https://explorer-git-fork-hoodieshq-refactor-u-49ca17-solana-foundation.vercel.app/tx/inspector?message=gAEAAQMMMLe3YoGRo3K1%2FT5QfmVUXG8b7svyH9T32%2FUutO6CZQljvMFxLJLdvwFng1FAga7VlwYT%2FgoHXUOmvvzR7TwBAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACcyfP%2FJFK6VWlYc2EGEyE78uvE0UgAvaW5%2BvYLw8BggAECAgABDAIAAABAS0wAAAAAAAA%3D) — the fixture the tx-page DOM diff was run against.
-   [Inspector — System Create Account](https://explorer-git-fork-hoodieshq-refactor-u-49ca17-solana-foundation.vercel.app/tx/inspector?message=gAIAAQMMMLe3YoGRo3K1%2FT5QfmVUXG8b7svyH9T32%2FUutO6CZQy3%2FvTQDEJOWoAM9MX7grE4H3nMn6BBJUeHUq3C8%2BApAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACcyfP%2FJFK6VWlYc2EGEyE78uvE0UgAvaW5%2BvYLw8BggAECAgABNAAAAABAQg8AAAAAAGQAAAAAAAAAjJclj04kifG7PRApFI4NgwtaE5na%2FxCEBI572Nvp%2BFkA) — the richest System card, 5 rows plus `Program`.
-   [Inspector — System Create Account w/ Seed](https://explorer-git-fork-hoodieshq-refactor-u-49ca17-solana-foundation.vercel.app/tx/inspector?message=gAIBAQQMMLe3YoGRo3K1%2FT5QfmVUXG8b7svyH9T32%2FUutO6CZZ1CSy8KSCwmvH2EXJDg%2FzGyj%2FRxpxPWMXQNaRtmJDz5DLf%2B9NAMQk5agAz0xfuCsTgfecyfoEElR4dSrcLz4CkAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAJzJ8%2F8kUrpVaVhzYQYTITvy68TRSAC9pbn69gvDwGCAAQMDAAIBZQMAAACdQksvCkgsJrx9hFyQ4P8xso%2F0cacT1jF0DWkbZiQ8%2BQkAAAAAAAAAdGVzdC1zZWVkgIQeAAAAAADIAAAAAAAAAIyXJY9OJInxuz0QKRSODYMLWhOZ2v8QhASOe9jb6fhZAA%3D%3D) — a useful contrast: this one exercises the `seed` field descriptor.

Expect the three inspector links to settle on the generic IDL card (`System: CreateAccount`, with `bignum`/`pubkey` argument rows) about a second after load, per the IDL-ordering caveat in the Description — the migrated card is visible only briefly there. The tx-page link is unaffected and is the reliable one for reviewing this change.

## Related Issues

Closes [HOO-972](https://linear.app/solana-fndn/issue/HOO-972/share-instruction-card-chrome-via-surface-context-system-program)

## Checklist

## Additional Notes

The spec delta ships at `openspec/changes/unify-instruction-card-rendering/specs/instruction-card/spec.md`. It pins the four rules a future card migration must not break — chrome comes from the surface, fields are data, a card takes one node, and a program migrates whole — and states explicitly that those requirements bind migrated cards rather than describing the current tree.

Two details reviewers may want to look at directly:

- `InstructionFields` renders through `switch (field.kind)` with a `never` default, so adding a descriptor kind without a renderer fails the build rather than silently drawing a blank cell.
- `custom` takes a `ReactElement` rather than a `ReactNode`, which is what makes a labelled row with nothing in it unrepresentable.


openspec/changes/unify-instruction-card-rendering/proposal.md
```diff
@@ -0,0 +1,57 @@
...
+- Row markup is copied per field. The 13 System cards were 815 lines, almost all of it `<BaseTable.Row><Cell>label</Cell><Cell className="text-right">…</Cell></BaseTable.Row>` repeated.
+- `unify-instruction-parsing` named this work as an explicit out-of-scope follow-up: "collapsing the two per-surface card switches into one render pipeline."
+- That change is also what makes the leak observable for System. The inspector reaches System cards only because it landed the System slice and the shared dispatcher — `inspector/InstructionsSection.tsx:130` calls `dispatcher.fromTransactionInstruction(ix)` and `:213` routes `program === 'system'` to `SystemDetailsCard`. Before it, the inspector's own `into-parsed-data.ts` fed a separate path. System is therefore a program where one card component must already satisfy two surfaces, while still hardcoding the tx-page frame for both.
```
rogaldh: The reference to line `:130` seems outdated. Let's maybe use permalinks to the revisions so we don't have stale links.

openspec/changes/unify-instruction-card-rendering/proposal.md
```diff
@@ -0,0 +1,57 @@
...
+  - `ui/InstructionFields.tsx` — for migrated cards, the only place that knows row markup, alignment, and which address renderer the surface uses. The leading `Program` row is the exception: `ProgramField` owns that markup, because the inspector puts its own validator on that row.
+  - `ui/InstructionCardView.tsx` — resolves the frame from the surface. The escape hatch for cards that need their own markup, and the carrier for card-derived program events via an `events` prop.
+  - `model/define-instruction-card.tsx` — shorthand factory for the common label/value shape.
+  - `ui/TxInstructionSurface.tsx` — the transaction page surface.
+- The inspector declares `INSPECTOR_SURFACE` beside its own `InstructionsSection`. `INSPECTOR_RESULT` and `INSPECTOR_SIGNATURE` both survive this change — the surface carries `result` for System only, while unmigrated cards still take it as a prop (14 remaining `result={INSPECTOR_RESULT}` passes and 3 `signature={INSPECTOR_SIGNATURE}` passes). Both constants are deleted when the last inspector-reachable program migrates.
```
rogaldh:
```suggestion
- The inspector declares `INSPECTOR_SURFACE` beside its own `InstructionsSection`. `INSPECTOR_RESULT` and `INSPECTOR_SIGNATURE` both survive this change — the surface carries `result` for System only, while unmigrated cards still take it as a prop (13 of the 14 remaining `result={INSPECTOR_RESULT}` passes and all 3 `signature={INSPECTOR_SIGNATURE}` passes). Both constants are deleted when the last inspector-reachable program migrates and the shared `components/instruction/UnknownDetailsCard` fallback, which keeps System's `result` pass alive, migrates with it.
```

openspec/changes/unify-instruction-card-rendering/proposal.md
rogaldh:
```suggestion
- That change is also what makes the leak observable for System. The inspector reaches System cards only because it landed the System slice and the shared dispatcher — `inspector/InstructionsSection.tsx` calls `dispatcher.fromTransactionInstruction(ix)` and routes its `SYSTEM_PROGRAM_LABEL` case to `SystemDetailsCard`. Before it, the inspector's own `into-parsed-data.ts` fed a separate path. System is therefore a program where one card component must already satisfy two surfaces, while still hardcoding the tx-page frame for both.
```

openspec/changes/unify-instruction-card-rendering/proposal.md
```diff
@@ -0,0 +1,57 @@
...
...ch — `inspector/InstructionsSection.tsx:175` returns `IdlInstructionCard` and `:250` is the `SYSTEM_PROGRAM_LABEL` case — so once `useIdlInstructionDecode` resolves for the System program, the IDL card replaces `SystemDetailsCard` outright. Measured against the dev server with the `CreateAccount` fixture, `SystemDetailsCard` is on screen for roughly 0.75–1.75 s and the IDL card holds the surface from then on. The unit tests never resolve an IDL, which is why they observe the migrated card and pin the deltas. The tx page has the opposite order (`:227` switch before `:403` IDL tier), so `SystemDetailsCard` wins there and is what users actually read. This ordering is pre-existing and identical on `master`; this change does not touch it. It does mean the inspector half of the two-surface argument is theoretical for any program that has a resolvable IDL, and that every program migrated after System is shadowed the same way wherever an IDL resolves for it — which depends on an IDL actually being published for that program, and on the cluster and RPC in use. Tracked as [HOO-1194](https://linear.app/solana-fndn/issue/HOO-1194/inspector-prefers-a-fetched-idl-over-curated-instruction-cards).
```
rogaldh: This one has a lot of references. I'd actualise them

openspec/changes/unify-instruction-card-rendering/specs/instruction-card/spec.md
```diff
@@ -0,0 +1,106 @@
+# instruction-card
+
+## Purpose
+
+One rendering layer for instruction cards, shared by `/tx/[signature]` and `/tx/(inspector)/inspector`. A card declares what an instruction means; the surface it renders on owns the chrome. Adding a field is one descriptor; adding a surface is one Context value.
+
+**Scope.** These requirements bind every card that has migrated to the surface, and every card that migrates from here on. They are not a description of the current tree: System (13 cards) is the only migrated program, and the remaining 58 frame-drawing cards still take their chrome as props. Each requirement below therefore reads as the rule a migration must satisfy, not as an invariant that already holds repo-wide.
```
rogaldh:
```suggestion
**Scope.** These requirements bind every card that has migrated to the surface, and every card that migrates from here on. They are not a description of the current tree: System (13 cards) is the only migrated program, and the remaining 58 frame-drawing cards still resolve their own chrome — 49 hardcode `InstructionCard`, 7 take it as a prop, 2 hardcode it while being inspector-reachable. Each requirement below therefore reads as the rule a migration must satisfy, not as an invariant that already holds repo-wide.
```

openspec/changes/unify-instruction-card-rendering/specs/instruction-card/spec.md
```diff
@@ -0,0 +1,106 @@
...
+
+## ADDED Requirements
+
+### Requirement: Surface-owned chrome
+
+Instruction cards MUST NOT receive surface chrome as props. The card frame, the address renderer, whether the leading `Program` row is emitted, and the transaction's `SignatureResult` SHALL be supplied by an `InstructionSurface` value through React Context.
+
+This supersedes the `InstructionCardComponent`, `AddressComponent`, and `showProgramField` injection props, and the inspector's `INSPECTOR_RESULT` / `INSPECTOR_SIGNATURE` placeholder constants — for migrated cards. All five still exist for the unmigrated ones and are removed program by program, the constants last. `useInstructionSurface` fails loudly rather than falling back to a default, matching `useInstructionParser`.
```
rogaldh:
```suggestion
This supersedes the `InstructionCardComponent`, `AddressComponent`, and `showProgramField` injection props, and the inspector's `INSPECTOR_RESULT` / `INSPECTOR_SIGNATURE` placeholder constants — for migrated cards. All five still exist for the unmigrated ones; the props go program by program, and the constants go last, once the shared `UnknownDetailsCard` fallback stops requiring `result` as well. `useInstructionSurface` fails loudly rather than falling back to a default, matching `useInstructionParser`.
```

openspec/changes/unify-instruction-card-rendering/specs/instruction-card/spec.md
```diff
@@ -0,0 +1,106 @@
...
+- **WHEN** a card renders outside an `InstructionSurfaceProvider`
+- **THEN** `useInstructionSurface` MUST throw
+
+### Requirement: Fields declared as data
+
+A card's rows SHALL be declared as `InstructionField` descriptors rather than as table markup. Among migrated cards, `InstructionFields` MUST be the only component that knows row markup, cell alignment, and which address renderer the current surface uses. `ProgramField` is the single sanctioned exception: it owns the markup of the leading `Program` row, because the inspector renders that row through its own validator.
+
+The union is closed: `address`, `sol`, `bytes`, `seed`, `text`, `custom`. `text` MUST accept only `string | number`, and `custom` MUST be the sole markup door. `custom` MUST take a `ReactElement` rather than a `ReactNode`, so that a labelled row with nothing in it is not representable; a field with no value to show MUST be omitted from the list instead.
+
+`InstructionFieldList` MUST admit `false` and `undefined` so optional rows read as `cond && address(...)`, and MUST NOT admit `null`, which `unicorn/no-null` forbids callers from writing anywhere under `app/**`.
```
rogaldh:
```suggestion
`InstructionFieldList` MUST admit `false` and `undefined` so optional rows read as `cond && address(...)`, and MUST NOT admit `null`, which `unicorn/no-null` forbids in card sources under `app/**` (the rule is relaxed only for tests and stories).
```

I'd use that or drop it to respect the project-wide rule

rogaldh (dismissed):
LGTM

openspec/changes/unify-instruction-card-rendering/proposal.md
askov: Applied in 0040ac74.

askov: Fixed in 0040ac74. Rather than permalink each site, I dropped the line numbers and named the symbols instead — `inspector/InstructionsSection.tsx` calls `dispatcher.fromTransactionInstruction(ix)` and routes its `SYSTEM_PROGRAM_LABEL` case. That is what your suggestion below does, and it cannot go stale at all, whereas a permalink pins a revision that stops matching `master`. Both remaining line-number sites in the proposal are gone; `grep -n ':[0-9]' openspec/changes/unify-instruction-card-rendering/` is now empty.

openspec/changes/unify-instruction-card-rendering/proposal.md
askov: Applied in 0040ac74. Verified the counts before taking it: 14 `result={INSPECTOR_RESULT}` prop passes in `inspector/InstructionsSection.tsx`, one of which is the `SYSTEM_PROGRAM_LABEL` case — `SystemDetailsCard` still declares `result` because its `default`/`catch` arms spread props into `components/instruction/UnknownDetailsCard`, which requires `result: SignatureResult`. So 13 of 14 belong to unmigrated cards, exactly as you wrote.

openspec/changes/unify-instruction-card-rendering/proposal.md
askov: Fixed in 0040ac74, same treatment as the `:130` thread. `:175`/`:250` and `:227`/`:403` were all stale — the IDL tier now sits at 171/188 and the `SYSTEM_PROGRAM_LABEL` case at 262. The bullet now states the ordering by symbol: the inspector returns `IdlInstructionCard` as soon as `useIdlInstructionDecode` resolves and reaches its `SYSTEM_PROGRAM_LABEL` case only when it does not; the tx page runs its program switch first and falls through to the IDL tier. Same claim, nothing left to actualise. The PR description carried the same four numbers and is updated too.

openspec/changes/unify-instruction-card-rendering/specs/instruction-card/spec.md
askov: Applied in 0040ac74. 49 + 7 + 2 = 58 matches the audit already in the proposal, and the 7 check out against the tree: `ComputeBudget`, `token/TokenDetailsCard`, `UnknownDetailsCard`, `ZkElGamalProof`, `AssociatedToken`, `Pmp`, `MetaplexTokenMetadata`.

openspec/changes/unify-instruction-card-rendering/specs/instruction-card/spec.md
askov: Applied in 0040ac74.

openspec/changes/unify-instruction-card-rendering/specs/instruction-card/spec.md
askov: Took the substance in 0040ac74, with one correction — I checked `eslint.config.mjs` and "relaxed only for tests and stories" is not quite right. `unicorn/no-null` is `error` globally and off for `TEST_AND_STORY_FILES`, but there is also a per-file legacy exemption list (the TODO cleanup block) that already contains 8 instruction card files, `AnchorDetailsCard.tsx` and `token/TokenDetailsCard.tsx` among them. Pinning "only tests and stories" in a spec would be false the moment someone greps for it.

So the line now reads:

> ... and MUST NOT admit `null`, which `unicorn/no-null` forbids in card sources under `app/**` (the rule is off for tests and stories, and for a per-file legacy list that a migrated card MUST NOT join).

That keeps your point — say where the rule actually binds instead of overstating it — and turns the exemption list into a constraint on migrations rather than a loophole. Happy to drop the parenthetical entirely if you prefer the shorter form.

Woody4618 (approved):

PR #1240, opened 2026-08-21
refactor(inspector): replace tweetnacl with kit verifySignature

amilz:
Migrate signature verification from tweetnacl to WebCrypto via @solana/kit.

- `SignaturesCard`: verification moves from nacl's sync `sign.detached.verify` to kit's async `verifySignature` (WebCrypto). Results are identity-guarded against the props they were computed from, so a new transaction never renders the previous transaction's verdicts. A wrong-length signature renders Invalid; an environment without WebCrypto Ed25519 renders N/A rather than a false verdict.
- `sign-web3js-transaction.spec.ts`: wallet-fake signing/verification moves to kit (`generateKeyPairSigner`, `signBytes`, `verifySignature`).
- New `SignaturesCard.spec.tsx` covering valid, invalid, malformed, and missing signatures.

[Test preview
](https://explorer-git-refactor-hoo-1167-tweetna-a89eb5-solana-foundation.vercel.app/)`/tx/inspector` with a signed transaction: Valid badge per verified signer; mess up a signature should be Invalid;

Closes: HOO-1167 (Task 0.9)

### DIFF
--- a/app/components/inspector/SignaturesCard.tsx
+++ b/app/components/inspector/SignaturesCard.tsx
@@ -1,15 +1,41 @@
 import { Address } from '@components/common/Address';
 import { Signature } from '@components/common/Signature';
-import { getBase58Encoder } from '@solana/kit';
+import {
+    getBase58Encoder,
+    getPublicKeyFromAddress,
+    isSolanaError,
+    signatureBytes,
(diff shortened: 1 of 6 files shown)


rogaldh (approved):
LGTM

PR #1241, opened 2026-08-22
refactor(domains): replace @onsol/tldparser with vendored kit ANS module

amilz:
rolled our own onsol parser to drop 2 deps: @onsol/tldparser and borsh
they do have a [kit SDK](https://www.npmjs.com/package/@onsol/tldparser-kit) but it has old version of kit as a devdep (not peer) so this would have created some conflicts in our codebase (and duplicates of everything). 
@Woody4618 happy to discuss the approach. 

closes HOO-1197


Woody4618 (changes requested):
I dont think its a good idea to maintain that ourselfs but instead maybe we can use the new version? https://www.npmjs.com/package/@onsol/tldparser

Woody4618 (approved):

PR #1249, opened 2026-08-26
Add simd 553 fee comparison to transaction page

Woody4618:
## Description

Added a comparison fo fees under SIMD 0553

image


app/features/transaction/ui/SummaryCard.tsx
rogaldh, line 176: The fee row in Summary displays the actual fee, including priority fees, not the price for a single signature.

Here is the Tx with priority fees: https://explorer.solana.com/tx/2Vr1o2mL4531v9DuYtkGdTnC9f7DpJZP3N2jwkBm7huGkopfNoYXPQ1c2nz4P1nvtzXW52dfycJoGKjWUo6hMqoX

The fee under SIMD uses costModel and does not account for priority fees. That might produce wrong percentages

rogaldh (approved):

PR #1252, opened 2026-08-27
fix(supply): read chain supply through a cached server route

askov:
## Description

The home page blocked ~8 s on `getSupply` — a full-ledger scan, run uncached from the browser on every load, and only after the genesis-hash check finished.

The figure is identical for every visitor, so `/api/supply` now serves it from the CDN, like `/api/ping` and `/api/stake-rewards`. Custom and localhost endpoints keep asking RPC directly, because the server must never be aimed at a client-supplied URL.

Two things followed from that:

- **`SupplyProvider` became an SWR hook.** Three of its five states existed only to describe waiting on the cluster connection, which the request no longer does.
- **`useVoteAccounts` moved to SWR too**, keyed by cluster and endpoint. Supply used to be the slow half, and its wait hid that vote accounts sat in unkeyed `useState`: switching cluster left the previous cluster's stake on screen, against the new cluster's supply.

The cluster entity gains `connectableUrl` — the endpoint a fetch may actually use, absent until a custom URL is settled. It is a branded string, so reaching for the plain `url` beside it no longer compiles.

Visible changes: the supply card stays up while staking data is still coming, cluster switches show skeletons instead of stale numbers, and zero delinquent stake reads `Delinquent stake: 0.0%` instead of hiding the row — which had made a healthy cluster and an unknown one look alike.

## Type of change

-   [x] Bug fix

## Testing

- [Home page — Circulating Supply and Active Stake render after one cached round trip](https://explorer-git-fork-hoodieshq-fix-cache-f7b120-solana-foundation.vercel.app/)
- [Devnet — figures belong to the cluster in the URL, and switching between them shows skeletons rather than stale numbers](https://explorer-git-fork-hoodieshq-fix-cache-f7b120-solana-foundation.vercel.app/?cluster=devnet)
- [Custom cluster — the consent prompt, and no supply request until it is answered](https://explorer-git-fork-hoodieshq-fix-cache-f7b120-solana-foundation.vercel.app/?cluster=custom&customUrl=https://my-node.example)
- [The route itself — supply as decimal strings, with the CDN cache headers](https://explorer-git-fork-hoodieshq-fix-cache-f7b120-solana-foundation.vercel.app/api/supply?cluster=0)

## Related Issues

Closes [HOO-1048](https://linear.app/solana-fndn/issue/HOO-1048/home-page-blocks-8s-on-getsupply-loading-supply-data-loading-staking)

## Checklist

## Additional Notes

The rationale — the cache TTLs, the one canonical query shape, why each failure reports at the tier it does, and the consent rule behind `connectableUrl` — is in [`openspec/changes/cache-supply-behind-server-route/proposal.md`](https://github.com/hoodieshq/explorer/blob/1977faa22f2a8a3367e7a195fb32b6498e67ad86/openspec/changes/cache-supply-behind-server-route/proposal.md).

Three things worth a reviewer's attention, since they reach past supply:

- **`errors.ts` gains connection-level retryable codes, and that classifier is shared.** `/api/idl-latest` and `/api/security-txt` import `isRetryableError` too, so four codes move from `panic` to `warn` on those routes as well. Deliberate — a DNS blip should not wake anyone — but it is a behaviour change outside this PR's title.
- **One fix here is older than this PR.** `SolanaClusterStatsProvider` guarded its polling on `url`, which always resolves, so the home page contacted the fallback endpoint while a custom URL was still awaiting consent. It now gates on `connectableUrl`.
- **Failures are cached in the visitor's own browser, not at the CDN.** A shared cache holds only a handful of status codes and no error status is among them, so `s-maxage` does nothing for a `5xx`. The per-visitor bound stops one visitor's retries from each reaching the node; nothing bounds them across visitors.

Home page First Load JS goes 520 kB → 530 kB. Several unrelated routes show +10 kB in `bench/BUILD.md` as well; the shared chunk grew slightly and those routes sat near a rounding boundary.


app/components/StakingSection.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    const circulatingPercentage = percentage(supply.circulating, supply.total, 2).toFixed(1);
+
+    return (
+        <div className="flex flex-col md:flex-row md:gap-6">
+            <div className="w-full md:w-1/2">
+                <Card ui="dashkit" className="mb-3 md:mb-6">
+                    <CardBody ui="dashkit">
+                        <h4>Circulating Supply</h4>
+                        <h1 className="mb-3">
```
rogaldh, line 52: A bit strange that h1 contains the supply data (semantically)

rogaldh (dismissed):
LGTM

app/components/StakingSection.tsx
askov: I agree but it comes directly from production version, I tried to preserve as much as possible

askov: Created a task for it https://linear.app/solana-fndn/issue/HOO-1388/home-page-heading-outline-the-only-h1s-are-numbers

Woody4618 (approved):

PR #1251, opened 2026-08-27
feat(search): match feature gates by SIMD number

askov:
## Description

Feature gates registered before the SIMD process carry no SIMD number in their title — `MoveStake and MoveLamports` is SIMD 148 — so the search bar could not reach them at all. This matches the registry's `simds` field alongside the title.

Accepted spellings: `148`, `0148`, `simd148`, `simd 148`, `SIMD-0148`. Registry entries are zero-padded and carry stray upstream whitespace (`['0178', ' 0189', ' 0377']`), so both sides of the comparison parse to a decimal number.

A SIMD number matches exactly, not as a prefix, because it is frequently a valid epoch or block number too — 35 of the numbers now reachable are past Solana epochs. For the same reason the provider's priority drops below the block and epoch providers in the local tier, so typing `219` still lists `Epoch #219` above the feature gate.

Query parsing lives in `app/features/search/lib/parse-simd-number.ts`, beside the existing `parse-natural-number.ts`, where it is covered without touching the registry. `parseNaturalNumber` itself cannot serve here: it rejects zero-padding, which the registry stores.

## Type of change

-   [x] New feature

## Testing

Manual testing on the preview deployment — open the search bar and type the query:

- [`148` — a gate reachable only by its number](https://explorer-git-fork-hoodieshq-feat-featu-aaf29e-solana-foundation.vercel.app/) → Feature Gates group with `MoveStake and MoveLamports`, whose title never mentions 148
- [`0148`, `simd148`, `simd 148`, `SIMD-0148` — every spelling](https://explorer-git-fork-hoodieshq-feat-featu-aaf29e-solana-foundation.vercel.app/) → each returns that same single row
- [`162` — one number, two gates](https://explorer-git-fork-hoodieshq-feat-featu-aaf29e-solana-foundation.vercel.app/) → two rows, both titled `Remove accounts executable flag checks`, on different addresses
- [`219` — a SIMD number that is also an epoch](https://explorer-git-fork-hoodieshq-feat-featu-aaf29e-solana-foundation.vercel.app/) → `Epoch #219` listed above the Feature Gates group
- [`simd17` — a prefix, not a number](https://explorer-git-fork-hoodieshq-feat-featu-aaf29e-solana-foundation.vercel.app/) → nothing. SIMD 173 and 174 both exist, but a prefix of a SIMD number must not reach them
[...]

PR #1253, opened 2026-08-27
refactor(instruction-card): migrate stake cards to the surface

askov:
## Description

Migrates all 16 Stake instruction cards onto the `instruction-card` surface introduced in #1239. Phase A of the follow-up epic — Stake is the largest tx-page-only program, so it exercises the field DSL at scale with no inspector chrome to review.

Cards now declare `InstructionField` descriptors instead of drawing their own rows. `DetailRow.tsx` and the `StakeCardProps` spread are gone; each card takes one `InstructionNode` and gets its frame, address renderer, and `Program` row from the surface. Leaf cards go from 708 to 422 lines.

**Two entity additions the migration forced:**

- `address()` and `FieldAddress` now accept `PublicKey | Address`. Stake is kit-native — `DetailRow` took `@solana/kit` `Address` and rendered `KitAddress`. `InstructionFields` coerces in one place, the same single-conversion rule `KitAddress` already follows, so kit slices stay kit-native. Vote and Address Lookup Table hit this next; the alternative was `toLegacyPublicKey()` at ~35 call sites.
- A `timestamp` field kind for the two lockup timestamps. It takes unix seconds and owns the UTC conversion, so neither card spells out a date pipeline or imports `@utils/date`.

`Epoch` rows use `custom()` rather than a new kind: two sites, both needing a child component. Worth promoting if Vote repeats it.

**One deliberate rendering change.** The Authority Seed row moves onto the existing `seed()` kind, so it gains a copy affordance and renders inside `<code>`. Master drew it as a bare string in a `font-mono` cell. An earlier revision of this PR added a `mono` kind to preserve that exactly, but a kind that names a font rather than a meaning is the thing the field DSL exists to prevent — and `seed()` is what the four System `*WithSeed` cards already use, so taking it makes the seed rows consistent across programs. This is the only user-visible difference in the PR.

**Three cards took a `title` or `variant` prop from the dispatcher.** Those are now two cards from one shared field list (`AuthorizeWithSeed`/`AuthorizeCheckedWithSeed`, `SetLockup`/`SetLockupChecked`, `MoveStake`/`MoveLamports`), so the title lives in the card.

### Import cycle in `providers/accounts`

`app/providers/accounts/index.tsx` imported the `@features/stake` barrel. Once the cards imported the instruction-card entity, that closed a cycle — entity → … → `providers/accounts` → stake barrel → cards → entity — and `defineInstructionCard` was `undefined` at module init, breaking two unrelated specs. Fixed by importing the two leaf modules directly. The file already does exactly this for `transaction-history` and `vote`, with the same rationale; stake was the outlier. The stake barrel no longer re-exports `getStakeActivation` or `StakeAccount` at all — with the deep imports in place they had no remaining consumers, and leaving them would let a future provider import them from the barrel and silently close the cycle again. A comment on the barrel says so.

This shifts chunk boundaries. Rebased onto current master, whose `bench/BUILD.md` was regenerated in #1251, so the committed table diff is exactly this PR's effect — I confirmed it by building master's tip and the branch from clean trees and diffing the two tables:

| Route | master | branch | cause |
|---|---|---|---|
| `/block/[slot]` (+3 sub-routes) | 680 / 670 kB | 660 / 650 kB | provider change |
| `/tx/[signature]/inspect`, `/tx/inspector` | 850 kB | 840 kB | provider change |
| `/`, `/_not-found` | 520 / 400 kB | 530 / 410 kB | provider change |
| `/address/[address]` | 890 kB | 900 kB | provider change |
| `/address/[address]/anchor-program`, `/address/[address]/feature-gate` | 860 kB | 870 kB | provider change |
| `/address/[address]/instructions` | 870 kB | 880 kB | card migration |

First Load JS, `build:info` rounds to 10 kB. Builds are deterministic — I confirmed two builds of the same tree produce an identical table. Breaking the cycle moves weight off the block and inspector routes and onto the shared chunk; the card migration itself only touches the instructions route. Flagging it since the landing page is in the +10 kB group.

## Type of change

-   [x] Other: refactor — one deliberate rendering change, called out above

## Screenshots

One row changes: the Authority Seed row on `AuthorizeWithSeed` / `AuthorizeCheckedWithSeed` gains a copy icon and renders as `<code>`, matching the System `*WithSeed` cards. Everything else is unchanged — `ProgramField` with the default `showExtendedInfo` resolves to the same `<Address pubkey alignRight link />` the old `StakeProgramRow` produced, and every other label, value, and row order is identical. Pinned by the tests below rather than by screenshots.

## Testing

- [Initialize + Delegate — address rows, lockup epoch](https://explorer-git-fork-hoodieshq-refactor-h-48370c-solana-foundation.vercel.app/tx/hy7ikw21fEe2ithWArnWXf65gH96Lb7WkTmpW8ezeeNAFf4FQeou6UyAnEqiV2GZS6KWeTkGGK8R1B6UBWc1PH2)
- [Withdraw — SOL amount row](https://explorer-git-fork-hoodieshq-refactor-h-48370c-solana-foundation.vercel.app/tx/3BWirfz9vFv3ELLvEobHi3QiGuWVaUoBCkKitunRx9V67rEksxb1PpSaJTvWyQHNoCt2bcfkqiW2DDuoUBhSygQC)
- [Authorize ×10 — authority-type text row, optional custodian](https://explorer-git-fork-hoodieshq-refactor-h-48370c-solana-foundation.vercel.app/tx/4aNWbwfxbMJa3v1SjRSFkVh8JmcmgM2whcymWLcXMGVFbzy9oWes5WFRNSVE471HhM5nmDzQz8EqRXUoPkDWrzrz)
- [Initialize Checked](https://explorer-git-fork-hoodieshq-refactor-h-48370c-solana-foundation.vercel.app/tx/4WWLVqNqQBGNvQ3MD4J5Y8g8Mae8pDm2Z43dcUr9Jdt6zJuA7xPhgi4cascNiBoP4wrSEm5TPAxAZJYXXSn52RgD)
- [Deactivate](https://explorer-git-fork-hoodieshq-refactor-h-48370c-solana-foundation.vercel.app/tx/5faR9xVdBJbXinKVtuih4ZL48fkdTHrjRGQEnUrC9NTuZhHA1yDJ1BCHUdLY7gEvFZDZqeMpUtjxyZR9zsNHLLUa)

I could not find a recent mainnet `SetLockup` or `AuthorizeWithSeed` to link, so the authority-seed and lockup-timestamp rows are covered by the specs rather than by a preview link.

## Related Issues

Closes [HOO-1268](https://linear.app/solana-fndn/issue/HOO-1268/migrate-stake-instruction-cards-to-the-instruction-card-surface)

Part of [HOO-1267](https://linear.app/solana-fndn/issue/HOO-1267/epic-unify-instruction-card-rendering-across-the-tx-page-and-inspector)

## Checklist

## Additional Notes

Stake had **no** test coverage for any instruction card before this — one spec in the whole feature, none for the cards, no stories. Nothing pinned the migration, so this PR adds three specs, 40 tests:

- `__tests__/StakeInstructionCards.spec.tsx` pins every row of all 16 cards — label, order, count, **and the value each row resolves to**. Address rows are read from `data-address` so truncation does not hide them; every other kind is read from its rendered text, so a wrong SOL amount or authority type fails rather than reading as an empty cell. Covers both branches of every optional row (custodian on three cards, Merge's independently-optional sysvars, all three lockup args), the epoch and the UTC-converted timestamp, the system-program custodian omission, and the copyable-code seed row.
- `__tests__/StakeDetailsCard.spec.tsx` pins the dispatcher. Paired cards render near-identical rows, so a mis-routed type would still produce a plausible card.
- `__tests__/RawStakeDetailsCard.spec.tsx` pins the raw discriminator path — all three `classifyRawStakeInstruction` branches, plus the `InstructionNode` that card hand-builds.

Both zero-boundary cases are covered, since `SetLockup` guards on `!== undefined` while `Initialize` guards on `> 0` — an asymmetry inherited from master that now sits in two adjacent files.

I verified the assertions bite rather than assuming, by mutating the source and confirming a failure each time: transposing `source`/`destination` in `MergeDetailsCard` fails 3 tests; pointing `moveStake` at the wrong card fails the dispatch test; dropping `unixTimestampToMs` from the renderer fails the timestamp test; zeroing a `sol()` amount or swapping an authority type for another field fails 1–2 tests each; and pointing either the `Program` row or the raw card's node at a hardcoded program id — rather than at the instruction — fails too. Removing a kind from the field renderer's style map fails the **build**, not a test.

`openspec/` is deliberately untouched here. The delta for `timestamp` and `FieldAddress` lands separately, so this PR is code and tests only.

Per the spec delta, a program migrates whole — every Stake card is in this PR, so the inspector never sees two frames side by side. (Stake is tx-page-only regardless.)


rogaldh (approved):
LGTM

PR #1254, opened 2026-08-27
refactor(providers): move tokens and rewards providers off web3.js Connection

amilz:
Replaces the `new Connection(...)` constructions in the tokens and rewards account sub-providers with the central kit RPC accessor, swapping the transport while leaving the providers' public APIs and cache semantics unchanged.

Token holdings now query both token programs concurrently rather than sequentially. kit upcasts integral fields of a jsonParsed response to bigint unless the key path is allow-listed, so token-2022 extension state runs through `withNumbersInsteadOfBigInts` and inflation rewards are mapped field by field.

Closes: HOO-1318

### DIFF
--- a/app/providers/accounts/rewards.tsx
+++ b/app/providers/accounts/rewards.tsx
@@ -1,14 +1,16 @@
 'use client';
 
+import { getRpc } from '@entities/cluster';
 import { ActionType } from '@providers/block';
 import * as Cache from '@providers/cache';
 import { FetchStatus } from '@providers/cache';
 import { useCluster } from '@providers/cluster';
-import { Connection, InflationReward, PublicKey } from '@solana/web3.js';
+import { InflationReward, PublicKey } from '@solana/web3.js';
 import { Cluster } from '@utils/cluster';
 import React from 'react';
 
 import { Logger } from '@/app/shared/lib/logger';
+import { toKitAddress } from '@/app/shared/lib/web3js-compat';
 
 const REWARDS_AVAILABLE_EPOCH = new Map<Cluster, number>([
     [Cluster.MainnetBeta, 132],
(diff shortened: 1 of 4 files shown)


askov (approved):

rogaldh (approved):

app/providers/accounts/tokens.tsx
```diff
@@ -63,22 +66,27 @@ async function fetchAccountTokens(dispatch: Dispatch, pubkey: PublicKey, cluster
...
+            rpc
+                .getTokenAccountsByOwner(
+                    owner,
+                    { programId: toKitAddress(programId) },
+                    { commitment: 'processed', encoding: 'jsonParsed' },
+                )
+                .send();
+
+        const [{ value: tokenAccounts }, { value: token2022Accounts }] = await Promise.all([
+            fetchByProgram(TOKEN_PROGRAM_ID),
```
Woody4618, line 81: Does fetch by program need a commitment?

app/providers/accounts/rewards.tsx
```diff
@@ -89,12 +91,13 @@ async function fetchRewards(
...
     const lowestAvailableEpoch = REWARDS_AVAILABLE_EPOCH.get(cluster) || 0;
-    const connection = new Connection(url);
+    const rpc = getRpc(url);
+    const address = toKitAddress(pubkey);
 
     if (!fromEpoch) {
         try {
-            const epochInfo = await connection.getEpochInfo();
-            fromEpoch = epochInfo.epoch - 1;
+            const epochInfo = await rpc.getEpochInfo().send();
```
Woody4618, line 99: Do we need to set commitment here?

Woody4618 (approved):

app/providers/accounts/tokens.tsx
amilz: It's in there, matching legacy (`commitment: 'processed'`) line 76.

app/providers/accounts/rewards.tsx
amilz: legacy did not specify and just went to default. following the same here.

PR #1261, opened 2026-08-31
refactor(instruction-card): migrate address lookup table cards to the surface

askov:
## Description

All five Address Lookup Table instruction cards drew their own rows, each with a hardcoded `Program` row and its own `InstructionCard` import. ALT is tx-page only — the inspector has no ALT case.

Drawing rows by hand also hardcodes the transaction page's address renderer, which keeps these cards out of the inspector. Phase A of #1267, after Stake (#1253).

### The change

-   **Declare rows as descriptors.** All five cards return `InstructionField` lists and take one `InstructionNode`, so the surface owns the frame, the address renderer and the `Program` row.
-   **Validate each type against its own schema.** `type()` ignores unknown keys, so the two-field `Freeze` schema matched every payload and won the union — Create, Extend and Close never ran.
-   **Coerce extend addresses in the schema.** `newAddresses` becomes `array(PublicKeyFromString)`, so the card stops calling `new PublicKey` itself.
-   **Fall back rather than render a half-validated card.** A payload missing a required field now routes to `UnknownDetailsCard`; master rendered a card.
-   **Right-align the nested new-addresses table.** It builds on `BaseTable` so it spans the value cell; a bare `<table>` shrank to its contents and left entries 228px short of the rows above.

## Type of change

-   [x] Other: refactor — carries one validation tightening and one alignment fix, both listed above

## Testing

-   **The new-addresses entries right-align.** Open the extend transaction — [preview](https://explorer-git-fork-hoodieshq-refactor-a-ffc2e0-solana-foundation.vercel.app/tx/2p1wFbeNGDNEDdetzyacZ6tnDtQdL2hNPJ8heP2ss22KH8yBfY6Sx6WURzsSbAh2Azmtav7bDfrtewnE9yvfFtd1) · [production](https://explorer.solana.com/tx/2p1wFbeNGDNEDdetzyacZ6tnDtQdL2hNPJ8heP2ss22KH8yBfY6Sx6WURzsSbAh2Azmtav7bDfrtewnE9yvfFtd1) — and read the `New Addresses` row of instruction `#3`. Expect all 13 entries on the same right edge as the `Lookup Table` and `Lookup Table Authority` rows above them. Production sits them well short of that edge.
-   **Create keeps its payer, recent slot and bump seed.** Open the create transaction — [preview](https://explorer-git-fork-hoodieshq-refactor-a-ffc2e0-solana-foundation.vercel.app/tx/351nK5vkcW7oPvgLMnK437XsvrTxS1REY6PDbeSBu9g3Yrs6Gyc6qZtaS9KGHpdg9dcky3X9CxpqxZjGKBDb5G68) · [production](https://explorer.solana.com/tx/351nK5vkcW7oPvgLMnK437XsvrTxS1REY6PDbeSBu9g3Yrs6Gyc6qZtaS9KGHpdg9dcky3X9CxpqxZjGKBDb5G68) — and compare every row against production. Expect the same labels, values and order, and the recent slot still linking to its block page.
-   **Close still shows the recipient the tightening requires.** Open the close transaction — [preview](https://explorer-git-fork-hoodieshq-refactor-a-ffc2e0-solana-foundation.vercel.app/tx/3ajPHGNTzZKmKBq14Uo4QkTevqyWg6aUrRuhJuZ9qwC3PAjkYCVfMVPxzLR6xsrHrLttAtGf9Awn1qXCGjr1MSDG) · [production](https://explorer.solana.com/tx/3ajPHGNTzZKmKBq14Uo4QkTevqyWg6aUrRuhJuZ9qwC3PAjkYCVfMVPxzLR6xsrHrLttAtGf9Awn1qXCGjr1MSDG) — and read the card. Expect the same three rows as production, not the `Unknown` fallback.
-   **Deactivate renders unchanged.** Open the deactivate transaction — [preview](https://explorer-git-fork-hoodieshq-refactor-a-ffc2e0-solana-foundation.vercel.app/tx/4jW6ZdYSPoQTZS6uswPXkniZR7PSKrn1acPBJwWw6Acy2mWz5gWST7GSsAqrwQcwrNRNSrid23tD49pQhdM4wVux) · [production](https://explorer.solana.com/tx/4jW6ZdYSPoQTZS6uswPXkniZR7PSKrn1acPBJwWw6Acy2mWz5gWST7GSsAqrwQcwrNRNSrid23tD49pQhdM4wVux) — and compare the card against production. Expect identical rows and the same title.

## Related Issues

Closes [HOO-1396](https://linear.app/solana-fndn/issue/HOO-1396/migrate-address-lookup-table-instruction-cards-to-the-instruction-card)

Part of [HOO-1267](https://linear.app/solana-fndn/issue/HOO-1267/epic-unify-instruction-card-rendering-across-the-tx-page-and-inspector)

## Checklist

## Additional Notes

Two specs, 17 tests:

-   `__tests__/AddressLookupTableInstructionCards.spec.tsx` pins every row of all five cards — label, order, count and the value each row resolves to. Address rows are read from `data-address`, so truncation cannot hide a wrong value. It also pins the nested table's `[index, address]` pairs, the recent slot's link target, and — rendering Extend under a stub surface — that every address goes through the surface's renderer rather than a hardcoded one.
-   `__tests__/AddressLookupTableDetailsCard.spec.tsx` pins the dispatcher. Freeze, Deactivate and Close render identical rows, so a mis-routed type still produces a plausible card and only the title tells them apart. It covers all five routes, both fallback paths, and the extend coercion the old union skipped.

### DIFF
--- a/app/components/instruction/address-lookup-table/ExtendLookupTableDetails.tsx
+++ b/app/components/instruction/address-lookup-table/ExtendLookupTableDetails.tsx
@@ -1,58 +1,42 @@
-import { InstructionDetailsProps } from '@features/transaction';
-import { AddressLookupTableProgram, PublicKey } from '@solana/web3.js';
+import { address, custom, defineInstructionCard, useInstructionSurface } from '@entities/instruction-card';
+import type { PublicKey } from '@solana/web3.js';
 
-import { Address } from '@/app/components/common/Address';
-import { InstructionCard } from '@/app/components/instruction/InstructionCard';
 import { BaseTable } from '@/app/shared/ui/Table';
 
 import { ExtendLookupTableInfo } from './types';
 
-export function ExtendLookupTableDetailsCard(props: InstructionDetailsProps & { info: ExtendLookupTableInfo }) {
-    const { ix, index, result, innerCards, childIndex, info } = props;
+export const ExtendLookupTableDetailsCard = defineInstructionCard<ExtendLookupTableInfo>({
+    fields: info => [
+        address('Lookup Table', info.lookupTableAccount),
+        address('Lookup Table Authority', info.lookupTableAuthority),
+        custom('New Addresses', <NewAddresses addresses={info.newAddresses} />),
+    ],
+    title: 'Address Lookup Table: Extend Lookup Table',
+});
+
+/**
+ * A whole table in one cell, which no field kind describes — so it takes the
+ * `custom` door and reads the address renderer off the surface itself, the way
+ * `InstructionFields` would.
+ */
+function NewAddresses({ addresses }: { addresses: PublicKey[] }) {
+    const { Address } = useInstructionSurface();
+
     return (
-        <InstructionCard
-            ix={ix}
-            index={index}
-            result={result}
-            title="Address Lookup Table: Extend Lookup Table"
-            innerCards={innerCards}
-            childIndex={childIndex}
-        >
-            <BaseTable.Row>
-                <BaseTable.Cell>Program</BaseTable.Cell>
-                <BaseTable.Cell className="text-right">
-                    <Address pubkey={AddressLookupTableProgram.programId} alignRight link />
-                </BaseTable.Cell>
-            </BaseTable.Row>
-            <BaseTable.Row>
-                <BaseTable.Cell>Lookup Table</BaseTable.Cell>
-                <BaseTable.Cell className="text-right">
-                    <Address pubkey={info.lookupTableAccount} alignRight link />
-                </BaseTable.Cell>
-            </BaseTable.Row>
-            <BaseTable.Row>
-                <BaseTable.Cell>Lookup Table Authority</BaseTable.Cell>
-                <BaseTable.Cell className="text-right">
-                    <Address pubkey={info.lookupTableAuthority} alignRight link />
-                </BaseTable.Cell>
-            </BaseTable.Row>
-            <BaseTable.Row>
-                <BaseTable.Cell>New Addresses</BaseTable.Cell>
-                <BaseTable.Cell style={{ paddingRight: '1rem' }}>
-                    <table>
-                        <BaseTable.Body>
-                            {info.newAddresses.map((address, index) => (
-                                <BaseTable.Row key={address.toString()}>
-                                    <BaseTable.Cell className="w-px font-mono">{index}</BaseTable.Cell>
-                                    <BaseTable.Cell className="text-right">
-                                        <Address pubkey={new PublicKey(address)} alignRight link />
-                                    </BaseTable.Cell>
-                                </BaseTable.Row>
-                            ))}
-                        </BaseTable.Body>
-                    </table>
-                </BaseTable.Cell>
-            </BaseTable.Row>
-        </InstructionCard>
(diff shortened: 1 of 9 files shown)


app/components/instruction/address-lookup-table/ExtendLookupTableDetails.tsx
```diff
@@ -1,58 +1,40 @@
...
-                                        <Address pubkey={new PublicKey(address)} alignRight link />
-                                    </BaseTable.Cell>
-                                </BaseTable.Row>
-                            ))}
-                        </BaseTable.Body>
-                    </table>
-                </BaseTable.Cell>
-            </BaseTable.Row>
-        </InstructionCard>
+        <table>
```
mikhd: Should there be `BaseTable` instead of table?
Also I noticed that "New addresses" are `left` aligned in the instruction table-cell. See [2p1wFbeNGDNEDdetzyacZ6tnDtQdL2hNPJ8heP2ss22KH8yBfY6Sx6WURzsSbAh2Azmtav7bDfrtewnE9yvfFtd1](https://explorer-git-fork-hoodieshq-refactor-a-ffc2e0-solana-foundation.vercel.app/tx/2p1wFbeNGDNEDdetzyacZ6tnDtQdL2hNPJ8heP2ss22KH8yBfY6Sx6WURzsSbAh2Azmtav7bDfrtewnE9yvfFtd1) tx. Could you please double-check?

Image

mikhd (changes requested):

app/components/instruction/address-lookup-table/ExtendLookupTableDetails.tsx
askov: Both were the same bug, fixed in 960a2c77.

You were right that it should be `BaseTable` — and that is exactly what caused the left alignment. `BaseTableRoot` applies `w-full`; a bare `<table>` has none, so it shrank to its contents (481px inside a 725px cell). `Address` already right-aligns via `flex w-full md:justify-end`, but it was aligning inside a cell that was only as wide as the text, so the whole nested table hugged the left edge — entries ended **228px short** of where every other row ends.

Switching the root to `BaseTable` closed most of it and left 24px. That residue was the card table's edge padding: `[&_tbody_td:last-child]:pr-6` matches descendants, so it lands on the nested cells too. Zeroing it needs `tbody tr td` to outweigh that selector — the same specificity trick `BaseTable`'s own `body="subtle"` variant uses.

Measured on your tx after the fix, all 13 entries and both rows above them land on the same edge:

```
siblingRowAddrRight: [992, 992]   // Lookup Table, Lookup Table Authority
nestedAddrRights:    [992]        // all 13 entries
```

Full suite green (419 files, 4714 tests). `bench/BUILD.md` untouched — a local regen moves two unrelated address routes by 10kB, but I confirmed that identical diff on the pristine base with my change stashed, so it is pre-existing drift, not this PR.

mikhd (approved):

app/components/instruction/address-lookup-table/ExtendLookupTableDetails.tsx
askov: Same as on production. I fixed it look a bit better

Woody4618 (approved):

PR #1262, opened 2026-08-31
feat(holdings): order token holdings by verification status

askov:
## Description

Relates to [HOO-974](https://linear.app/solana-fndn/issue/HOO-974/add-sorting-for-the-tokens-by-verification). Orders an address's token holdings by verification status, and resolves every held mint before the card renders so the order is known on the first paint.

## Changes

- Resolves every held mint up front through `/api/token-info`, replacing the per-row lookup the card made as rows scrolled into view.
- Adds `orderMintsByVerification`, which splits mints into verified, listed-but-unverified, and unlisted, and keeps RPC order inside each tier.
- Adds `useTokenInfos` for callers that need every mint before they can render.
- Adds `fetchTokenInfos`, which chunks the mint list by `TOKEN_INFO_REQUEST_LIMIT` and resolves at most 6 chunks at a time.
- Sorts mints before chunking, so the same holdings produce the same request bodies whatever order the RPC returned them in.
- Keys the lookup on the resolved chain id rather than `(cluster, genesisHash)`.
- Skips the on-chain Metaplex fallback on this path: a mint the list does not carry is absent, and that absence is the answer the ordering needs.
- Answers a failed upstream list lookup with `503` in `/api/token-info`, where it answered `200` and an empty `content` array.
- Keeps answering `200` with what resolved when the lookup drops only some mints.
- Bounds both ends of the call: `createAbortSignal` on the route's upstream request, and a 15s per-chunk timeout in the browser.
- Keeps the chunks that resolved when one chunk rejects, which `fetchAll` used to discard as a `Promise.all`.
- Checks the response `content` is an array before iterating it.
- Raises `TOKEN_INFO_REQUEST_LIMIT` from 128 to 256, one constant read by both the route's cap and the client's chunk size.
- Makes `TokenRow` presentational: metadata arrives by prop, not through `useTokenInfo` and `useCluster`.
- Collapses `HoldingsDetail` and `HoldingsSummary` into one aggregator.
- Renders rows through the `BaseTable` compound API.
- Adds `app/entities/token-info/client.ts` and exports `useTokenInfos` from it, keeping `index.ts` reachable from the server graph.
- Converts `fetchTokenInfosFromApi` to an options object.
- Removes three dead client-barrel exports and the unreachable miss branch in the card's row mapping.
- Adds `chunk` to `app/shared/lib/array.ts`, which throws on a size that is not a positive integer.
- Fixes `gen.address`, which took every byte mod 256 and so returned 256 distinct addresses for a request of 257. Seeds 0-255 are unchanged.
- Adds an openspec proposal and a `token-holdings` spec under `openspec/changes/sort-token-holdings-by-verification/`.
- Adds 55 tests covering the tiering, Load More over the reordered list, the `display=detail` column, within-tier stability, the 503 / partial-drop / all-not-found split, a rejecting chunk, per-chunk timeouts, and `chunk`'s size guard.
- Reseeds the `WithLogos` story on the bulk lookup and asserts the logos in a `play`.

## Behavior Changes

- Verified holdings lead an address's token list, then listed-but-unverified mints, then mints the list does not carry.
- Holdings inside a tier keep the order the RPC returned.
- No holding is hidden. Unrecognised mints sink to the end of the list.
- Rows carry their logo and symbol on first paint instead of filling in as they scroll into view.
- `Load More` issues no token-info request.
- The card holds its loading state until every mint resolves, rather than painting an order that reshuffles a moment later.
- `/api/token-info` answers `503` when the upstream list lookup resolves nothing.
- A stalled token-info request now times out instead of holding the card on its loading state for as long as the platform allows.

### Ordering, not filtering

On-chain every holding is equally valid, so the Explorer reports a fact from a named source rather than deciding which of a user's tokens deserve to exist. Cheaper signals were measured and rejected: only 5 of the 1051 mints hold a zero balance, and Jupiter names 74 of 100 sampled junk mints.

Measured on `86xCnPeV69n6t3DnyGvkKobf9FdN2H9oiVDdaMpo2MMY`:

| Tier | Condition | Mints |
| --- | --- | --- |
| 1 | The list returned the mint and `verified` is true | 78 |
| 2 | The list returned the mint, `verified` false or absent | 93 |
| 3 | The list did not return the mint | 880 |

### Request cost

The up-front lookup replaces the per-row one, so cost stays proportional to the holdings count:

| Unique mints | Before, first screen to fully expanded | After |
| --- | --- | --- |
| 20 | 1 to 1 | 1 |
| 520 | 1 to ~26 | 3 |
| 1051 | 1 to ~53 | 5 |

Chunks fire once, at most 6 at a time, and `Load More` then costs nothing. The regression: opening a whale address and not scrolling costs 5 requests instead of 1.

## Type of change

- [x] New feature

## Testing

- **Verified holdings lead the list.** Open the whale address's Token Holdings — [preview](https://explorer-git-fork-hoodieshq-feat-hoo-9-aff212-solana-foundation.vercel.app/address/86xCnPeV69n6t3DnyGvkKobf9FdN2H9oiVDdaMpo2MMY/tokens) · [production](https://explorer.solana.com/address/86xCnPeV69n6t3DnyGvkKobf9FdN2H9oiVDdaMpo2MMY/tokens) — and read the first screen of rows. Expect named, verified tokens at the top, each with its logo already drawn. Production lists the same holdings in RPC order, so the first screen is mostly airdropped mints.
- **Load More continues the order and costs no request.** Open the 520-mint address — [preview](https://explorer-git-fork-hoodieshq-feat-hoo-9-aff212-solana-foundation.vercel.app/address/MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2/tokens) · [production](https://explorer.solana.com/address/MJKqp326RZCHnAAbew9MDdui3iCKWco7fsK9sVuZTX2/tokens) — and click `Load More` with the Network tab open. Expect the next rows to continue the same tiers and no `/api/token-info` call. Production issues another token-info request for each batch of newly revealed rows.
- **The Detailed view orders the same way.** Open the 183-mint address and set the display dropdown to `Detailed` — [preview](https://explorer-git-fork-hoodieshq-feat-hoo-9-aff212-solana-foundation.vercel.app/address/8BseXT9EtoEhBTKFFYkwTnjKSUZwhtmdKY2Jrj8j45Rt/tokens?display=detail) · [production](https://explorer.solana.com/address/8BseXT9EtoEhBTKFFYkwTnjKSUZwhtmdKY2Jrj8j45Rt/tokens?display=detail). Expect the `Account Address` column and the same tier order as the summary view. Production shows the column in RPC order.
- **Devnet order is unchanged.** Open the whale address on devnet — [preview](https://explorer-git-fork-hoodieshq-feat-hoo-9-aff212-solana-foundation.vercel.app/address/86xCnPeV69n6t3DnyGvkKobf9FdN2H9oiVDdaMpo2MMY/tokens?cluster=devnet) · [production](https://explorer.solana.com/address/86xCnPeV69n6t3DnyGvkKobf9FdN2H9oiVDdaMpo2MMY/tokens?cluster=devnet). Expect the same order as production: the list carries no devnet entries, so every mint falls into tier 3 and keeps RPC order.

## Related Issues

Closes [HOO-974](https://linear.app/solana-fndn/issue/HOO-974/add-sorting-for-the-tokens-by-verification)

## Checklist

- [x] All checks pass locally (`pnpm test`, `pnpm lint`, `pnpm exec tsc --noEmit`, `pnpm format`)

## Additional Notes

`bench/BUILD.md` is unchanged, so this is bundle-neutral.


taethai22-tech (approved):

mikhd (approved):
LGTM!

Woody4618 (approved):

PR #1278, opened 2026-09-01
fix: HOO-1409 remove bigint tojson polyfill

amilz:
## Problem

`app/types/bigint.ts` patched `BigInt.prototype.toJSON` globally (loaded via the address layout). `JSON.stringify` invokes `toJSON` before kit's replacer, so every bigint RPC param serialized as a JSON string and the RPC rejected it with `-32602 INVALID_PARAMS`.

## Changes

- Delete the global polyfill and its layout import.
- `JsonViewer` (the reason the polyfill existed — react-json-view throws on bigints) now converts bigint primitives to strings in `src`, memoized. Display is unchanged.
- `nftoken.ts`: drop the `as unknown as bigint` cast workaround, restore real bigint offsets.
- Regression spec pins that `BigInt.prototype` has no `toJSON` and that a bigint `memcmp.offset` hits the wire as a JSON number.

Note: upstream react-json-view's `bigNumber` prop (microlinkhq/react-json-view#55) only handles BigNumber-wrapper instances, not bigint primitives, so the pre-pass is still required.

## Testing

- New spec passes; full suite 4604 passed, typecheck + lint clean.

Fixes HOO-1409


app/components/common/JsonViewer.tsx
```diff
@@ -11,12 +11,33 @@ const ReactJsonView = dynamic(() => import('@microlink/react-json-view'), {
...
 export type JsonViewerProps = ComponentProps<typeof ReactJsonView>;
 
+/**
+ * react-json-view serializes values with JSON.stringify, which throws on bigints. Decoded
+ * on-chain data (borsh u64/u128 fields and friends) routinely contains them, so replace every
+ * bigint with its decimal string before handing the tree to the viewer. Only plain objects and
+ * arrays are traversed; anything else (Uint8Array, Map, class instances) is passed through
+ * unchanged.
+ */
+function withStringsInsteadOfBigInts(value: unknown): unknown {
```
askov: **suggestion**: Reuse withNumbersInsteadOfBigInts from shared utils. If it doesn't fit for some reason, let's move it of Ui component, add some tests and explain why shared helper usage is not possible

app/entities/cluster/@x/account/index.ts
```diff
@@ -0,0 +1,3 @@
+// Cross-entity public API (FSD `@x` notation): the slice of the `cluster` entity that the
+// `account` entity is allowed to consume.
+export { getRpc, type SolanaRpc } from '../../api/get-rpc';
```
askov, line 3: **todo**: please rebase to master and check that CI still green, we've added some lint rules/fixes in this area recently

app/entities/token-info/api/fetch-token-metaplex.ts
```diff
@@ -79,16 +79,21 @@ async function fetchDecimals(
     const decimals = new Map<string, number>();
     if (mints.length === 0) return decimals;
 
-    const connection = new Connection(rpcEndpoint);
+    const rpc = createSolanaRpc(rpcEndpoint);
```
askov: **question**: Why not getRpc from entity?

askov (approved):

Woody4618 (approved):

PR #1285, opened 2026-09-03
docs(firewall): record the Vercel WAF ruleset and why each rule exists

rogaldh:
## Description

The Vercel WAF config is project state that lives in the dashboard and cannot be committed. Until now the only record
was a prose block at the tail of `docs/development.md` covering two of the eight live rules, premised on Attack
Challenge Mode being enabled — it is not; the blocker is Bot Filter at `challenge`.

This adds `docs/firewall.md` as the record: why each rule exists, why a bypass works at all, an incident ladder, and
the operator traps that have already cost us. Rules themselves are documented at the endpoint they serve, so a route
and its firewall rule change together.

Writing it surfaced a live hole, since fixed: `OG img bypass` used `Matches` `/og/*`, which Vercel evaluates as
unanchored PCRE — `*` quantifies the preceding `/`, so the pattern reduced to "contains `/og`". Every account address
and transaction signature beginning `og`, in any casing, skipped the entire WAF. It is now `Starts with` `/og/`.

Also fixed while verifying: `robots.txt`, the three sitemaps, `manifest.json`, and all twelve PWA icons were
challenged and are now covered; feature-gate link previews never unfurled because the page carrying `og:image` had no
rule, only the image did.

## Type of change

-   ~~[ ] Bug fix~~
-   ~~[ ] New feature~~
-   ~~[ ] Protocol integration~~
-   [x] Documentation update
-   ~~[ ] Other (please describe):~~

## Screenshots

No UI changes.

## Testing

Every documented rule was applied and probed against production, judging on `x-vercel-mitigated` rather than status
code — a leaking path returns a healthy-looking `404` once the WAF waves it through.

-   32-assertion sweep: 18 static assets (including all twelve `icon-*.png`), 4 crawler files, 9 near-misses that must
    stay blocked, and a control proving Bot Filter is still challenging.
-   Near-misses confirm the anchored patterns do not leak: `/faviconfoo`, `/favicon.icox`, `/icon-abc.png`,
    `/manifest.jsonx`, `/robots.txtx` all `429`.
-   OG hole closed: `/address/ogAbC…`, `/address/OGAbC…`, `/tx/og` were `200` unchallenged, now `429`, while
    `/og/receipt/…` and `/og/feature-gate/…` still reach their handlers.
-   Unfurl chains end to end: receipt and feature-gate pages return HTML with correct `og:image`, and those image URLs
    render 1200×630 PNGs.
-   `/mcp` end to end: `initialize` → `tools/list` → `ping` → `inspect_entity` on Tokenkeg returns decoded data.
-   Regex patterns verified offline with `pcre2grep` against the real `public/` filenames before applying.

## Related Issues

None.

## Checklist

-   ~~[ ] I have added tests that prove my fix/feature works~~
-   ~~[ ] I have run `build:info` script to update build information~~
-   ~~[ ] Screenshots included — required for protocol screens, recommended for other UI changes~~

## Additional Notes

The dashboard is the source of truth; these docs describe live config as of firewall version 23, which the ruleset now
matches. `Rate limit MCP` is intentionally disabled — `/mcp` is public and unmetered by design, and the rule is
switched on during a spike.

Vercel doc claims were tested and refuted, both recorded with evidence: that a `Bypass` still reaches the
managed ruleset, and that a `Log` rule terminates evaluation. Only `Bypass` exempts a path.

One known gap is documented rather than fixed: `getFeatureGateOpenGraph` also sets `og:image` on the bare
`/address/<addr>`, which no rule covers. Scoping a rule to it would mean exempting every account page, so a redirect
to the `/feature-gate` URL is the cheaper fix if that short form matters.


.gitignore
```diff
@@ -26,8 +26,8 @@ yarn-debug.log*
 yarn-error.log*
 
 # local env files
-.env
-.env*.local
```
Woody4618, line 30: should we keep env.local here?

Woody4618 (approved):

.gitignore
rogaldh: No, we need to exclude it. Covered by [.env*](https://github.com/solana-foundation/explorer/pull/1285/changes/daa7efcabb698124ae8cb70ec18e27a7860a6ea3#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R29)

PR #1287, opened 2026-09-04
feat(feedback): add the feedback component

rogaldh:
## Description

Part of [HOO-1403](https://linear.app/solana-fndn/issue/HOO-1403/collect-user-feedback-in-the-explorer-via-sentry-user-feedback). Adds an env-gated feedback widget that submits to Sentry User Feedback, and enables the browser Sentry SDK behind a reporting guard.

## Changes

-   Adds `app/features/feedback/` with the form, the floating widget, the star rating, and the inline trigger.
-   Mounts `FeedbackWidget` in `app/layout.tsx` inside `ClusterProvider`.
-   Adds a `Feedback` trigger to `app/components/Footer.tsx`.
-   Submits through `sendFeedback`, which resolves only on confirmed delivery.
-   Tags every submission with `cluster`, `rating`, `source` and `type`.
-   Adds `NEXT_PUBLIC_FEEDBACK_ENABLED` to gate both triggers.
-   Adds `NEXT_PUBLIC_FEEDBACK_ISSUES_URL` and `NEXT_PUBLIC_FEEDBACK_BUG_ISSUES_URL`, defaulting to the repo's issue templates.
-   Reads `NEXT_PUBLIC_SENTRY_DSN` as the browser DSN in `sentry/config.mjs`, and as the server/edge fallback behind `SENTRY_DSN`.
-   Adds `NEXT_PUBLIC_SENTRY_CLIENT_ERRORS` to gate all browser error reporting, defaulting to off.
-   Adds a client `beforeSend` that drops a browser error event unless that flag is on and the event carries the Logger's opt-in tag.
-   Adds `sentry/client-report.mjs` holding that tag.
-   Changes the `Logger` `sentry` option from `boolean` to `boolean | 'always'`.
-   Reports `sentry: true` from the server only.
-   Reports `sentry: 'always'` from the browser as well.
-   Keeps `Logger.panic` capturing on every runtime.
-   Adds `app/shared/lib/sentry/client.ts` for the browser-only `sendFeedback` export.
-   Exports `clusterSlug` from `@entities/cluster`.
-   Removes `.storybook/vitest.setup.ts`, superseded by addon-vitest.
-   Adds 48 specs covering the widget, the feedback envelope, the Logger gating, and the `beforeSend` guard.

## Behavior Changes

-   A floating feedback trigger appears on every page when `NEXT_PUBLIC_FEEDBACK_ENABLED` is `true`.
-   The footer gains a `Feedback` link beside `Terms of Services`.
-   Without `NEXT_PUBLIC_SENTRY_DSN` the floating trigger offers only the two GitHub links, and the footer link does not render.
-   A failed submission keeps the form open with its content and shows `Could not send feedback`.
-   No browser error event reaches Sentry unless `NEXT_PUBLIC_SENTRY_CLIENT_ERRORS` is `true`, so a DSN can ship the feedback widget on its own.
-   With that flag on, client render crashes reach Sentry through `Logger.panic`; the browser previously had no DSN, so every client capture was a no-op.

### Why the guards ship in the same PR

Setting `NEXT_PUBLIC_SENTRY_DSN` turns on the browser SDK for the first time, and two things would go live with it.

Roughly 25 latent `sentry: true` call sites, most written with the server in mind. The tag guard makes those server-only.

Four browser panic sources: `app/error.tsx`, `app/global-error.tsx`, `SentryErrorBoundary.tsx`, and the IDL mismatch in `app/features/decode-instruction-with-idl/lib/decode-instruction-with-idl.ts`, which runs inside `useIdlInstructionDecode` on transaction pages. That last one fires on chain data disagreeing with a published IDL rather than on a bug, once per instruction render, on the pages bots crawl hardest. `NEXT_PUBLIC_SENTRY_CLIENT_ERRORS` holds all four off, so the DSN can land for feedback alone and browser error sources are adopted one environment at a time.

## Type of change

-   [x] New feature

### Sentry view

context
tags

### Desktop
desktop

### Mobile
mobile

## Testing

-   **The form submits to Sentry.** Open the home page — [preview](https://explorer-git-fork-hoodieshq-feat-feedb-f97397-solana-foundation.vercel.app/) · [production](https://explorer.solana.com/) — click the round trigger at the bottom right, choose `Share feedback`, set a star rating, write a message, and submit. Expect the form to close and a `Feedback sent!` toast. Production has no trigger.
-   **The entry carries its cluster.** Submit from devnet — [preview](https://explorer-git-fork-hoodieshq-feat-feedb-f97397-solana-foundation.vercel.app/?cluster=devnet) — then open the entry in Sentry User Feedback. Expect the message, the X handle as the name, and the page URL on the entry, with `cluster`, `rating`, `source` and `type` under Tags. The Tags list is alphabetical and scrolls, so `rating` sits below `level`.
-   **The footer trigger opens the same form.** Scroll to the footer — [preview](https://explorer-git-fork-hoodieshq-feat-feedb-f97397-solana-foundation.vercel.app/) · [production](https://explorer.solana.com/) — and click `Feedback`. Expect the `Give feedback` dialog. Production's footer has only `Terms of Services`.
-   **Mobile renders a bottom drawer.** Open the home page below 640px wide — [preview](https://explorer-git-fork-hoodieshq-feat-feedb-f97397-solana-foundation.vercel.app/) — and open the form. Expect a full-width panel flush to the bottom edge, not a centered dialog.
-   **A blocked send keeps the content.** With a content blocker active, submit the form — [preview](https://explorer-git-fork-hoodieshq-feat-feedb-f97397-solana-foundation.vercel.app/). Expect the form to stay open with the message intact, and a `Could not send feedback` toast pointing at the GitHub links.

## Related Issues

Closes [HOO-1543](https://linear.app/solana-fndn/issue/HOO-1543/implement-the-feedback-report-widget-popover-on-desktop-drawer-on)

Closes [HOO-1545](https://linear.app/solana-fndn/issue/HOO-1545/feature-flag-the-feedback-widget)

Closes [HOO-1548](https://linear.app/solana-fndn/issue/HOO-1548/enable-client-side-sentry-reporting-public-dsn-browser-guard)

Part of [HOO-1403](https://linear.app/solana-fndn/issue/HOO-1403/collect-user-feedback-in-the-explorer-via-sentry-user-feedback)

## Checklist

-   ~~[ ] For security-related features, I have included links to related information~~

## Additional Notes

No existing `sentry: true` call site opts into browser reporting yet. Choosing the `sentry: 'always'` sites — and the report-only-after-retries-exhausted treatment for the SWR-retried ones — is deliberate follow-up work.

With `NEXT_PUBLIC_SENTRY_CLIENT_ERRORS` on, the `beforeSend` backstop still drops the SDK's automatic captures of uncaught errors and unhandled rejections; render crashes report through the error boundaries' `Logger.panic` instead. Passing `mechanism.handled === false` events through would reopen unhandled-error capture in one line — deliberately not done here to keep bot noise out.

The desktop form is a centered dialog rather than a popover anchored to the trigger. Aligning it with the mockups is [HOO-1544](https://linear.app/solana-fndn/issue/HOO-1544/align-the-feedback-form-design-with-current-mockups).


app/features/feedback/env.ts
mikhd, line 1: **question**: I noticed that common pattern is to name such files `env.ts`. For example, `features/feature-gate/env.ts`, `features/receipt/env.ts`, `features/token-info/env.ts`. They contain constants and "isEnabled" helpers. So was just wondering if it should be `features/feedback/env.ts` as well?

app/features/feedback/ui/FeedbackTrigger.tsx
```diff
@@ -0,0 +1,40 @@
...
+import type { ReactNode } from 'react';
+
+import {
+    BUG_REPORT_ISSUES_URL,
+    FEEDBACK_ISSUES_URL,
+    isFeedbackEnabled,
+    isFeedbackWidgetEnabled,
+} from '../lib/feedback-options';
+import { useFeedbackForm } from '../model/use-feedback-form';
+import { BaseFeedbackForm } from './BaseFeedbackForm';
```
mikhd: **question**: will this make sense to dynamically import components and some imports in useFeedbackForm, so we won't include additional code in js bundle when widget is disabled?

app/features/feedback/model/use-feedback-form.ts
```diff
@@ -0,0 +1,52 @@
+import { useCluster } from '@entities/cluster';
+import { clusterSlug } from '@utils/cluster';
```
mikhd: **question**: since this PR adds `clusterSlug` to a `app/entities/cluster/index.ts` exports, should this be imported from `@entities/cluster` instead?

mikhd (approved):

app/features/feedback/model/use-feedback-form.ts
askov: Updated the import. We should get rid of compatibility re-export to avoid these cases :)

app/features/feedback/ui/BaseStarRating.tsx
```diff
@@ -0,0 +1,37 @@
...
+                    aria-label={`${star} of ${STARS.length} stars`}
+                    className="cursor-pointer border-0 bg-transparent p-1"
+                    onClick={() => onChange?.(star)}
+                    role="radio"
+                    type="button"
+                >
+                    <Star
+                        aria-hidden="true"
+                        className={cn(
+                            star <= value ? 'fill-accent text-accent' : 'fill-heavy-metal-600 text-heavy-metal-600',
```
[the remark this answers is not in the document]
askov: fixed

app/features/feedback/env.ts
askov: We don't have a strong agreement on this, but I updated for consistency

app/features/feedback/ui/FeedbackTrigger.tsx
askov: yes, good idea, done

mikhd (approved):

PR #1292, opened 2026-09-07
fix(slot-time): measure the dashboard rate with the entity's formula

askov:
## Description

The home page measured milliseconds per slot with its own formula, which disagreed with the one the feature-gate countdown uses. It read high, so the epoch countdown built on it overstated the time remaining.

`solanaDashboardInfo` averaged the per-sample rates unweighted. A slow minute produces fewer slots by definition, so weighting every sample alike gives the slow minutes more than their share. On five mainnet samples at the 300 ms gate with one degraded minute it returned 371 ms where the cluster ran at 347 ms — 2.88 h apart over a full epoch.

### The fix

-   **Measure the dashboard rate with the slot-time entity.** `measureMsPerSlot` divides total time by total slots once, so both home-page figures agree with the feature-gate countdown.
-   **Make the sample window a parameter.** The dashboard measures over 60 samples and over 1, the countdown keeps its 5, and the entity no longer hardcodes one window.
-   **Name the unit in the fields.** `avgSlotTime_1h` and `avgSlotTime_1min` held seconds and became `msPerSlot_1h` and `msPerSlot_1min`, which drops three unit conversions at the call sites.
-   **Hold the figures when the newest minute produced no slot.** The old code labelled an older minute "1min average"; nothing new is reported until the next poll.

## Type of change

-   [x] Bug fix

## Testing

-   **The 1hr average no longer runs high.** Read `Slot time (1hr average)` in `Live Cluster Stats` — [preview](https://explorer-git-fork-hoodieshq-fix-unify-69ae0a-solana-foundation.vercel.app) · [production](https://explorer.solana.com). Expect a figure at or below production's, matching `Slot time (1min average)` on a steady cluster; production reads high whenever any recent minute was slow.
-   **Epoch time remaining shortens to match.** Read `Epoch time remaining (approx.)` in the same card — [preview](https://explorer-git-fork-hoodieshq-fix-unify-69ae0a-solana-foundation.vercel.app) · [production](https://explorer.solana.com). Expect it to agree with the `Cluster Activation` countdown on a feature-gate account; production overstates it by the same proportion the rate is inflated.

## Related Issues

Closes [HOO-1416](https://linear.app/solana-fndn/issue/HOO-1416/unify-slot-time-measurement-two-formulas-disagree)

## Checklist

## Additional Notes

`build:info` regenerates `bench/BUILD.md` with a first-load shift across the `/address/*` routes, but a build of `master` with no changes produces the same table, so the drift predates this PR and is left out of the diff.

The two countdowns now share a formula, not a window: the home page measures over 60 samples and the feature-gate countdown over 5, so they can still differ slightly on a cluster whose rate is moving.

### DIFF
--- a/app/providers/stats/solanaDashboardInfo.tsx
+++ b/app/providers/stats/solanaDashboardInfo.tsx
@@ -1,10 +1,15 @@
+import { measureMsPerSlot } from '@entities/slot-time';
+
 import { ClusterStatsStatus } from './solanaClusterStats';
 import { PerformanceSample } from './solanaPerformanceInfo';
 
+/** Each sample covers one minute, so an hour of history is 60 of them. */
+const SAMPLES_PER_HOUR = 60;
+
 export type DashboardInfo = {
     status: ClusterStatsStatus;
-    avgSlotTime_1h: number;
-    avgSlotTime_1min: number;
+    msPerSlot_1h: number;
+    msPerSlot_1min: number;
     epochInfo: EpochInfo;
     blockTime?: number;
     lastBlockTime?: BlockTimeInfo;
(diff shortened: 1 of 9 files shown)


mikhd (approved):

app/providers/stats/solanaDashboardInfo.tsx
```diff
@@ -75,55 +80,40 @@ export function dashboardInfoReducer(state: DashboardInfo, action: DashboardInfo
...
-            const samples = action.data
-                .filter(sample => {
-                    return sample.numSlots !== BigInt(0);
-                })
-                .map(sample => {
-                    return sample.samplePeriodSecs / Number(sample.numSlots);
-                })
-                .slice(0, 60);
+            const msPerSlot_1h = measureMsPerSlot(action.data, SAMPLES_PER_HOUR);
+            const msPerSlot_1min = measureMsPerSlot(action.data, 1);
```
mikhd: **nit**: use const as well `const SAMPLES_PER_MINUTE = 1`?

mikhd (approved):

app/providers/stats/solanaDashboardInfo.tsx
askov: added

rogaldh (approved):

Woody4618 (approved):

PR #1293, opened 2026-09-07
refactor(hooks): replace use-async-effect with SWR

askov:
## Description

Closes [HOO-1135](https://linear.app/solana-fndn/issue/HOO-1135/replace-use-async-effect). Replaces both `use-async-effect` call sites with SWR, deletes the unreachable `useTokenMetadata` hook, and drops the dependency.

## Changes

- Reads token symbols in the transaction Tokens card through SWR.
- Takes the symbol fetcher's mints from the SWR key instead of a closure over `rows`.
- Throws in the symbol fetcher when `getTokenInfos` returns undefined.
- Skips the symbol request when no chainId can be derived from the cluster and genesis hash.
- Deletes `useTokenMetadata` and its `@entities/nft` export.
- Removes the unused `useMetadata` prop from `Address`.
- Removes `use-async-effect` from `package.json`.
- Adds 7 tests covering symbol mapping, the `tokens` fallback, refetch on a cluster change, and the Custom-cluster gate.

## Behavior Changes

- The four `/block/[slot]` routes ship about 90 kB less JavaScript.
- A failed symbol request no longer pins the `tokens` fallback for the rest of the session.

## Type of change

-   [x] Other (please describe): refactor and dependency removal

## Testing

- **The Tokens card resolves symbols.** Open a transaction that moves SPL tokens — [preview](https://explorer-git-fork-hoodieshq-refactor-r-52db00-solana-foundation.vercel.app/tx/rCtrr1ADMEnKj54CLq3mB3313YrLAki5gcLxSnLSTa979Fe7NGfuyM2DJqJE5zxL992nEMRgxmWyckvhVJwNsfS) · [production](https://explorer.solana.com/tx/rCtrr1ADMEnKj54CLq3mB3313YrLAki5gcLxSnLSTa979Fe7NGfuyM2DJqJE5zxL992nEMRgxmWyckvhVJwNsfS) — and expand `Tokens`. Production shows the same symbols; this step confirms the SWR rewrite did not drop them.

## Related Issues

Closes [HOO-1135](https://linear.app/solana-fndn/issue/HOO-1135/replace-use-async-effect)

## Checklist

### Deleting the hook instead of migrating it

`useMetadata` has never been passed to `Address` in the repo's history, so `useTokenMetadata` has never run. Deleting it is the end state @mikhd asked for in review.

The deletion removes the only `@entities/nft` import in `Address`. That barrel re-exports `getUmi`, which pulls in `@metaplex-foundation/mpl-token-metadata`, so every route that rendered an `Address` and nothing else Metaplex carried the library. That is the 90 kB the block routes drop.


app/features/transaction/ui/TokenBalancesCard.tsx
```diff
@@ -60,18 +60,11 @@ export type TokenBalancesCardInnerProps = {
...
-        async isMounted => {
-            const mints = rows.map(r => new PublicKey(r.mint));
-            const tokens = await getTokenInfos(mints, cluster, genesisHash);
-            if (!isMounted()) return;
-            setTokenSymbols(new Map(tokens?.map(t => [t.address, t.symbol])));
-        },
-        [mintKey, cluster, genesisHash],
+    const { data: tokenSymbols } = useSWRImmutable(
+        mintKey ? (['token-balance-symbols', mintKey, cluster, genesisHash] as const) : undefined,
+        fetchTokenSymbols,
```
mikhd, line 71: **question**: should we add `keepPreviousData: true`?
I mean is this possible for users to see 'tokens' fallback when key changes and tokenSymbols is `undefined`?

app/entities/nft/model/use-token-metadata.ts
```diff
@@ -3,35 +3,27 @@
 import { fetchMetadata, findMetadataPda, type Metadata } from '@metaplex-foundation/mpl-token-metadata';
 import { publicKey } from '@metaplex-foundation/umi';
 import { useCluster } from '@providers/cluster';
-import { useState } from 'react';
-import useAsyncEffect from 'use-async-effect';
+import useSWRImmutable from 'swr/immutable';
 
 import { getUmi } from '../lib/umi';
 
 export const useTokenMetadata = (useMetadata: boolean | undefined, pubkey: string) => {
```
mikhd: By the way is `useTokenMetadata` hook being used in Explorer?
Looks like it's imported by [Address.tsx](https://github.com/solana-foundation/explorer/blob/master/app/components/common/Address.tsx#L75) only but never actually enabled via `useMetadata` prop. Worth checking if it can be just removed.

mikhd (approved):

app/features/transaction/ui/TokenBalancesCard.tsx
askov: No, I don't think it earns its place here. The key only changes when the mints or the cluster change. Within a transaction page `rows` are fixed, so the key is stable after the first fetch. When the cluster does change the symbols genuinely differ, and `keepPreviousData` would show the previous cluster's symbols against the new one.

You're right that the first paint shows `tokens`, but the effect this replaces did the same: `tokenSymbols` started as an empty `Map`, so `.get(mint)` missed until the request resolved. No change in that window.

app/entities/nft/model/use-token-metadata.ts
askov: Good catch - done in e9d9ceae.

`useMetadata=` has never been passed anywhere in the repo's history (`git log -S 'useMetadata=' --all` returns nothing), so the hook has never run. Deleted the hook, its `@entities/nft` export, and the prop on `Address`.

It turned out to be worth more than the dead code: that was the only `@entities/nft` import in `Address`, and the barrel re-exports `getUmi`, which pulls in `@metaplex-foundation/mpl-token-metadata`. Any route that rendered an `Address` and nothing else Metaplex carried the library. The four `/block/[slot]` routes drop about 90 kB in `bench/BUILD.md`.

app/features/transaction/ui/TokenBalancesCard.tsx
```diff
@@ -99,13 +92,22 @@ export function TokenBalancesCardInner({ rows }: TokenBalancesCardInnerProps) {
...
     );
 }
 
+type TokenSymbolsSwrKey = readonly ['token-balance-symbols', string, Cluster, string | undefined];
+
+// genesisHash is required to derive a chainId on the Custom cluster - without it getTokenInfos returns [].
+async function fetchTokenSymbols([, mintKey, cluster, genesisHash]: TokenSymbolsSwrKey): Promise<Map<string, string>> {
+    const mints = mintKey.split(',').map(mint => new PublicKey(mint));
+    const tokens = await getTokenInfos(mints, cluster, genesisHash);
+    return new Map(tokens?.map(t => [t.address, t.symbol]));
```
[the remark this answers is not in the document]
askov: Confirmed and fixed in e9d9ceae. `fetchTokenInfosFromApi` swallows both a non-ok response and a thrown request into `undefined`, and reading that as an empty `Map` cached the fallback under an immutable key.

The fetcher now throws on `undefined`, so SWR records an error instead of a successful empty result.

One thing your description doesn't separate: `getTokenInfos` also returns `undefined` when `getChainId` finds no chain, which is permanent for that cluster rather than transient. Throwing on it would retry forever. So the card now gates the SWR key on `getChainId(cluster, genesisHash)` and only the genuine request failure throws.

Both paths are covered in `TokenBalancesCard-symbols.spec.tsx`.

app/features/transaction/ui/TokenBalancesCard.tsx
```diff
@@ -60,18 +60,11 @@ export type TokenBalancesCardInnerProps = {
...
-            const mints = rows.map(r => new PublicKey(r.mint));
-            const tokens = await getTokenInfos(mints, cluster, genesisHash);
-            if (!isMounted()) return;
-            setTokenSymbols(new Map(tokens?.map(t => [t.address, t.symbol])));
-        },
-        [mintKey, cluster, genesisHash],
+    const { data: tokenSymbols } = useSWRImmutable(
+        mintKey ? (['token-balance-symbols', mintKey, cluster, genesisHash] as const) : undefined,
+        fetchTokenSymbols,
     );
```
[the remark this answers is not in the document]
askov: Added in e9d9ceae: `app/features/transaction/ui/__tests__/TokenBalancesCard-symbols.spec.tsx`, 7 tests covering the mints passed to `getTokenInfos`, address-to-symbol mapping, the `tokens` fallback for a mint the response omits, refetching when the cluster changes, the failed-request case, and both Custom-cluster branches.

app/entities/nft/model/__tests__/use-token-metadata.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+        await waitFor(() => expect(result.current.data).toEqual({ name: `name:pda:${OTHER_MINT}` }));
+    });
+
+    it('should report no data for a mint without a metadata account', async () => {
+        mocks.fetchMetadata.mockRejectedValue(new Error('account not found'));
+
+        const { result } = renderTokenMetadata(true, MINT);
+
+        await waitFor(() => expect(mocks.fetchMetadata).toHaveBeenCalled());
+        expect(result.current.data).toBeUndefined();
```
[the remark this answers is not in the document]
askov: The assertion was weak as you describe - `data` is already `undefined` on the first render, so the `waitFor` on `fetchMetadata` having been called proved nothing about post-rejection behaviour.

It's moot in e9d9ceae: `useTokenMetadata` and this spec are both deleted. `Address`'s `useMetadata` prop has never been passed by any caller, so the hook was unreachable and went rather than moving to SWR.

Woody4618 (approved):

PR #1294, opened 2026-09-07
feat(seo): env-gate robots.txt, sitemaps, and the noindex header for staging

askov:
## Description

`public/robots.txt` is a static file, so every deployment serves the same allow-all policy and every sitemap stays reachable. Staging deployments are fully crawlable as a result, and bots have scanned one.

Closing a deployment to crawlers takes three signals, and `robots.txt` carries only one. `Disallow: /` stops a crawl but cannot express `nofollow`, and it does nothing about URLs a crawler already holds.

### What's new

-   **Gate the robots policy on one server-only flag.** `SEO_DISALLOW_BOTS` picks between the allow policy and `Disallow: /`. `isSeoDisallowBots()` in `app/utils/env.ts` is the reader everywhere Next bundles the code. `config/headers.mjs` re-checks the variable itself, because Node loads `next.config.mjs` unbundled and cannot import the TypeScript helper.
-   **Serve robots.txt from a route handler.** `app/robots.txt/route.ts` replaces the static file and owns its `Cache-Control`, with a sha256 `ETag` per variant and a `304` on a matching `If-None-Match`. The allow policy is byte-identical to the deleted file, `/mcp` rules from #1308 included — deleting the file would otherwise drop them.
-   **Return 404 for the three sitemaps when flagged.** `proxy.ts` matches `/sitemap.xml`, `/default-sitemap.xml` and `/accounts-sitemap.xml`. The generator script and its CI drift check stay untouched.
-   **Send `X-Robots-Tag: noindex, nofollow` when flagged.** `buildHeaders()` in `config/headers.mjs` adds it on every path. It catches bots that fetch a URL before re-reading `robots.txt`, and bots that ignore `robots.txt` but honour the header. It does **not** remove URLs an index already holds: `Disallow: /` stops a compliant crawler from fetching the page, so the header never gets read. Clearing what staging already leaked is a separate job.
-   **Cover both flag states with specs.** New specs for the route and for `buildHeaders()`, extended specs for the proxy. The header table moved out of `next.config.mjs` to sit beside `buildRedirects()`, for the same reason that one was extracted: it is testable without loading the Sentry/BotID wrappers.

## Type of change

-   [x] New feature

## Testing

-   **Default robots.txt.** Request `/robots.txt` with `SEO_DISALLOW_BOTS` unset. Expect the allow policy with its `Sitemap:` line and an `ETag` header, then send that `ETag` back as `If-None-Match` and expect `304`.
-   **Flagged robots.txt.** Build with `SEO_DISALLOW_BOTS=true` and request `/robots.txt`. Expect `Disallow: /`, no `Sitemap:` line, and a different `ETag`.
-   **Sitemaps.** Request `/sitemap.xml`, `/default-sitemap.xml` and `/accounts-sitemap.xml` in both builds. Expect `200` unflagged and `404` flagged.
-   **Page header.** Request any page in both builds. Expect no `X-Robots-Tag` unflagged, and `X-Robots-Tag: noindex, nofollow` flagged.
-   **The flag is read at build time.** `next.config.mjs` reads it inside `headers()`, so the flagged state needs a build with the variable set. A deployment without it exercises the default state only.

## Related Issues

Closes [HOO-1431](https://linear.app/solana-fndn/issue/HOO-1431/improve-robotstxt-handling-for-staging-bot-protection)

## Checklist

## Additional Notes

The regenerated `bench/BUILD.md` adds the `/robots.txt` row and nothing else. An earlier revision of this branch also moved `/address/*` first-load sizes by 10–20 kB; that came from the stale base and is gone after the rebase onto `4432becad`.

`Bypass crawler files` in `docs/firewall.md` exempts robots and sitemaps from Bot Filter, which is what lets a crawler read the disallow policy at all. Firewall parity between staging and production is tracked separately.


config/headers.mjs
```diff
@@ -0,0 +1,24 @@
+const SEO_FILE_HEADERS = [{ key: 'Cache-Control', value: 'public, max-age=3600, stale-while-revalidate=86400' }];
+
+// robots.txt is absent here: its route handler owns its own Cache-Control.
+const SITEMAP_PATHS = ['/sitemap.xml', '/default-sitemap.xml', '/accounts-sitemap.xml'];
```
mikhd, line 4: **question** : is this possible and make sense to extract const and share between proxy.ts/headers.mjs/route.ts?

```
// config/constants.mjs
export const SITEMAP_PATHS = ['/sitemap.xml', '/default-sitemap.xml', '/accounts-sitemap.xml'];
export const CRAWLER_BYPASS_PATHS = ['/robots.txt', ...SITEMAP_PATHS];
export const CACHE_CONTROL_HEADERS = [{ key: 'Cache-Control', value: 'public, max-age=3600, stale-while-revalidate=86400' }]
```

.env.example
```diff
@@ -23,6 +23,9 @@ NEXT_PUBLIC_METADATA_USER_AGENT="Solana Explorer"
 ## Configuration for "interactive IDL" feature enabled
 NEXT_PUBLIC_INTERACTIVE_IDL_ENABLED=true
 NEXT_PUBLIC_BAD_TOKENS=
+## Configuration for SEO bots access. Set to "true" on staging builds to close the deployment for bots:
+### robots.txt serves "Disallow: /", sitemaps return 404, every page carries "X-Robots-Tag: noindex, nofollow"
+SEO_DISALLOW_BOTS=false
```
mikhd, line 28: Just a question: will we be able to disable it for dev prod when we need to test shared pages on social media? (It's the case when bots actually should reach website)

mikhd (approved):

config/headers.mjs
askov: I tried the spread - matcher: ['/api/:path*', ...SITEMAP_PATHS] - and the build fails: Next.js can't recognize the exported config field in route. matcher needs to be a static string or array of static strings or array of static objects.

.env.example
askov: yes, but redeploy is required

PR #1311, opened 2026-09-14
feat(alpenglow): track the Alpenglow migration on the home page

askov:
## Description

Closes [HOO-1629](https://linear.app/solana-fndn/issue/HOO-1629/track-alpenglow-upgrade-on-the-explorer-homepage). Adds a home page card that reads `getAgGenesisCert` to show whether the active cluster has switched to Alpenglow consensus, and the genesis slot once it has.

## Changes

- Adds `app/features/alpenglow/`, holding the RPC call, the certificate parser, the status hook, and the card.
- Calls `getAgGenesisCert` through `@solana/kit`, which carries the method and upcasts the slot to a `bigint`.
- Parses the certificate through one constructor that checks the block id is 32 bytes and the slot is a non-negative `bigint`, since kit types the response without checking it at runtime.
- Returns `unsupported`, `refused` and `unreadable` as values rather than throwing, so none of them is retried.
- Classifies a decline from the JSON-RPC code where the body carries one and from the HTTP status where it does not, reading a 404 as the method not being served, because kit throws on a non-2xx status before reading the body.
- Reads the status with `useSWR` and revalidation off on focus, reconnect and stale, so a focus, a reconnect or a revisit never re-asks a settled answer.
- Avoids `swr/immutable`, which sets those three flags but also pins `refreshInterval` to 0 and would stop the poll.
- Polls every 60s only while a cluster is still waiting for its certificate.
- Caps error retries at 3, since SWR retries indefinitely when `errorRetryCount` is unset.
- Renders the card on the home page between the cluster stats and the developer resources.
- Moves `isMethodNotFound` from `app/features/transaction-history/lib/` to `app/shared/lib/`, now that two features classify the same error.
- Adds 59 tests covering the parser, the RPC call, the status hook, and the card.

## Behavior Changes

- The home page shows an `Alpenglow on <cluster>` card naming the consensus in force.
- The card adds `Genesis slot` and `Certified block` rows once a cluster has migrated, with the slot linked to its block.
- The card renders nothing on an endpoint that does not implement `getAgGenesisCert` or refuses it.

## Type of change

-   [x] New feature

## Screenshots
desktop

mobile

## Testing

- **The card reports testnet consensus.** Open the home page on testnet — [preview](https://explorer-git-fork-hoodieshq-feat-alpen-124d4f-solana-foundation.vercel.app/?cluster=testnet) · [production](https://explorer.solana.com/?cluster=testnet) — and scroll to between `Live Cluster Stats` and the developer resources. Expect an `Alpenglow on Testnet` card whose `Consensus` row reads `TowerBFT — Alpenglow not activated`. Production has no such card.
- **The card reports devnet consensus.** Same page on devnet — [preview](https://explorer-git-fork-hoodieshq-feat-alpen-124d4f-solana-foundation.vercel.app/?cluster=devnet) · [production](https://explorer.solana.com/?cluster=devnet). Expect an `Alpenglow on Devnet` card in the same position. Production has no such card.
- **No card on an endpoint that refuses the method.** Open the home page on mainnet — [preview](https://explorer-git-fork-hoodieshq-feat-alpen-124d4f-solana-foundation.vercel.app/) · [production](https://explorer.solana.com/) — and check the same position. Expect no Alpenglow card and no error or placeholder, because the mainnet endpoint rate-limits `getAgGenesisCert`. Production is unchanged here.
- **The upgrade link leaves the Explorer.** On the testnet card — [preview](https://explorer-git-fork-hoodieshq-feat-alpen-124d4f-solana-foundation.vercel.app/?cluster=testnet) — click `About the upgrade` in the card header. Expect `solana.com/upgrades/alpenglow` in a new tab.
- **The rest of the home page is untouched.** Compare testnet — [preview](https://explorer-git-fork-hoodieshq-feat-alpen-124d4f-solana-foundation.vercel.app/?cluster=testnet) · [production](https://explorer.solana.com/?cluster=testnet). Expect `Circulating Supply`, `Active Stake`, `Live Cluster Stats` and `Live Transaction Stats` to read the same on both, with the new card added below them.

## Related Issues

Closes [HOO-1629](https://linear.app/solana-fndn/issue/HOO-1629/track-alpenglow-upgrade-on-the-explorer-homepage)

## Checklist

## Additional Notes

No route bundle moved, so `bench/BUILD.md` is unchanged.


app/features/alpenglow/api/fetch-ag-genesis-cert.ts
```diff
@@ -0,0 +1,117 @@
...
+        return { cert: parseGenesisCert(body.result), kind: 'present' };
+    } catch (cause) {
+        // The node answered, and will answer the same way next time. Worth saying out loud —
+        // a node minting a certificate this app cannot read is someone's bug.
+        Logger.warn(`[alpenglow] ${METHOD} returned a result that is not a certificate`, { cause });
+        return { kind: 'unreadable' };
+    }
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
```
mikhd, line 107: **nit**: probably could be useful as shared helper. I just had a feeling that I already saw the same, for example:

- `packages/entity-inspector/src/shared/parse-helpers.ts#L4` (this one in packages, though)
- `app/features/token-batch/lib/rpc-parsed-batch.ts#L78`

mikhd (approved):

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: It's gone due to kit upgrade

app/__tests__/mock-rpc.ts
```diff
@@ -39,6 +39,10 @@ export const mockFirstAvailableBlock = (block?: bigint): bigint => block ?? gen.
 
 /** Creates a mock RPC object matching the shape returned by createSolanaRpc() */
 export const mockSolanaRpc = (overrides?: Partial<ClusterInfo> & { genesisHash?: string }) => ({
+    // Null is what every cluster still answers, and what the home page's Alpenglow card reads.
```
Woody4618: This comment will be out of date soon 
and “what the card reads” describes a consumer rather than explaining what null means.

Should be something like 
"Before the Alpenglow transition, no genesis certificate exists, so the RPC returns null."

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
```diff
@@ -0,0 +1,102 @@
...
+
+import { type AlpenglowGenesisCert, parseGenesisCert } from '../lib/genesis-cert';
+
+/**
+ * What a node can say about the Alpenglow migration. `unsupported`, `refused` and `unreadable` are
+ * answers rather than failures: the endpoint does not have this method, will not serve this caller,
+ * or served something that is not a certificate. The next request gets the same reply, so all three
+ * are carried as values — and against `refused` the retries are the very thing being limited.
+ */
+export type AgGenesisCertAnswer =
```
Woody4618, line 16: From where did you get these 5 answers?

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
```diff
@@ -0,0 +1,102 @@
...
+    | { kind: 'unsupported' }
+    | { kind: 'refused' }
+    | { kind: 'unreadable' }
+    | { kind: 'absent' }
+    | { kind: 'present'; cert: AlpenglowGenesisCert };
+
+const METHOD = 'getAgGenesisCert';
+
+// Helius declines a method it does not serve with this, and kit throws on the status before the
```
Woody4618: What does helius have to do with this?

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
```diff
@@ -0,0 +1,102 @@
...
+// body saying so can be read.
+const NOT_SERVED = 404;
+
+// Observed against `explorer-api.mainnet-beta.solana.com`, which refuses this method with it while
```
Woody4618: This sentence is so weird.  Its not clear what "it" is and what is the in the same second supposed to mean?

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
```diff
@@ -0,0 +1,102 @@
...
+// answering getEpochInfo from the same caller in the same second.
+const RATE_LIMITED = 429;
+
+// The endpoint will not serve this caller — typically a missing or wrong key on a custom RPC. The
+// next request carries the same key, so retrying only spends it against the same wall.
+const UNAUTHORIZED = [401, 403];
+
+/**
+ * Asks a node for the Alpenglow genesis certificate.
```
Woody4618: This reads like an architectural essay. 
Should be something like this: 
/**
 * Fetches the Alpenglow genesis certificate directly from the RPC endpoint.
 *
 * Transient errors are thrown for retrying; permanent errors are returned as values.
 */

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
```diff
@@ -0,0 +1,102 @@
...
+        return { cert: parseGenesisCert(result), kind: 'present' };
+    } catch (cause) {
+        // The node answered, and will answer the same way next time. Worth saying out loud —
+        // a node minting a certificate this app cannot read is someone's bug.
+        Logger.warn(`[alpenglow] ${METHOD} returned a result that is not a certificate`, { cause });
+        return { kind: 'unreadable' };
+    }
+}
+
+/**
```
Woody4618: These sentences are just so weird. "Undefined is the latter" forces you to reread the first part of the sentence.  

Should be something like: 
/**
 * Returns a definitive response, or `undefined` for a transient failure
 * that the caller should rethrow/retry.
 */

app/features/alpenglow/lib/__tests__/genesis-cert.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    it.each([
+        ['negative', -1n],
+        ['fractional', 1.5],
+        ['a plain integer kit would have upcast', 460_012_345],
+        ['a string', '460012345'],
+    ])('should reject a slot that is %s', (_label, slot) => {
+        expect(() => parseGenesisCert(certPayload({ block: { blockId: BLOCK_ID_BYTES, slot } }))).toThrow();
+    });
+
+    // A block id that is the wrong kind of thing, not merely the wrong length. The card reads a
```
Woody4618: Same here, "wrong kind i thing" i super vague. 

Should be something like: 
// Block IDs must be exactly 32 bytes; reject other types and malformed byte arrays.

app/features/alpenglow/model/__tests__/use-alpenglow-status.spec.tsx
```diff
@@ -0,0 +1,300 @@
...
+        mocks.fetchAgGenesisCert.mockRejectedValue(new Error('node is unhealthy'));
+
+        const { result } = renderStatus({ dedupingInterval: 0 });
+        await waitFor(() => expect(result.current.kind).toBe('unavailable'));
+        await advanceBy(60_000);
+
+        expect(mocks.fetchAgGenesisCert).toHaveBeenCalledTimes(ERROR_RETRY_COUNT + 1);
+    });
+
+    // The card is already right; a blip on the poll must not take it down. Data is read before
```
Woody4618: The wording is unnecessarily metaphorical:

“The card is already right”

app/features/alpenglow/lib/__tests__/genesis-cert.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    // base58 string straight out of this, and every one of these would decode to something.
+    it.each([
```
Woody4618, line 51: Im stopping here but there seems to be a consistent style in these comments: simple concepts are being expressed through conversational/metaphorical phrases like “node blipped,” “card is already right,” “reply the endpoint will repeat,” “wrong kind of thing.” That's making the code harder rather than easier to read.

app/features/alpenglow/ui/AlpenglowCard.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { useCluster } from '@providers/cluster';
+
+import { useAlpenglowStatus } from '../model/use-alpenglow-status';
+import { BaseAlpenglowCard } from './BaseAlpenglowCard';
+
+/**
+ * Renders nothing until a node has answered, and nothing at all where one cannot: an endpoint that
+ * predates the method has no migration to report, so a skeleton there would promise a card that
+ * never arrives.
+ */
```
Woody4618, line 13: Should be 
/**
 * Renders nothing until the first response, or when the RPC endpoint
 * does not support the method.
 */

Woody4618 (changes requested):

app/__tests__/mock-rpc.ts
askov: Fixed

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: It just describes what is happening actually:
- `present` / `absent` are the method's contract: a certificate, or null before
the migration.
- `unsupported`: Agave reports -32601; some providers report -32603 with a
  "method not found" message; others answer HTTP 404. kit throws on the status before the body is read, so 404 is classified by status alone
- `refused`: 429 from a rate limiter, 401/403 for a missing or wrong key.
- `unreadable`: the result does not parse as a certificate

Fixed the comment

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: Triton is not the only rpc explorer can be used with, so we need to support different cases, this one is for Helius. Changes the comment to be more generic

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: It = error under comment. And the second part means that the 429 error here is not usual, not because of throttling or something but just because of the cluster. Rewrote the comment

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: fixed

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: fixed

app/features/alpenglow/model/__tests__/use-alpenglow-status.spec.tsx
askov: fixed

app/features/alpenglow/lib/__tests__/genesis-cert.spec.ts
askov: I see, did a full pass on everything

app/features/alpenglow/ui/AlpenglowCard.tsx
askov: fixed

app/features/alpenglow/lib/__tests__/genesis-cert.spec.ts
askov: fixed

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
Woody4618: Where did you get these from? Are these part of the RPC spec?

Woody4618 (changes requested):

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
Woody4618: Explorer should not be build around certain RPC providers but the RPC standart. Also its just a 404. please remove.

askov: But that is just an app abstraction we need to answer questions like: can we draw it? can we retry it? All these are standard: -32601 (method not found) and -32603 (internal error) are from json-rpc 2.0, 404, 429 and 401/403 are ordinary http. So there is nothing invented and nothing to delete. Other places in explorer are also supporting different providers, it's not the first one

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: I answered this here https://github.com/solana-foundation/explorer/pull/1311#discussion_r4026775573

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
Woody4618: What im saying is its not about helius. 
Its probably just: "An RPC that does not support this call will return 404" 
also 404 generally stands for not_found not not_served

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
Woody4618: I wanted to know where 
{ kind: 'unsupported' }
    | { kind: 'refused' }
    | { kind: 'unreadable' }
    | { kind: 'absent' }
    | { kind: 'present';
come from. Is that rpc spec?

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
Woody4618: What im saying is that it probably is not tied to helius but just:
"If an RPC does not support this call it will return 404"

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
Woody4618: I wanted to know where
{ kind: 'unsupported' }
| { kind: 'refused' }
| { kind: 'unreadable' }
| { kind: 'absent' }
| { kind: 'present';
come from. Is that rpc spec?

app/features/alpenglow/api/fetch-ag-genesis-cert.ts
askov: I see your point about naming, not_found is standard and I changed it. About everything else:



It **is** about helius in this case (but I changed this comment to more generic). Here is what I'm talking about:

- Here is what helius does when it can't find the method for cluster (404, as expected, according to standards)
helius

- And here is what triton does for this method on mainnet (429, no idea why this code, it's not real 429 and we should not retry it!)
triton

Screenshot 2026-09-18 at 13 14 46

And nothing except this api module should know about these quirks and it's the only one to be changed if triton or helius decides to change something. The union with 5 distinct answers is what the ui model/ui are interested in. I hope it's clear that there is a reason for this code and comments

Woody4618 (approved):

Woody4618:
Why is that commit not signed? 
image

askov:
> Why is that commit not signed? image

Fixed, something went wrong during git conf switch

PR #1314, opened 2026-09-15
chore: add support for v1 in config and execute fetch

amilz:
bump web3js and support v1 in a couple nit locations
todo add send support in the idl tab


askov (approved):

app/features/decode-account-pmp/api/find-config-in-transactions.ts
```diff
@@ -43,7 +44,9 @@ export async function findConfigInTransactions(
             // A failed transaction never wrote a header, so its declared config was never committed.
             if (entry.err) continue;
 
-            const tx = await connection.getTransaction(entry.signature, { maxSupportedTransactionVersion: 0 });
+            const tx = await connection.getTransaction(entry.signature, {
+                maxSupportedTransactionVersion: MAX_SUPPORTED_TRANSACTION_VERSION,
```
Woody4618, line 48: Not sure this is a good idea. Probably switching to a new supported transaction version should be a concious decission and not be automatic? Wdyt?

Woody4618 (approved):

ZaferKenana (approved):

